Issue #2, opened 2012-02-26
Add licensing information


mitar:
Open source project needs licensing information so that other users and developers know under which terms they can use and distribute our project. As we are developing a web portal I propose to use [AGPL](https://www.gnu.org/licenses/agpl.html) license so that any additions to source code have to be shared even if the program is only run on the server and not distributed. For media content I propose [Creative Commons](https://creativecommons.org/) Attribution-ShareAlike license.

You can help yourself with the `COPYING` and `LICENSE` files of the [nodewatcher](http://dev.wlan-si.net/browser/nodewatcher) project.

PR #12, opened 2012-02-28
Issue #2

domenc:
Add licensing information (license & copying) and add readme file (accidentally deleted).


mitar:
Not good. You have to make a branch from the initial repository state (my first commit). See [this first answer here](http://stackoverflow.com/questions/1655361/how-to-clone-an-old-git-commit-and-some-more-questions-about-git). Not from the end, then there is no use of that. The idea is that the branch contains only few commits directly connected with the issue. Not also testing commits. Try again. ;-)

Issue #10, opened 2012-02-27
Configure gitignore file


mitar:
In the repository there should be `.gitignore` file configured so that unwanted files are not tracker or committed by git (by accident). For example, `*.pyc` files, `.*` files, and `env_piplmesh` are very good candidates for files/directories we will probably not want to have in our repository.

PR #14, opened 2012-02-28
Issue #10

domenc:
Add .gitignore file - ignore env_piplmesh folder & all *.pyc files.


mitar:
Not good. You have to make a branch from the initial repository state (my first commit). See [this first answer here](http://stackoverflow.com/questions/1655361/how-to-clone-an-old-git-commit-and-some-more-questions-about-git). Not from the end, then there is no use of that. The idea is that the branch contains only few commits directly connected with the issue. Not also testing commits. Try again. ;-)

BTW, you are committing `.pyc` files what is directly against the idea of issue #10.

Issue #3, opened 2012-02-26
Add Python requirements


mitar:
It is common to include `requirements.txt` file which lists all Python requirements of the project so that they can be installed with `pip install -r requirements.txt`. File can be created with `pip freeze` (inside virtualenv, of course).

PR #17, opened 2012-02-28
Pull request for issue #3

NejcZupec:
requirements.txt created with pip freeze.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+django 1.3.1
+python 2.7.2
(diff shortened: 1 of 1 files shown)


NejcZupec:
I have no idea :) I ran pip freeze command and this is result.

mitar:
It seems wsgiref is a part of standard Python distribution since 2.5, so you could remove that. For your practice, make a new pull request with only Django as output of pip freeze in the file. ;-)

Issue #1, opened 2012-02-26
Initial Django installation with static welcome page


mitar:
Initial [Django](https://www.djangoproject.com/) installation with static welcome page. On the top there should be logo on the left and on the right Google search box for users to be able to easily start their searches on Google (they are redirected to Google, we are not doing a custom Google search).

Put Django site in the `piplmesh` subdirectory of the repository, and then structure features as Django apps inside (Python submodules). This static page could be in the `frontend` app (so the whole name of the app is `piplmesh.frontend`). For example, we might get a structure something like:

```
<repository root>/
  docs/
    ...
  piplmesh/
    account/
      __init__.py
      models.py
      ...
    api/
      __init__.py
      models.py
      ...
    formats/
      en/
        ...
      __init__.py
    frontend/
      static/
        piplmesh/
          css/
            ...
          js/
            ...
          ...
      templates/
        base.html
        header.html
        footer.html
        home.html
        ...
      tests/
        __init__.py
        ...
      __init__.py
      models.py
      views.py
    __init__.py
    manage.py
    settings.py
    urls.py
  COPYING
  LICENSE
  README
  requirements.txt
```

Split templates according to Django best practices (Google that) using [template inheritance](https://docs.djangoproject.com/en/dev/topics/templates/) and define common template blocks.

PR #18, opened 2012-02-28
Pull request for issue #1

aljana:
Pull request for issue #1. Static html page with some basic CSS with (temporary :)) logo on the left and google search on the right.


mitar:
Fiuu. Great!

mitar:
OK. Good work. But bad pull request. Commit 60e133a is not the way to do things. You have to make a new branch and use `git cherry-pick` to add only those commits which are really part of the pull request (first testing commits are not). So please split those commits into multiple pull requests.

Do not delete README file!

aljana:
How do we make new branch with no commits? It seems impossible because there always appear previous commits from master branch in it.

mitar:
I wrote on the forum. You first have to `git checkout` to the revision you want to branch from (in this case from the initial). You branch from the revision you are in.

Issue #5, opened 2012-02-26
Document development installation


mitar:
Document development (local) installation of PiplMesh. Documentation usually resides in `docs` directory and is written using [Sphinx](http://sphinx.pocoo.org).

You can see an example in [nodewatcher](http://dev.wlan-si.net/browser/nodewatcher) project.

PR #29, opened 2012-02-29
Pull request for Issue #5

kubili:
I added some realy short documentation with some info on project.


mitar:
Nice. Content could be removed somewhat (remove example code and so on), but you should definitely not commit the `_build` directory. Please make a new pull request without that.

PR #34, opened 2012-02-29
Pull request for Issue #5

andrejcesen:
Last pull request for Issue #5.


mitar:
Please create a new pull request, without _build directory even being committed once. The problem is that it would otherwise stay in the repository forever and make repository big, even if you remove it from the current repository state.

Issue #6, opened 2012-02-26
Gravatar support for our users


mitar:
[Gravatar](https://secure.gravatar.com/) support for our users' avatars. Avatars should be automatically retrieved and displayed.

PR #43, opened 2012-03-03
Pull request for issue-#6

domenc:
Gravatar function which return avatar images depends on your email


mitar:
What happens when user does not have a gravatar?

domenc:
Default picture return gravatar services.

piplmesh/api/gravatar.py
```diff
@@ -0,0 +1,10 @@
+# import libs for encoding urls and generating md5 hash
+import urllib, hashlib
+
+# avatar function which return URL to avatar picture
+def gravatarURL(email, size = 40):
+	# construct the url
+	gravatarUrl = "http://www.gravatar.com/avatar/" + hashlib.md5(email.lower()).hexdigest() + "?"
+	gravatarUrl += urllib.urlencode({'s':str(size)})
```
mitar: Please do not concatenate strings, use string interpolation, all at ounce, also the dict-version of interpolation for increased readability.

piplmesh/api/gravatar.py
```diff
@@ -0,0 +1,10 @@
+# import libs for encoding urls and generating md5 hash
+import urllib, hashlib
+
+# avatar function which return URL to avatar picture
+def gravatarURL(email, size = 40):
+	# construct the url
+	gravatarUrl = "http://www.gravatar.com/avatar/" + hashlib.md5(email.lower()).hexdigest() + "?"
```
mitar: Is there a HTTPS version of it? So that URLs works also on HTTPS-based websites (otherwise you get a mix-content warning).

piplmesh/api/gravatar.py
```diff
@@ -0,0 +1,10 @@
+# import libs for encoding urls and generating md5 hash
```
mitar: Probably unnecessary comment, anyway, start it with a upper case.

piplmesh/api/gravatar.py
```diff
@@ -0,0 +1,10 @@
...
+import urllib, hashlib
+
+# avatar function which return URL to avatar picture
```
mitar: Comment about the function should go inside the function, into docstring.

mitar:
Is it possible to have our own default picture? So that it is branded in our way?

domenc:
Default avatar (user who hasn't register in Gravatar): http://www.gravatar.com/avatar/RADNOM_HASH?s=150
OK, i will fix it (strings and ssl)

mitar:
I think it would be better to make this a template tag, not a function. Can you convert this to a Django template tag? (Maybe there even exists something like that.)

domenc:
I will try (first, i need to tech it :) )

domenc:
Yes, we can have own default avatar, but i have problem with HTTP_HOST variable, because I need absolute url to picture (like: http://localhost:800/piplmesh/static/...../avat.png) and HTTP_HOST doesn't work.

Is any other solution to get URL without use HttpRequest class?

mitar:
https://docs.djangoproject.com/en/dev/ref/contrib/sites/

domenc:
I try this too, but i don't understand why this need a database (return error about database engine)?

mitar:
Because it is configured in the database. Please read new documentation for our project (in docs dir). There you have how to install and run our project. And then you have an admin interface, where you can configure sites.

domenc:
Changes:
- string with string interpolation
- add support for ssl (ssl='on'/ssl='off')
- add default avatar

mitar:
Please update the pull request only when you think that you fixed all my concerns. Or at least write that it is partially done thing. So that I know if I should comment again some things or not. ;-)

domenc:
Deleted file gravatar.py from api folder (move to templatetags folder), fix secure gravatar services, update image with my own, create/register template tag for gravatar

piplmesh/settings.py
```diff
@@ -127,6 +130,7 @@
     # Uncomment the next line to enable admin documentation:
     # 'django.contrib.admindocs',
     'piplmesh.frontend',
+	'piplmesh'
```
mitar: Why is this necessary?

mitar: Ah, no. Move to templatags in frontend app. Template tags are used in frontent, no? ;-)

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+
+from django.contrib.sites import models as sites_models
+from django.conf import settings
+
+from django import template
+
+register = template.Library()
+
+@register.inclusion_tag('gravatar.html')
+def gravatar(email, size = 50, default = 'unknown.png'):
```
mitar: Just code style (mine ;-) ): do not use spaces around = in arguments, please.

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
+import urllib, hashlib
+
+from django.contrib.sites import models as sites_models
+from django.conf import settings
+
```
mitar: Unnecessary line.

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
+import urllib, hashlib
+
+from django.contrib.sites import models as sites_models
+from django.conf import settings
```
mitar: Please group imports (all Python together, all Django together, all local together) - you have done that AND order them by alphabet.

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+
+from django import template
+
+register = template.Library()
+
+@register.inclusion_tag('gravatar.html')
+def gravatar(email, size = 50, default = 'unknown.png'):
+	"""
```
mitar: Hm, aren't we using for spaces everywhere?

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+		Description: gravatar function return URL image, depends on user email.
```
mitar: Docstring is probably already a description. Just write what it does, no need to explain that you are explaining.

Indentation is unnecessary here.

piplmesh/settings.py
domenc: ok :D

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+		
+		Template tag - usage:
```
mitar: Please check how other template tags are documented. You can check here for [mine](https://bitbucket.org/mitar/django-missing).

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+		{% load gravatar %}
+		{% gravatar "piplmesh@piplmesh" 50 %}
+	"""
+
+	if settings.GRAVATAR_HTTPS_DEFAULT:
```
mitar: This will fail if `GRAVATAR_HTTPS_DEFAULT` is not defined. For non-default settings use `getattr(settings, 'GRAVATAR_HTTPS_DEFAULT', False)`.

piplmesh/templatetags/gravatar.py
domenc: import urllib, hashlib
from django import template
from django.contrib.sites import models as sites_models

???

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+		http = "https"
+	else:
+		http = "http"
```
mitar: Something like this would be clearer?

```
schema = 'https' if getattr(settings, 'GRAVATAR_HTTPS_DEFAULT', False) else 'http'
```

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+
+	site = sites_models.Site.objects.get_current().domain
```
mitar: Shouldn't this variable be named domain?

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+
+	# Construct the url for default avatar and gravatar services
+	default_avatar_url = "%(http)s://%(site)s%(static)spiplmesh/images/%(avatar)s" % {"http": http, "site": site, "static": settings.STATIC_URL, "avatar": default}
+	email_hash = hashlib.md5(email.lower()).hexdigest()
+	gravatar_url = "https://secure.gravatar.com/avatar/%(email_hash)s?s=%(size)s&d=%(default_avatar)s" % {"email_hash": email_hash, "size": str(size), "default_avatar": urllib.quote(default_avatar_url)}
+
+	return {'gravatar': {'url': gravatar_url, 'size': size}}
```
mitar: Please split this into multiple lines:

```
return {
    'gravatar': {
        'url': gravatar_url,
        'size': size,
    }
}
```

piplmesh/templatetags/gravatar.py
domenc: Is more transparent??

piplmesh/templatetags/gravatar.py
mitar: Please create also a fallback to [RequestSite](https://docs.djangoproject.com/en/dev/ref/contrib/sites/#requestsite-objects) if sites framework is not installed. You can get request object from the context passed to the template tag. You might have to increase Django version we are using (to current development version) to be able to get the context easily in the inclusion tag. But be careful, request object might not be available in the context as it depends on the `TEMPLATE_CONTEXT_PROCESSORS` setting. So in that case really fail. But be careful how template tags fail! Read the doc!

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+		http = "https"
+	else:
+		http = "http"
+
+	site = sites_models.Site.objects.get_current().domain
+
+	# Construct the url for default avatar and gravatar services
+	default_avatar_url = "%(http)s://%(site)s%(static)spiplmesh/images/%(avatar)s" % {"http": http, "site": site, "static": settings.STATIC_URL, "avatar": default}
+	email_hash = hashlib.md5(email.lower()).hexdigest()
+	gravatar_url = "https://secure.gravatar.com/avatar/%(email_hash)s?s=%(size)s&d=%(default_avatar)s" % {"email_hash": email_hash, "size": str(size), "default_avatar": urllib.quote(default_avatar_url)}
```
mitar: `str(size)` is not really needed as you are using `%s` which already does that.

piplmesh/templatetags/gravatar.py
```diff
@@ -0,0 +1,32 @@
...
+
+	if settings.GRAVATAR_HTTPS_DEFAULT:
+		http = "https"
+	else:
+		http = "http"
+
+	site = sites_models.Site.objects.get_current().domain
+
+	# Construct the url for default avatar and gravatar services
+	default_avatar_url = "%(http)s://%(site)s%(static)spiplmesh/images/%(avatar)s" % {"http": http, "site": site, "static": settings.STATIC_URL, "avatar": default}
```
mitar: Please sync variable names with string interpolation key names. (static -> static_url, for example)

Please split dictionary into multiple lines (similar to the return bellow, { stays on top, and then go to new line).

piplmesh/templatetags/gravatar.py
mitar: Yes, just an empty line between Django group and Python built-in group.

piplmesh/templatetags/gravatar.py
mitar: Sorry. _four_ spaces. Not _for_. ;-)

domenc: I don't get it

mitar: OK. The question is how we indent. Tabs, spaces, two or for spaces? So, I think we are using four spaces everywhere, not two. Nor tabs.

domenc: And where is two spaces?

mitar: It displays like that to me about in GitHub?

mitar: Line 12.

domenc: I use tabs, which should be mapped to four spaces

mitar: As you see, they are not. Please never use tabs in the code. And we will not have a discussion about this now. ;-) This is a code style for this project. (This is the problem when multiple people are working on the project, code style has to be consistent.)

piplmesh/templatetags/gravatar.py
domenc: Could this new issue? :D

mitar: No. ;-)

mitar:
BTW, I think you have not forked anew from "official" repository from milestone 0.1 before you started working. Your settings.py file for example is different.

domenc:
Yes, I fork before you wrote on eUcilnica and I update fork today for me and others :)
Can I copy a settings.py from your repository and only add a new const?
Or I close this pull request and create one again?

mitar:
Anyway you will have to make a new pull request once we clean all things. ;-)

domenc:
Aha, ok.
Now, i close this and create new with updated settings and repo :)

Issue #41, opened 2012-03-03
Usernames should be case-insensitive


mitar:
In Django, usernames (for model-based authentication) are by default case-sensitive. This is really crazy and it would be great to make them insensitive, but case-preserving. (Passwords should still be sensitive, of course.)

aljana:
Do we have to allow miXeD-case letters for username or can we simply convert letters to lowercase?

mitar:
Updated issue description.

Usernames should be case-preserving. ;-)

Hint: Check our regular existing project we are stealing from. ;-)

aljana:
Ok I see, but this requires changes in account/forms.py, which is also used by issue #9. So when I make pull request, should the file only contain function that is necessary for this issue? But then it won't work until it's merged?

mitar:
Yes, you have coordinate this with your team. You can make a pull request together for both tickets if there is really no way to split it. Or you could wait (this is why it is in "bonus" milestone) and do it some other time.

mitar:
Or, you could even make two pull requests for this issue, one which is independent of the #9 and one which depends on the pull request of your team. In this way we could pull at least a part of your changes even if some other team's #9 pull request is pulled in at the end.

mitar:
Done with #102.

Issue #9, opened 2012-02-26
User authentication and registration


mitar:
When users visit our portal we want to give them opportunity to register and authenticate. Smart idea is to create username automatically and allow them to use some features of the portal (like customization) immediately, and invite them to register to have their preferences and content persistent (and accessible across sessions). One such way is to use [django-lazysignup](https://github.com/danfairs/django-lazysignup), but you should also research possible other ways to achieve this.

mitar:
Done with #48.

PR #44, opened 2012-03-04
Pull request for Issue #41

aljana:
Pull request that is independent of #9 (so it can't be tested) - only the code that is neccessary for enabling case insensitive usernames. For integration with #9 all you have to do is to extend registration form class with the class CaseInsensitveRegForm which is defined in forms.py.


piplmesh/account/backends.py
```diff
@@ -0,0 +1,22 @@
+from django.db import models
+from django.contrib.auth.models import User
+
+
+
+#This backend uses case insensitive username authentication which is not supported by default
```
mitar: Please do docstring comments for classes.

piplmesh/account/backends.py
```diff
@@ -0,0 +1,22 @@
+from django.db import models
+from django.contrib.auth.models import User
+
+
```
mitar: Unnecessary empty lines?

piplmesh/account/backends.py
```diff
@@ -0,0 +1,22 @@
+from django.db import models
+from django.contrib.auth.models import User
```
mitar: Do not import objects directly. Just modules (`from django.contrib.auth import models as auth_models`).

piplmesh/account/backends.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+
+#This backend uses case insensitive username authentication which is not supported by default
+class CaseInsensitiveModelBackend(object):
```
mitar: Why not simply extend `django.contrib.auth.backends.ModelBackend` and override `authenticate` method?

piplmesh/account/forms.py
```diff
@@ -0,0 +1,16 @@
+from django import forms
+from django.contrib.auth import forms as auth_forms
+from django.contrib.auth import models as auth_models
+from django.contrib.auth.models import User
```
mitar: Again, see above.

piplmesh/account/forms.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+#For enabling case-insensitive but case-preserving registration simply extend registration form class with this class
```
mitar: Again, see above. Please form comments in a consistent way (space after #, upper case begining).

piplmesh/settings.py
```diff
@@ -140,7 +140,7 @@
 INSTALLED_APPS = (
     # Ours are first so that we can override default templates in other apps
     'piplmesh.frontend',
-
```
mitar, line 143: Please keep the empy line for clearity.

piplmesh/settings.py
```diff
@@ -173,3 +173,10 @@
         },
     }
 }
+
+LOGIN_REDIRECT_URL='/'
```
mitar: Please use consistent assignment syntax. (Space before and after `=`.)

piplmesh/account/forms.py
```diff
@@ -0,0 +1,16 @@
...
+class CaseInsensitveRegForm(auth_forms.UserCreationForm):
+	#Nice hack for checking username existence in a case-insensitive manner
```
mitar: I wouldn't really name this a hack, but simply a normal way to do it. Please keep comment aligned with the rest of the code. And move it inside the method.

piplmesh/account/forms.py
```diff
@@ -0,0 +1,16 @@
+from django import forms
+from django.contrib.auth import forms as auth_forms
+from django.contrib.auth import models as auth_models
+from django.contrib.auth.models import User
+
+
+#For enabling case-insensitive but case-preserving registration simply extend registration form class with this class
+class CaseInsensitveRegForm(auth_forms.UserCreationForm):
```
mitar, line 4: You could maybe make this a mixin instead?

piplmesh/settings.py
aljana: I will remove this, it's not related to this issue.

aljana:
Have some problems with understanding modules, with idents and commits.... It's better now I think.

piplmesh/account/forms.py
aljana, line 4: I made this class abstract but I'm not sure if this makes it mixin?

piplmesh/account/backends.py
```diff
@@ -0,0 +1,16 @@
+from django.contrib.auth import models as auth_models
+from django.contrib.auth import backends
+
+class CaseInsensitiveModelBackend(backends.ModelBackend):
+    """
+    This backend uses case insensitive username authentication which is not supported by default
```
mitar: In docstrings, please use full sentences (dot at the end).

piplmesh/account/backends.py
```diff
@@ -0,0 +1,16 @@
+from django.contrib.auth import models as auth_models
+from django.contrib.auth import backends
```
mitar: Organize imports in alphabetical order. Also, those two could be combined into one:

```
from django.contrib.auth import backends, models as auth_models
```

piplmesh/account/forms.py
```diff
@@ -0,0 +1,19 @@
+from django import forms
+from django.contrib.auth import forms as auth_forms
+from django.contrib.auth import models as auth_models
```
mitar: Those two imports could be combined.

piplmesh/account/backends.py
mitar: _case-insensitive_, not _case insensitive_ (because you have like that bellow – just be consistent).

piplmesh/settings.py
```diff
@@ -139,8 +139,9 @@
 
 INSTALLED_APPS = (
     # Ours are first so that we can override default templates in other apps
-    'piplmesh.frontend',
-
+    'piplmesh.frontend',	
+	'piplmesh.account',
```
mitar: Indentation.

aljana: I have constant problems with indentation, keep fixing it but it's never ok, hmm

piplmesh/account/forms.py
```diff
@@ -0,0 +1,19 @@
...
+    def clean_username(self):
+        username = super(CaseInsensitveRegForm, self).clean_username()
+        try:
+            auth_models.User.objects.get(username__iexact=username)
+        except auth_models.User.DoesNotExist:
+            return username
+        raise forms.ValidationError(u"A user with that username already exists.")
+    
+        class Meta:
+            abstract = True
```
mitar, line 18: No, this is wrong. This is something very Django-specific (not Python) and is used in Django model definitions to mark models as abstract (so that database table is not created for those models). Mixin just means that you create a class which by its own does not do anything useful and has to be used as a parent of some other class.

For example, in your case, you could do:

```
class CaseInsensitveUsernameMixin(object):
    ...
```

So a simple Python class without any inheritance. And then you can create case-insensitive form like:

```
class CaseInsensitveRegForm(CaseInsensitveUsernameMixin, auth_forms.UserCreationForm):
    pass
```

Python supports multiple inheritance and we are using this here (this is what means mixin, a parent in multiple inheritance, which just adds or overrides some functionality).

piplmesh/account/backends.py
```diff
@@ -0,0 +1,16 @@
...
+    This backend uses case insensitive username authentication which is not supported by default
+    """    
+    def authenticate(self, username=None, password=None):
+        try:
+            user = auth_models.User.objects.get(username__iexact=username)
+            if user.check_password(password):
+                return user
+        except auth_models.User.DoesNotExist:
+            return None
+
```
mitar: Unnecessary empty lines.

piplmesh/settings.py
mitar: Which editor are you using? In general do not use tabs (but spaces) and set to use 4 spaces everywhere.

aljana: I use spaces too, but the editor (notepad++)'s preferences were set to tabs, I guess that was a problem.

Issue #4, opened 2012-02-26
Display Google search results inside the portal


mitar:
Display Google search results inside the portal and not redirect to the Google itself. In this way we can retain users on our portal and later on also integrate our own results from the portal.

Ads provided by Google for user searches should be displayed so that we can [monetize](https://www.google.com/adsense/static/en_US/WsOverview.html) that we have a Google search form integrated.

mitjap:
If we are to implement this progamatically (and not just including google's page with iframe or someting similar) this API is the best way to do it (http://code.google.com/apis/customsearch/v1/overview.html). Unfortunately free plan is limited by 100 queries per day. Do we proceed with this issue?

mitar:
No, there is also Google Custom Search approach. That API is if you want directly the results and use them for yourself. We do not mind that results are displayed by Google and that Google add its ads.

mitjap:
There are two ways to do this. One way is to do this programmatically as described above, and the second way is by using javascript that handels the search. 

Here are links to both solutions.
http://www.google.com/cse/
http://code.google.com/apis/customsearch/

I gues  javascript solution that emphasizes results on wlan-si.net and PiplMesh and also searches the web would suit as well.

mitar:
CSE also search the web and is not limited to number of queries as I understand:

https://code.google.com/apis/customsearch/docs/js/cselement-devguide.html
https://code.google.com/apis/customsearch/docs/js/rendering.html

I think we would just have two sections in the results. Our own results + Google results.

mitar:
Done with #47.

PR #47, opened 2012-03-05
Pull request for Issue #4

ghost:
Changed previous search engine with new custom search engine, which
displays search result on the portal.

### DIFF
--- a/piplmesh/frontend/static/piplmesh/css/style.css
+++ b/piplmesh/frontend/static/piplmesh/css/style.css
@@ -54,11 +54,11 @@ body {
 	}
 	
 	.result {
-        text-align: left;
-        padding-top: 110px;
-        padding-right: 110px;
-        padding-left: 110px;
-        padding-bottom: 10px;
+		text-align: left;
+		padding-top: 110px;
+		padding-right: 110px;
+		padding-left: 110px;
+		padding-bottom: 10px;
 	}
 	
 	.footer {
--- a/piplmesh/frontend/static/piplmesh/css/style.css
+++ b/piplmesh/frontend/static/piplmesh/css/style.css
@@ -53,6 +53,14 @@ body {
 		position: relative;
 	}
 	
+	.result {
+        text-align: left;
+        padding-top: 110px;
+        padding-right: 110px;
+        padding-left: 110px;
+        padding-bottom: 10px;
+	}
+	
 	.footer {
 		border-top: 1px solid #CCC;
 		padding-top:20px;
--- a/piplmesh/frontend/static/piplmesh/js/query.js
+++ b/piplmesh/frontend/static/piplmesh/js/query.js
@@ -1,3 +1,5 @@
-var query = location.search;
-var url = 'http://www.google.com/search' + query;
-document.getElementById("link").setAttribute("href", url);
+$(document).ready(function () {
+    var query = location.search;
+    var url = 'http://www.google.com/search' + query;
+    $("a#link").prop('href', url);
+});
--- a/piplmesh/frontend/static/piplmesh/js/query.js
+++ b/piplmesh/frontend/static/piplmesh/js/query.js
@@ -1,0 +1,3 @@
+var query = location.search;
+var scrt_var = 'http://www.google.com/search'+query;
+document.getElementById("link").setAttribute("href",scrt_var);
--- a/piplmesh/frontend/static/piplmesh/js/query.js
+++ b/piplmesh/frontend/static/piplmesh/js/query.js
@@ -1,5 +1,0 @@
-$(document).ready(function () {
-    var query = location.search;
-    var url = 'http://www.google.com/search' + query;
-    $("a#link").prop('href', url);
-});
--- a/piplmesh/frontend/static/piplmesh/js/query.js
+++ b/piplmesh/frontend/static/piplmesh/js/query.js
@@ -1,3 +1,3 @@
 var query = location.search;
-var scrt_var = 'http://www.google.com/search'+query;
-document.getElementById("link").setAttribute("href",scrt_var);
+var url = 'http://www.google.com/search' + query;
+document.getElementById("link").setAttribute("href", url);
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -1,7 +1,7 @@
-google.load("search", "1");
+google.load("search", "1", { language : {{ LANGUAGE_CODE }}})
 google.setOnLoadCallback(function () {
     var customSearchOptions = {};
-    var customSearchControl = new google.search.CustomSearchControl('003912915932446183218:zeq20qye9oa', customSearchOptions);
+    var customSearchControl = new google.search.CustomSearchControl({{ SEARCH_ENGINE_UNIQUE_ID }}, customSearchOptions);
     customSearchControl.setResultSetSize(google.search.Search.FILTERED_CSE_RESULTSET);
     var options = new google.search.DrawOptions();
     options.setAutoComplete(true);
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -1,0 +1,9 @@
+google.load('search', '1', {language : 'en'});
+google.setOnLoadCallback(function () {
+    var customSearchOptions = {};
+    var customSearchControl = new google.search.CustomSearchControl('003912915932446183218:zeq20qye9oa', customSearchOptions);
+    customSearchControl.setResultSetSize(google.search.Search.FILTERED_CSE_RESULTSET);
+    var options = new google.search.DrawOptions();
+    options.enableSearchboxOnly("/search");
+    customSearchControl.draw('cse-search-form', options);
+}, true);
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -5,7 +5,7 @@ google.setOnLoadCallback(function () {
     var options = new google.search.DrawOptions();
     options.setAutoComplete(true);
     options.enableSearchboxOnly('/search');
-    customSearchControl.draw('cse-search-form', options);
+    customSearchControl.draw('cse_search_form', options);
 }, true);
 
 
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -4,10 +4,9 @@ google.setOnLoadCallback(function () {
     var customSearchControl = new google.search.CustomSearchControl(search_engine_unique_id, customSearchOptions);
     var options = new google.search.DrawOptions();
     options.setAutoComplete(true);
-    options.enableSearchboxOnly('/search');
+    options.enableSearchboxOnly(search_page_url);
     customSearchControl.draw('cse_search_form', options);
 }, true);
-
 
 $(document).ready(function () {
     var query = location.search;
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -4,6 +4,7 @@ google.setOnLoadCallback(function () {
     var customSearchControl = new google.search.CustomSearchControl('003912915932446183218:zeq20qye9oa', customSearchOptions);
     customSearchControl.setResultSetSize(google.search.Search.FILTERED_CSE_RESULTSET);
     var options = new google.search.DrawOptions();
+    options.setAutoComplete(true);
     options.enableSearchboxOnly("/search");
     customSearchControl.draw('cse-search-form', options);
 }, true);
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -1,4 +1,4 @@
-google.load("search", "1", { language : {{ LANGUAGE_CODE }}})
+google.load("search", "1", { language : {{ LANGUAGE_CODE }}});
 google.setOnLoadCallback(function () {
     var customSearchOptions = {};
     var customSearchControl = new google.search.CustomSearchControl({{ SEARCH_ENGINE_UNIQUE_ID }}, customSearchOptions);
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -1,14 +1,16 @@
 google.load('search', '1', { 'language' : language_code });
 google.setOnLoadCallback(function () {
     var customSearchOptions = {};
-    var imageSearchOptions = {};
-    imageSearchOptions['layout'] = google.search.ImageSearch.LAYOUT_POPOUT;
-    customSearchOptions['enableImageSearch'] = true;
-    customSearchOptions['imageSearchOptions'] = imageSearchOptions;  
     var customSearchControl = new google.search.CustomSearchControl(search_engine_unique_id, customSearchOptions);
-    customSearchControl.setResultSetSize(google.search.Search.FILTERED_CSE_RESULTSET);
     var options = new google.search.DrawOptions();
     options.setAutoComplete(true);
     options.enableSearchboxOnly('/search');
     customSearchControl.draw('cse-search-form', options);
 }, true);
+
+
+$(document).ready(function () {
+    var query = location.search;
+    var open_in_google_link = 'http://www.google.com/search' + query;
+    $('#google_results_link').prop('href', open_in_google_link);
+});
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -1,10 +1,10 @@
-google.load("search", "1", { language : {{ LANGUAGE_CODE }}});
+google.load('search', '1', { language : lang_code });
 google.setOnLoadCallback(function () {
     var customSearchOptions = {};
-    var customSearchControl = new google.search.CustomSearchControl({{ SEARCH_ENGINE_UNIQUE_ID }}, customSearchOptions);
+    var customSearchControl = new google.search.CustomSearchControl(unique_id, customSearchOptions);
     customSearchControl.setResultSetSize(google.search.Search.FILTERED_CSE_RESULTSET);
     var options = new google.search.DrawOptions();
     options.setAutoComplete(true);
-    options.enableSearchboxOnly("/search");
+    options.enableSearchboxOnly('/search');
     customSearchControl.draw('cse-search-form', options);
 }, true);
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -1,7 +1,7 @@
-google.load('search', '1', { language : lang_code });
+google.load('search', '1', { 'language' : language_code });
 google.setOnLoadCallback(function () {
     var customSearchOptions = {};
-    var customSearchControl = new google.search.CustomSearchControl(unique_id, customSearchOptions);
+    var customSearchControl = new google.search.CustomSearchControl(search_engine_unique_id, customSearchOptions);
     customSearchControl.setResultSetSize(google.search.Search.FILTERED_CSE_RESULTSET);
     var options = new google.search.DrawOptions();
     options.setAutoComplete(true);
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -1,6 +1,10 @@
 google.load('search', '1', { 'language' : language_code });
 google.setOnLoadCallback(function () {
     var customSearchOptions = {};
+    var imageSearchOptions = {};
+    imageSearchOptions['layout'] = google.search.ImageSearch.LAYOUT_POPOUT;
+    customSearchOptions['enableImageSearch'] = true;
+    customSearchOptions['imageSearchOptions'] = imageSearchOptions;  
     var customSearchControl = new google.search.CustomSearchControl(search_engine_unique_id, customSearchOptions);
     customSearchControl.setResultSetSize(google.search.Search.FILTERED_CSE_RESULTSET);
     var options = new google.search.DrawOptions();
--- a/piplmesh/frontend/static/piplmesh/js/search_bar.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_bar.js
@@ -1,4 +1,4 @@
-google.load('search', '1', {language : 'en'});
+google.load("search", "1");
 google.setOnLoadCallback(function () {
     var customSearchOptions = {};
     var customSearchControl = new google.search.CustomSearchControl('003912915932446183218:zeq20qye9oa', customSearchOptions);
--- a/piplmesh/frontend/static/piplmesh/js/search_page.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_page.js
@@ -3,7 +3,9 @@ google.setOnLoadCallback(function () {
     var customSearchOptions = {};
     var customSearchControl = new google.search.CustomSearchControl(search_engine_unique_id, customSearchOptions);
     customSearchControl.setResultSetSize(google.search.Search.LARGE_RESULTSET);
-    customSearchControl.draw('cse');
+    var options = new google.search.DrawOptions();
+    options.enableSearchResultsOnly();
+    customSearchControl.draw('cse', options);
     function parseParamsFromUrl() {
         var params = {};
         var parts = window.location.search.substr(1).split('&');
--- a/piplmesh/frontend/static/piplmesh/js/search_page.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_page.js
@@ -8,12 +8,12 @@ google.setOnLoadCallback(function () {
         var params = {};
         var parts = window.location.search.substr(1).split('&');
         for (var i = 0; i < parts.length; i++) {
-        var keyValuePair = parts[i].split('=');
-        var key = decodeURIComponent(keyValuePair[0]);
-        params[key] = keyValuePair[1] ? decodeURIComponent(keyValuePair[1].replace(/\+/g, ' ')) : keyValuePair[1];
+            var keyValuePair = parts[i].split('=');
+            var key = decodeURIComponent(keyValuePair[0]);
+            params[key] = keyValuePair[1] ? decodeURIComponent(keyValuePair[1].replace(/\+/g, ' ')) : keyValuePair[1];
+        }
+        return params;
     }
-    return params;
-}
 
 var urlParams = parseParamsFromUrl();
 var queryParamName = "q";
--- a/piplmesh/frontend/static/piplmesh/js/search_page.js
+++ b/piplmesh/frontend/static/piplmesh/js/search_page.js
@@ -1,6 +1,10 @@
 google.load('search', '1', { 'language' : language_code });
 google.setOnLoadCallback(function () {
     var customSearchOptions = {};
+    var imageSearchOptions = {};
+    imageSearchOptions['layout'] = google.search.ImageSearch.LAYOUT_POPOUT;
+    customSearchOptions['enableImageSearch'] = true;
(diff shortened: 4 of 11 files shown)


mitar:
Don't use tabs, use only spaces for indentation, 4 of them for one level of indentation.

mitar:
Hm. It would be better to make a separate simple view (like home), nest it under `/search` and then when a user makes a query, it is send to that page, where we will display some our results (in later milestone) and some Google's results. So do not put content in the current page, but go to the special page. And then you can make a simple hard-coded example of our results and Google's results. Also show any adds Google could maybe want to display.

ghost:
Where to put this /search folder in /frontend or in /piplmesh? So i should put the view.py for search page in that folder if I understand correctly.

mitar:
No no. No source directory, but in `urls.py` you specify new rule for `/search` URL, which calls your view, which you put into `frontend/views.py`.

ghost:
I found another solution I edited only urls.py file, but I think that this should be OK. It is working as expected and it is already prepared for searching on piplmesh in later versions, and for google AdSense or something similar.

piplmesh/frontend/static/piplmesh/js/search_page.js
```diff
@@ -0,0 +1,29 @@
+google.load('search', '1', {language : 'en'});
```
ghost: I think that here the language should be set to slovenian, not to be chosen by the user since this will bi portal in Slovenia so it would be logical that the search engine search primary slovenian pages. What do you think?

mitar: Does language selection of Google interface (as I understand) really influence also the search results. Is there a way to configure the results based on the location? Then we could configure user to configure this in her user profile? So language and Google results.

ghost: I will check if is possible to configure the results based on the location and I will get back to you with my findings.

ghost: I tested both languages, but there is no difference if i change the language, not by search result, not by search engine language.

mitar: OK, then take that value from locale ([current language](https://docs.djangoproject.com/en/dev/topics/i18n/translation/#other-tags) Django is rendering the template in). It seems that for API it supports much less languages than otherwise, thought.

Also implement local search as shown in [this example](https://developers.google.com/web-search/docs/).

ghost: I found that if you don't specify the language, google uses current language that you use, so one problem fixed.

About that local search, since Google Web Search API is deprecated I used Google custom search engine instead so local search option isn't available here. One solution is that I rewrite not using google custom search engine, or we specify search options. This search engine can be fully customized and it is also already prepared  for AdSense. So if you would like later to use different search engine specifications or AdSense or something, you just change the Search engine unique ID with your own.

Please check if this search engine is customizable enough for our needs. more info on: http://www.google.com/cse/

piplmesh/urls.py
```diff
@@ -7,7 +8,7 @@
 
 urlpatterns = patterns('',
     url('^$', frontend_views.HomeView.as_view(), name='home'),
-    
+    url(r'^search/$', frontend_views.SearchView.as_view(), name='search'),
```
mitar: Trailing `/`.

mitar:
So you are using [this](https://code.google.com/apis/customsearch/docs/js/rendering.html)?

I missed that that API is deprecated. OK, then let's do it with custom search.

piplmesh/frontend/static/piplmesh/js/search_bar.js
```diff
@@ -0,0 +1,10 @@
+google.load("search", "1");
+google.setOnLoadCallback(function () {
+    var customSearchOptions = {};
+    var customSearchControl = new google.search.CustomSearchControl('003912915932446183218:zeq20qye9oa', customSearchOptions);
```
mitar: Move this to settings.

ghost:
Last commit is not working! My local server crashed or something, so I got wrong behavior.

piplmesh/frontend/static/piplmesh/js/search_bar.js
```diff
@@ -0,0 +1,10 @@
+google.load("search", "1", { language : {{ LANGUAGE_CODE }}});
```
mitar: Hm, since when does Django renders static JavaScript files?

You have to put those constants into the template (store them into global JavaScript variables) and after that load those external JavaScript files.

piplmesh/settings.py
```diff
@@ -173,3 +173,5 @@
         },
     }
 }
+
+SEARCH_ENGINE_UNIQE_ID = '003912915932446183218:zeq20qye9oa'
```
mitar: Please comment a bit this value: describe how you can get your own, what are current settings of this custom search engine.

piplmesh/urls.py
```diff
@@ -7,7 +8,7 @@
 
 urlpatterns = patterns('',
     url('^$', frontend_views.HomeView.as_view(), name='home'),
-    
+    url(r'^search/$', frontend_views.SearchView.as_view(), name='/search'),
```
mitar: Remove trailing `/`. Not in name, but in regular expression at the beginning!

piplmesh/frontend/templates/search.html
```diff
@@ -0,0 +1,8 @@
+<form>
+    <input type="button" value="Pipl Mesh" onclick="window.location.href='/'"> 
+</form>
+<hr />
+<div id="cse">Loading</div>
+    <script src="https://www.google.com/jsapi" type="text/javascript"></script>
+    <script type="text/javascript" src="{{ STATIC_URL }}piplmesh/js/search_page.js"></script>
+<link rel="stylesheet" href="https://www.google.com/cse/style/look/default.css" type="text/css" />
```
mitar: Align.

piplmesh/frontend/static/piplmesh/js/search_page.js
```diff
@@ -0,0 +1,29 @@
...
+            params[key] = keyValuePair[1] ? decodeURIComponent(keyValuePair[1].replace(/\+/g, ' ')) : keyValuePair[1];
+        }
+        return params;
+    }
+
+var urlParams = parseParamsFromUrl();
+var queryParamName = "q";
+if (urlParams[queryParamName]) {
+    customSearchControl.execute(urlParams[queryParamName], null,
+    {'oq': urlParams['oq'],
```
mitar: Format in the way to move `{` one line up.

mitar: And then indent the content of the dict one indentation more to the right.

piplmesh/frontend/static/piplmesh/js/search_page.js
```diff
@@ -0,0 +1,29 @@
...
+        for (var i = 0; i < parts.length; i++) {
+            var keyValuePair = parts[i].split('=');
+            var key = decodeURIComponent(keyValuePair[0]);
+            params[key] = keyValuePair[1] ? decodeURIComponent(keyValuePair[1].replace(/\+/g, ' ')) : keyValuePair[1];
+        }
+        return params;
+    }
+
+var urlParams = parseParamsFromUrl();
+var queryParamName = "q";
```
mitar: Use `'` for constant strings.

piplmesh/frontend/static/piplmesh/js/search_bar.js
ghost: Yeah, my local server crashed or something, and then i got the wrong behavior of the search engine, so I kinda thought that is possible.

ghost:
I think now everything is fixed as it should be.

piplmesh/frontend/static/piplmesh/js/search_page.js
```diff
@@ -0,0 +1,29 @@
...
+            var key = decodeURIComponent(keyValuePair[0]);
+            params[key] = keyValuePair[1] ? decodeURIComponent(keyValuePair[1].replace(/\+/g, ' ')) : keyValuePair[1];
+        }
+        return params;
+    }
+
+var urlParams = parseParamsFromUrl();
+var queryParamName = 'q';
+if (urlParams[queryParamName]) {
+    customSearchControl.execute(urlParams[queryParamName], null,{
```
mitar: Space after comma.

piplmesh/frontend/static/piplmesh/js/search_bar.js
```diff
@@ -0,0 +1,10 @@
+google.load('search', '1', { language : lang_code });
```
mitar: Dict keys put convert into strings:

```
google.load('search', '1', {'language': lang_code});
```

piplmesh/frontend/static/piplmesh/js/search_page.js
```diff
@@ -0,0 +1,29 @@
+google.load('search', '1', { language : lang_code });
```
mitar: Dict keys put convert into strings.

piplmesh/frontend/static/piplmesh/js/search_page.js
```diff
@@ -0,0 +1,29 @@
...
+var urlParams = parseParamsFromUrl();
+var queryParamName = 'q';
+if (urlParams[queryParamName]) {
+    customSearchControl.execute(urlParams[queryParamName], null,{
+        'oq': urlParams['oq'],
+        'aq': urlParams['aq'],
+        'aqi': urlParams['aqi'],
+        'aql': urlParams['aql'],
+        'gs_sm': urlParams['gs_sm'],
+        'gs_upl': urlParams['gs_upl']});
```
mitar: Those last brackets split into separate lines:

```
        'gs_upl': urlParams['gs_upl'],
    });
```

piplmesh/frontend/static/piplmesh/js/search_page.js
```diff
@@ -0,0 +1,29 @@
...
+        var parts = window.location.search.substr(1).split('&');
+        for (var i = 0; i < parts.length; i++) {
+            var keyValuePair = parts[i].split('=');
+            var key = decodeURIComponent(keyValuePair[0]);
+            params[key] = keyValuePair[1] ? decodeURIComponent(keyValuePair[1].replace(/\+/g, ' ')) : keyValuePair[1];
+        }
+        return params;
+    }
+
+var urlParams = parseParamsFromUrl();
```
mitar: You should indent the whole block from here four spaces to the right, no?

piplmesh/frontend/templates/header.html
```diff
@@ -5,12 +5,14 @@
...
-                <div>
-                    <img width="100" height="30" alt="{{ search_engine }} Search Engine" src="{{ STATIC_URL }}piplmesh/images/{{ search_engine_logo }}" />
-                    <input type="text" name="q" size="25" maxlength="255" value="" />
-                    <input type="submit" value="{{ search_engine }} Search" />
-                </div>
-            </form>
+            <div id="cse-search-form"></div>
+            <script src="https://www.google.com/jsapi" type="text/javascript"></script>
+            <script type="text/javascript">
+                var lang_code = '{{ LANGUAGE_CODE }}';
```
mitar: Good, but please name variables the same as it is in Django so that it is easier to understand what they are:

```
var language_code = '{{ LANGUAGE_CODE }}';
```

piplmesh/settings.py
```diff
@@ -173,3 +173,11 @@
         },
     }
 }
+
+# You can set up your own custom search engine on: http://www.google.com/cse/
+# just register with you google account and crate new search engine.
+# When you create new search engine, switch uniqe id with your own and new settings will apply,
+# unles you change something that needs to be changed in the code, but if you do that
+# you will be warrned before you will make the change.
```
mitar: Typo: _warrned_.

mitar: I don't understand who will warn who?

piplmesh/frontend/views.py
```diff
@@ -1,4 +1,5 @@
 from django.views import generic as generic_views
+from django.conf import *
```
mitar: Import only modules, not `*`. In this case, import like:

```
```

piplmesh/urls.py
```diff
@@ -1,4 +1,5 @@
 from django.conf.urls.defaults import patterns, include, url
+from django.views.generic.simple import *
```
mitar: Import only modules.

piplmesh/urls.py
```diff
@@ -7,7 +8,7 @@
 
 urlpatterns = patterns('',
     url('^$', frontend_views.HomeView.as_view(), name='home'),
-    
+    url(r'^search', frontend_views.SearchView.as_view(), name='search'),
     url(r'^admin/doc/', include('django.contrib.admindocs.urls')),
```
mitar, line 12: One empty line inbetween.

mitar:
Added some comments.

piplmesh/frontend/static/piplmesh/js/search_bar.js
ghost: Fixed.

piplmesh/frontend/static/piplmesh/js/search_page.js
ghost: Fixed.

piplmesh/frontend/static/piplmesh/js/search_page.js
ghost: Fixed

piplmesh/frontend/static/piplmesh/js/search_page.js
ghost: Fixed.

piplmesh/frontend/static/piplmesh/js/search_page.js
ghost: Fixed.

I am just not sure about that comma, it works with it so I putt there ilke you did.

piplmesh/frontend/templates/header.html
ghost: Fixed, variables renamed.

piplmesh/frontend/views.py
ghost: Fixed.

piplmesh/settings.py
ghost: Typo fixed. Also updated description little bit, hope it is more understandable now.

piplmesh/urls.py
ghost: Removed that import since it was redundant.

piplmesh/urls.py
ghost, line 12: Empty line between HomeView and SearchView or empty line between SearchView and admin/doc/ ?

piplmesh/frontend/templates/search.html
```diff
@@ -0,0 +1,12 @@
```
mitar, line 0: Why this template does not extend our base templates?

ghost, line 0: I tried to add that, but then search page didn't work. So I thought that i should not use it on the new page.

mitar, line 0: But when you open now `/search` then this is probably not a valid HTML page?

ghost, line 0: I didn't have any problem. I fixed that, please check if it is OK now.

piplmesh/frontend/templates/search.html
```diff
@@ -0,0 +1,18 @@
...
+<div id="cse"></div>
+<script src="https://www.google.com/jsapi" type="text/javascript"></script>
+<script type="text/javascript">
+    var language_code = '{{ LANGUAGE_CODE }}';
+    var search_engine_unique_id = '{{ SEARCH_ENGINE_UNIQUE_ID }}';
+</script>
+<script type="text/javascript" src="{{ STATIC_URL }}piplmesh/js/search_page.js"></script>
+<link rel="stylesheet" href="https://www.google.com/cse/style/look/default.css" type="text/css" />
+<hr />
+{% include "footer.html" %}
```
ghost: Is it OK if I use footer also on the search result page?

mitar: In fact I would extend `plain.html` and have both header and footer. (And fill in the search box of the header the same query as it is passed to the view.)

ghost: I think that is not OK, because then we would have two search bars (one in header and one default on the result search page).

mitar: This is not a problem, many sites have like that. Or you can hide it from the header if you want. But leave the rest of the header. Or we can only have the header one and display only results bellow.

ghost: OK, then we will have only header search bar. And since we will have header, I would recommend that click on the logo would take the user "home", because it look funny if we have another button to return user back home. And also css should be edited, because css class from google is overrun by our class.

mitar: I agree with everything you proposed. I don't see the problem if we override Google's CSS? In any way we want that also Google's results look like our web page? But yes, feel free to fix the CSS too.

ghost: Now I added logo link, removed default search bar, so now we only have header search bar. But I have trouble with CSS, if i try to use my own class search results are covered by header, but if I use class that is already created in css file I used "main" positioning is ok. I can't figure out what is the problem, since I am not very skilled in CSS, so any hint would be helpful.

EDIT:
inline CSS also works as expected.

mitar: Please do not use inline CSS. Use Firebug to see what CSS rules are applied to each element and which CSS rules are overridden by which other CSS rules.

You can also ask other members of your team to help you.

ghost: I won't use it I was just testing what was the problem. I will consult other team members.

mitar: If with inline CSS things work, this often means that something overrides your CSS. Use Firebug to discover who and where. There is also a switch for Google custom search not to include their CSS, if I am not mistaken? Maybe their CSS is added after yours and this is why it overrides yours?

ghost: Wil check that thanks.

EDIT: 
found a solution, safari doesn't refresh css if you don't entirely shut down application, when I used firefox and Firebug. everything was OK, after I restarted safari changes were shown there too.

ghost: Do we want to support image search?

mitar: Yes. In fact, we would like to make user uses our search instead of the Google's official one as much as possible. But we can also be friendly and put somewhere a link like "Use google.com", which then opens the query in official search.

ghost: OK I included image search on our page, so I will add also a link with google.com somewhere on the page. Probably best if it is under search results?

ghost: I added a link under results, to see query results on official google page.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -53,6 +53,14 @@ body {
 		position: relative;
 	}
 	
+	.result {
+        text-align: left;
```
mitar: Please keep indenting style consistent. Here is much more indented than elsewhere.

mitar:
I think it is better that link to google.com is under the search box. So the idea is that user will not have problems entering the search query in our search box, because she will know that it is one click away from switching to her Google search box if the results are not the way she wants. But this is not so important now. We can play with this later.

piplmesh/frontend/static/piplmesh/js/query.js
```diff
@@ -0,0 +1,3 @@
```
mitar: Why is this a separate file?

piplmesh/frontend/static/piplmesh/js/query.js
```diff
@@ -0,0 +1,3 @@
+var query = location.search;
+var scrt_var = 'http://www.google.com/search'+query;
```
mitar: Name of the variable is uncomprehensible. I cannot understand what this string is? Please use spaces around `+`.

piplmesh/frontend/static/piplmesh/js/query.js
```diff
@@ -0,0 +1,3 @@
...
+document.getElementById("link").setAttribute("href",scrt_var);
```
mitar: Space after `,`. Use jQuery for DOM manipulation (because we have it already loaded, so that everything is consistent).

piplmesh/frontend/static/piplmesh/js/query.js
mitar: I think the correct way is to use jQuery onready event handler and do DOM transforms there. Do not simply load JavaScript file after the DOM you want to change. (I prefer having all JavaScript in header section and then with event handlers attach to/manipulate DOM.)

ghost: Where should I put it then if not in the separate file? You probably meant "head" section? OK i will use jQuery to "attach" script files to DOM

mitar: I would put all search-related code into one external file. Now you have three (`query.js` and `search_bar.js` and `search_page.js`). Why? OK, I understand for search_page and search_bar (because page has additional code (but even there I see some code duplication I don't like), but why additional `query.js`?

piplmesh/frontend/static/piplmesh/js/query.js
```diff
@@ -0,0 +1,5 @@
+$(document).ready(function () {
+    var query = location.search;
+    var url = 'http://www.google.com/search' + query;
+    $("a#link").prop('href', url);
```
mitar: Use `'` for constant strings. `prop`? Is not `attr`? `#` is an ID selector - ID is unique so `a` there is redundant - it cannot be anything else.

But maybe give the element a more specific name ID (`"open-in-google-link"`, hm, should we use dashes or underscores?).

piplmesh/frontend/static/piplmesh/js/query.js
ghost: OK. I will try to put al of the code in one file.

piplmesh/frontend/static/piplmesh/js/query.js
ghost: I read that in newer versions of jQuery prop is used for DOM element attributes, but I can use attr if you like. 

I would use underscores like "open_in_google_link" because it is more readable then dashes if you agree.

piplmesh/frontend/static/piplmesh/js/query.js
ghost: I removed third file "query.js" and merge it with "search_bar.js". I also removed that duplicated code from "search_bar.js" since it was redundant so there is now almost no duplicated code, but i couldn't merge together  "search_bar.js" and "search_page.js"  I tried 10 different solutions but none of them worked.

Also moved "see on google" link under search bar.

piplmesh/frontend/static/piplmesh/js/query.js
mitar: Interesting this about `.prop`. I didn't know that. ;-)

I agree with dashes, just make sure that also other IDs are consistent with this style.

ghost: I didn't knew that either before. :) I changed ID's so now are consistent and also fixed what was wrong. I think that now everything is OK.

piplmesh/frontend/static/piplmesh/js/search_bar.js
```diff
@@ -0,0 +1,16 @@
+google.load('search', '1', { 'language' : language_code });
+google.setOnLoadCallback(function () {
+    var customSearchOptions = {};
+    var customSearchControl = new google.search.CustomSearchControl(search_engine_unique_id, customSearchOptions);
+    var options = new google.search.DrawOptions();
+    options.setAutoComplete(true);
+    options.enableSearchboxOnly('/search');
```
mitar: Hm, this URL probably should not be hardcoded, but reversed from `urls.py` (so you can set it in the same manner as other configuration variables).

piplmesh/frontend/static/piplmesh/js/search_bar.js
```diff
@@ -0,0 +1,16 @@
...
+    customSearchControl.draw('cse_search_form', options);
+}, true);
+
```
mitar, line 10: Unnecessary empty line.

piplmesh/frontend/templates/base.html
```diff
@@ -4,6 +4,7 @@
         <title>{% block fulltitle %}{% block title %}{% endblock %} – wlan slovenija{% endblock %}</title>
         <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
         <link rel="stylesheet" href="{{ STATIC_URL }}piplmesh/css/style.css" type="text/css" media="screen" />
+        <script src="http://code.jquery.com/jquery-latest.js"></script>
```
mitar: Please use own version of jQuery. We want predictability. And also that things work on localhost without Internet connectivity. And also we want that things work under HTTPS (and your code above would trigger warnings).

piplmesh/frontend/static/piplmesh/js/search_page.js
```diff
@@ -0,0 +1,36 @@
...
+    var imageSearchOptions = {};
+    imageSearchOptions['layout'] = google.search.ImageSearch.LAYOUT_POPOUT;
+    customSearchOptions['enableImageSearch'] = true;
+    customSearchOptions['imageSearchOptions'] = imageSearchOptions;  
+    var customSearchControl = new google.search.CustomSearchControl(search_engine_unique_id, customSearchOptions);
+    customSearchControl.setResultSetSize(google.search.Search.LARGE_RESULTSET);
+    var options = new google.search.DrawOptions();
+    options.enableSearchResultsOnly();
+    customSearchControl.draw('cse', options);
+    function parseParamsFromUrl() {
```
mitar, line 13: I really do not understand why is this necessary. Is there really not any other way? It really seems strange to me that we need to define such function ourselves.

piplmesh/frontend/templates/header.html
```diff
@@ -1,16 +1,23 @@
+{% load i18n %}
+
 <div class="container">
     <div class="header">
         <div class="logo">
-            <img src="{{ STATIC_URL }}piplmesh/images/logo.png" alt="PiplMesh" />
+            <a href="/">
+                <img src="{{ STATIC_URL }}piplmesh/images/logo.png" alt="PiplMesh" />
+            </a>
             <h1>Pipl Mesh</h1>
```
mitar: PiplMesh not Pipl Mesh.

piplmesh/frontend/templates/search.html
```diff
@@ -0,0 +1,14 @@
+{% extends "plain.html" %}
+{% load i18n %}
+
+{% block fulltitle %}wlan slovenija{% endblock %}
```
mitar: Why are you overriding full title? Why not set something for title? Like _Search_?

piplmesh/frontend/templates/search.html
```diff
@@ -0,0 +1,14 @@
+{% extends "plain.html" %}
+{% load i18n %}
```
mitar, line 3: Please make an empty line before this line.

piplmesh/frontend/views.py
```diff
@@ -1,4 +1,5 @@
 from django.views import generic as generic_views
+from django.conf import settings
```
mitar, line 1: Alphabetical order.

piplmesh/frontend/static/piplmesh/js/search_page.js
ghost, line 13: Well actually this function was auto generated by Google CSE, I also checked if there is any way to make this shorter, but there is no inbuilt function for parsing parameters from URL (or at least I din't find one).

piplmesh/frontend/static/piplmesh/js/search_bar.js
ghost: Fixed.

piplmesh/frontend/static/piplmesh/js/search_bar.js
ghost, line 10: Fixed.

piplmesh/frontend/templates/base.html
ghost: Fixed, aded new file containing jQuery library.

piplmesh/frontend/templates/search.html
ghost: Fixed.

piplmesh/frontend/templates/search.html
ghost, line 3: Fixed.

piplmesh/frontend/views.py
ghost, line 1: Fixed.

Issue #51, opened 2012-03-06
Design - frontend like wlan slovenia


NejcZupec:
Basic frontend layout from wlan slovenia site.

mitar:
I agree. Where is code? ;-)

mitar:
I found it. ;-)

mitar:
I [made public](http://dev.wlan-si.net/browser/homepage) current code for wlan slovenija homepage. It could be easier to work on this now.

mitar:
Done with #52.

mitar:
For design ideas I recommend you also check [nodewatcher](https://nodes.wlan-si.net/) web page (for example how to design login).

PR #52, opened 2012-03-06
Pull request for issue #51

NejcZupec:
I've added a completely new css file with logo and background image from wlan slovenia site.


piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -1,64 +1,75 @@
...
+
+/* -------------------- HEADER --------------------- */
+#header {
+    width: 960px;
+    height: 150px;
+    margin-left: auto;
+    margin-right: auto;
+}
+
+#header #language_chooser  {
```
mitar, line 27: This is useless. `#` signals page-unique ID of the element. There is completely no need to tell that it is a child of some other element.

PR #54, opened 2012-03-07
Issue 7

dcrystalj:


mitar:
Is this finalized code?

dcrystalj:
I'm not sure. It does what it's written in issue#7- after user click allow you can get information about user. But I'm not really sure how to access that data later. Also I still have no idea if it's correct to put script code directly in .html file like I did here in django.

mitar:
No. Move code to external files and made Facebook app id configurable through `settings.py`.

piplmesh/frontend/templates/header.html
```diff
@@ -13,4 +13,24 @@
                 </div>
             </form>
         </div>
-    </div>
+		<div id="fb-root"></div>
```
mitar: Are we indenting with four or two spaces? Please spell out whole names, be verbose!

mitar:
Closing in favor of other pull requests.

PR #58, opened 2012-03-11
Issue #4 rev 2 (team 4)

agiz:


piplmesh/frontend/templates/header.html
```diff
@@ -5,12 +7,19 @@
...
         </div>
         <div class="search">
-            <form method="get" action="http://www.google.com/search">
-                <div>
-                    <img width="100" height="30" alt="{{ search_engine }} Search Engine" src="{{ STATIC_URL }}piplmesh/images/{{ search_engine_logo }}" />
-                    <input type="text" name="q" size="25" maxlength="255" value="" />
-                    <input type="submit" value="{{ search_engine }} Search" />
-                </div>
-            </form>
+            <div id="cse-search-form" style="width: 100%;">Loading</div>
```
mitar: Don't use inline CSS.

Translatable string is here.

piplmesh/frontend/templates/header.html
```diff
@@ -5,12 +7,19 @@
...
+            <script src="//www.google.com/jsapi" type="text/javascript"></script>
+            <script type="text/javascript"> 
+              google.load('search', '1', {language : '{{ LANGUAGE_CODE }}'});
+              google.setOnLoadCallback(function() {
+                var customSearchOptions = {};  var customSearchControl = new google.search.CustomSearchControl(
```
mitar: Please move each variable definition into a separate line.

piplmesh/frontend/templates/header.html
```diff
@@ -5,12 +7,19 @@
...
+                  '{{ SEARCH_ENGINE_UNIQUE_ID }}', customSearchOptions);
+                customSearchControl.setResultSetSize(google.search.Search.LARGE_RESULTSET);
+                var options = new google.search.DrawOptions();
+                options.setSearchFormRoot('cse-search-form');
+                customSearchControl.draw( 'cse', options );
```
mitar: Unnecessary spaces around function arguments.

piplmesh/frontend/templates/home.html
```diff
@@ -4,6 +4,7 @@
 
 {% block content %}
 <div class="main">
+    <div id="cse"></div>
```
mitar, line 7: Shouldn't this be a part of a special page? Or a header?

mitar, line 7: Accessible for example under `/search`?

piplmesh/frontend/templates/header.html
agiz: Done.

piplmesh/frontend/templates/header.html
agiz: Done.

piplmesh/frontend/templates/header.html
agiz: What shall be done with java scripts and stylesheets? I changed stylesheet path to load from static. I left google jsapi to load from default URL. Line 13 and 20 are inconsistent. Should I wrap lines 13-17 in a single line or unwrap line 20? If later what indent to use?

piplmesh/frontend/templates/home.html
agiz, line 7: I don't follow. Don't we want to display results in the current page without any redirections? This way the results are shown above any main content. To restore to previous page just close search results (within main content). Would you rather have it toggle between current page and search results?

piplmesh/frontend/templates/header.html
mitar: I am just saying that line 20 has spaces after `(` and before `)`. Not that you have to split arguments in separate lines. ;-)

piplmesh/frontend/templates/home.html
mitar, line 7: OK, you can also overlay, but then you have to change URL in some way (for example using URL anchor) so that we user can bookmark/copy link to results. Copy+pasting the URL should open the same page content.

For search I do not think we really need overlay. (Facebook does not have it. And in general I do not really like overlay. Do you?)

But it is OK as it is. It is good to have different approaches.

mitar:
Closing in favor of #47.

PR #59, opened 2012-03-14
Updating active users via Ajax

byteme:
1) Extended User model with UserProfile (added last_active field)
2) Added JSON view response with list of active users (user objects)
3) Added new template tag (active_users) which renders javascript for updating active users
4) Added active user update time and user threshold to settings, enabled user profile
5) Added active_users script and jquery to plain.html


byteme:
I'm not sure if I'm on correct path.

Basically what this script does is do ajax request (using jquery) every (5 seconds - configurable in settings) and request active user list (based on user threshold). If user has active session his last_active profile field is update to datetime.now() and view returns JSON response.

mitar:
Interesting way but we said that we will use HTTP push and not polling.

PR #60, opened 2012-03-21
Updated issue 7 from team 2

Irrialite:


piplmesh/account/views.py
```diff
@@ -4,57 +4,84 @@
 from django.conf import settings
 from django.contrib import auth
 from django.core.urlresolvers import reverse
-from django.shortcuts import render_to_response, redirect
-from django.template import context
+from django.views import generic as generic_views
+from django.views.generic.simple import redirect_to
```
mitar: Import only modules.

piplmesh/account/views.py
```diff
@@ -4,57 +4,84 @@
...
-    return render_to_response('registration.html', data, context_instance=context.RequestContext(request))
+        return context
 
-def facebook_login(request):
+    # Overridden method
+    def get(self, request, *args, **kwargs):
+        context = self.get_context_data(**kwargs)
+        if request.user.is_authenticated():
+            return redirect_to(request, url='../', permanent=False)
+        if request.method == 'POST':
```
mitar: Doesn't this go to `post` method?

piplmesh/account/views.py
```diff
@@ -4,57 +4,84 @@
...
 from django.core.urlresolvers import reverse
-from django.shortcuts import render_to_response, redirect
-from django.template import context
+from django.views import generic as generic_views
+from django.views.generic.simple import redirect_to
 
 from piplmesh.account import forms
 
-def registration_view(request):
+class RegistrationView(generic_views.TemplateView):
```
mitar: Probably you want `ProcessFormView` here?

piplmesh/account/views.py
```diff
@@ -4,57 +4,84 @@
...
+        if request.method == 'POST':
+            form = forms.RegistrationForm(request.POST)
+            if form.is_valid():
+                username, password = form.save()
+                new_user = auth.authenticate(username=username, password=password)
+                auth.login(request, new_user)
+                return redirect_to(request, url='../', permanent=False)
+        return self.render_to_response(context)
+     
+class FacebookLoginView(generic_views.TemplateView):
```
mitar: You probably want here `RedirectView`?

piplmesh/account/views.py
```diff
@@ -4,57 +4,84 @@
...
-    return http.HttpResponseRedirect(settings.FACEBOOK_LOGOUT_REDIRECT)
+    # Overridden method
+    def get(self, request, *args, **kwargs):
+        context = super(FacebookLogoutView, self).get_context_data(**kwargs)
+        auth.logout(request)
+        url_logout = (settings.FACEBOOK_LOGOUT_REDIRECT).replace("%", "%%")
+        return redirect_to(request, url=url_logout, permanent=False)   
 
-def facebook_callback(request):
+class FacebookCallbackView(generic_views.TemplateView):
```
mitar: You probably want here `RedirectView`?

piplmesh/account/views.py
Irrialite: It could go there but all the post method does is it calls get(x....,) with same args.

mitar: You mean default implementation does that?

mitar: Check `ProcessFormView` first!

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
 from django.conf import settings
 from django.contrib import auth
 from django.core.urlresolvers import reverse
-from django.shortcuts import render_to_response, redirect
-from django.template import context
+from django.views import generic as generic_views
+from django.views.generic import simple
+from django.views.generic import edit as form_views
```
mitar: You can combine last two into one line. Also do `edit as edit_views`.

piplmesh/urls.py
```diff
@@ -4,6 +4,8 @@
 from piplmesh.frontend import views as frontend_views
 
 from django.contrib import admin
+from django.conf import settings
```
mitar: `f` is before `t` in the alphabet. ;-)

piplmesh/urls.py
```diff
@@ -4,6 +4,8 @@
 from piplmesh.frontend import views as frontend_views
```
mitar, line 7: Our imports should be grouped after Django imports (first built-in modules, then Django, then other libraries, then our imports).

piplmesh/settings.py
```diff
@@ -108,6 +108,7 @@
 )
 
 TEMPLATE_CONTEXT_PROCESSORS = (
+    'piplmesh.context_processors.global_vars',
```
mitar: Put this as the last context processor so that we can override values if needed.

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+
+    permanent = False
+    url = settings.FACEBOOK_LOGIN_REDIRECT
     
-    code = request.GET['code']
-    user = auth.authenticate(token=code, request=request)
-    auth.login(request, user)
-    return http.HttpResponseRedirect(settings.FACEBOOK_LOGIN_REDIRECT)
\ No newline at end of file
+    # Overridden method
```
mitar: Such comments always looked useless to me. ;-)

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+    def get(self, request, *args, **kwargs):
+        code = request.GET['code']
+        user = auth.authenticate(token=code, request=request)
```
mitar: I like to have variables named the same as function arguments. Can you change `code` to `token`?

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+    permanent = False
+    url = settings.FACEBOOK_LOGIN_REDIRECT
     
-    code = request.GET['code']
-    user = auth.authenticate(token=code, request=request)
-    auth.login(request, user)
-    return http.HttpResponseRedirect(settings.FACEBOOK_LOGIN_REDIRECT)
\ No newline at end of file
+    # Overridden method
+    def get(self, request, *args, **kwargs):
```
mitar, line 90: Why are you overriding `get` and not `get_redirect_url`?

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+        if url:
+            if self.permanent:
+                return http.HttpResponsePermanentRedirect(url)
+            else:
+                return http.HttpResponseRedirect(url)
+        else:
+            logger.warning('Gone: %s' % self.request.path,
+                        extra={
+                            'status_code': 410,
+                            'request': self.request
```
mitar: Comma at the end.

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+        auth.login(request, user)
+        url = self.get_redirect_url(**kwargs)
+        if url:
+            if self.permanent:
+                return http.HttpResponsePermanentRedirect(url)
+            else:
+                return http.HttpResponseRedirect(url)
+        else:
+            logger.warning('Gone: %s' % self.request.path,
+                        extra={
```
mitar: I put this into the same line as the rest (one line above).

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+                            'status_code': 410,
+                            'request': self.request
+                        })
+            return http.HttpResponseGone() 
 
-def facebook_logout(request):
+class FacebookLogoutView(generic_views.RedirectView):
```
mitar, line 70: Same comments as bellow.

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
     """
 
-    if request.user.is_authenticated():
-        return redirect('home')
-    form = forms.RegistrationForm()
-    if request.method == 'POST':
-        form = forms.RegistrationForm(request.POST)
+    template_name = 'registration.html'
+    form_class = forms.RegistrationForm
+    success_url = '../'
```
mitar: ??? Don't hard-code URLs?

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+        
+    def get_context_data(self, **kwargs):
+        context = super(RegistrationView, self).get_context_data(**kwargs)
+        form = forms.RegistrationForm()
+        
+        context.update({
+            'form': form
```
mitar: This is added automatically.

mitar: https://docs.djangoproject.com/en/dev/ref/class-based-views/#django.views.generic.edit.FormMixin.get_context_data

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+        })
+
+        return context
+
+    # Overridden method
```
mitar: Useless comment. (For me IDE tells me this.)

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+    def get(self, request, *args, **kwargs):
+        if request.user.is_authenticated():
+            return simple.redirect_to(request, url='../', permanent=False)
+        form_class = self.get_form_class()
+        form = self.get_form(form_class)
+        return self.render_to_response(self.get_context_data(form=form))
+
+    # Overridden method
```
mitar: Useless comment. (For me IDE tells me this.)

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+        context.update({
+            'form': form
+        })
+
+        return context
+
+    # Overridden method
+    def get(self, request, *args, **kwargs):
+        if request.user.is_authenticated():
+            return simple.redirect_to(request, url='../', permanent=False)
```
mitar: Why again hardcoded the same URL?

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
+        form_class = self.get_form_class()
+        form = self.get_form(form_class)
+        return self.render_to_response(self.get_context_data(form=form))
+
+    # Overridden method
+    def post(self, request, *args, **kwargs):
```
mitar, line 31: Shouldn't you just override `form_valid` and put:

```
username, password = form.save()
new_user = auth.authenticate(username=username, password=password)
auth.login(request, new_user)
return super(RegistrationView, self).form_valid(form)
```

into it and this is it?

piplmesh/account/views.py
```diff
@@ -4,57 +4,130 @@
...
-            return redirect('home')
-    data = {'form': form}
-    return render_to_response('registration.html', data, context_instance=context.RequestContext(request))
-
-def facebook_login(request):
+            return self.form_valid(form)
+        else:
+            return self.form_invalid(form)
+        
+class FacebookLoginView(generic_views.RedirectView):
```
mitar, line 42: Same comments as bellow.

piplmesh/account/views.py
Irrialite: If I put 'home' there it would always append it at the end as such: http://site/register/home

mitar: Then don't put home there. ;-)

https://docs.djangoproject.com/en/dev/topics/http/urls/#django-core-urlresolvers-utility-functions

piplmesh/account/views.py
Irrialite, line 90: Because we need request context for auth (and args in a different view). get_redirect_url only messes with the url string and *_kwargs. I'm not sure how python and C work together so I don't know what *_kwargs actually does at the moment.

piplmesh/account/views.py
Irrialite: These lines were copy paste from the original class but I'll change it I guess.

piplmesh/account/views.py
Irrialite, line 31: Problem is that form_valid doesn't get "request" passed to it. So we couldn't use auth.login(request, new_user). Have to do it in get() unless we pass request variable to the function as well, which would make it uglier.

mitar, line 31: OK, this is really badly documented, but it is here:

https://docs.djangoproject.com/en/dev/ref/class-based-views/#django.views.generic.base.View.dispatch

It means that in any class method you can use `self.request`, `self.args` and `self.kwargs` to access those variables as they would be passed to view if the view would be a function.

piplmesh/account/views.py
mitar: Weren't you on the class? We said the whole reason for using object-oriented programming is to reduce copy-pasting. If you have to copy-paste something, there is something very very wrong. This makes things unmaintainable (and unupgradeable).

piplmesh/account/views.py
Irrialite, line 31: Oh yea, I saw this as I was going through the source code but I just ignored it for some reason...

piplmesh/account/views.py
mitar, line 90: All this is accessible through `self`.

What has C with anything here? `kwargs` has nothing with C. Please read more about attribute passing in Python.

piplmesh/account/views.py
mitar, line 31: Yes. I also found this only when through the source code. It is really badly written in the documentation (this makes me think - we should open a ticket on Django for this, I will do it). But I don't ignore things but investigate them. ;-)

mitar:
Pull request as it is not cannot be merged cleanly. Please update with new commits from the main repository.

piplmesh/frontend/views.py
```diff
@@ -6,9 +6,4 @@ class HomeView(generic_views.TemplateView):
     def get_context_data(self, **kwargs):
```
mitar, line 6: Is then the whole method needed?

Irrialite, line 6: No, I forgot to delete it.

Irrialite:
Going to submit a new pull request with merged master.

mitar:
No, it is a normal thing that during a long open pull request, master goes on. It is nothing wrong to just update the pull request as this happens.

Irrialite:
Yes but I had some issues merging it since I'd not done it before so I just made a new pull request that's basically the same.

Issue #75, opened 2012-03-21
Use Facebook button


mitar:
Use Facebook button (that blue thing) and not text, so that it is more recognizable.

PR #78, opened 2012-03-21
Issue #75 Use Facebook button (team 3)

ghost:
Changed the link to Facebook login in to a button. The string on the button is easily translatable. Button's width is not fixed and does adjust to length of string. (#75)


mitar:
Great. But we said that we will move all this login stuff to separate page. This yet has to be done.

ghost:
I just want to know if i understand correctly, how separation of this login stuff should look. Do I need make a new file in piplmesh/frontend/templates called userbar.html for example, move the code for this "login menu" in there and then just {% include "userbar.html" %} in header.html? CSS stays in the same file as now. Am I correct?

mitar:
Just create a new view, new URL for it, and new template. Where you put all different ways for login there. So it would be (in spirit) similar to [this login page](http://stackoverflow.com/users/login).

Don't put anything in header - this would then display login in all pages. I am saying that it should be only on one page. Look how it is done in [nodewatcher](https://nodes.wlan-si.net/).

ghost:
Ok, now I understand. I thought we would have the Facebook login link always displayed right next to the regular login link, so I put it in the header.html.
Because we already have a login page I will just add this Facebook button there.

mitar:
No, we said in the class that we do not want all possible buttons to be displayed all the time in the header.

ghost:
Ok, understood.

PR #79, opened 2012-03-22
Issue 7 updated, merged; team 2

Irrialite:
Should be compatible with main repository now.


piplmesh/account/views.py
```diff
@@ -4,57 +4,90 @@
 from django.conf import settings
 from django.contrib import auth
 from django.core.urlresolvers import reverse
```
mitar: Imports ... ;-)

Irrialite: Assuming urlresolvers should be the import, not reverse?

piplmesh/account/views.py
```diff
@@ -4,57 +4,90 @@
...
-            auth.login(request, new_user)
-            return redirect('home')
-    data = {'form': form}
-    return render_to_response('registration.html', data, context_instance=context.RequestContext(request))
-
-def facebook_login(request):
+    template_name = 'registration.html'
+    form_class = forms.RegistrationForm
+
+    # Have to do this because we don't have reverse_lazy() yet
```
mitar, line 21: Please elaborate why would `reverse_lazy` be better in this case? So this comment is not the best. It should be more verbose, like "we have to do ... because ... which would ..."

Aaa, you mean because it cannot be as an attribute of the class. Yes.

You know, there is nothing wrong to push the requirements of Django to 1.4 release candidate version. Maybe do that? The rationale is that while you are developing a future product, it is OK to use a development version of the library as it will very probably become a stable version once you also finish your product. But you can already use the latest features (and thus be better).

So, bump the version in `requirements.txt`, go for lazy version.

piplmesh/account/views.py
```diff
@@ -4,57 +4,90 @@
...
+    def get_success_url(self):
+        return reverse('home')
+
+    def form_valid(self, form):
```
mitar, line 26: Is not this now simply beautiful? ;-)

piplmesh/account/views.py
mitar: Yes. ;-)

piplmesh/account/views.py
Irrialite, line 26: Perhaps :p

piplmesh/account/views.py
```diff
@@ -4,57 +4,90 @@
...
+        username, password = form.save()
+        new_user = auth.authenticate(username=username, password=password)
+        auth.login(self.request, new_user)
+        return super(RegistrationView, self).form_valid(form)
+
+    def get(self, request, *args, **kwargs):
```
mitar: It would be better to do this on the `dispatch` method. Because this view also does GET and POST it is a bit strange to do it in `get` only.

mitar: It is more natural to do on `dispatch`, because condition should work for both request types.

piplmesh/account/views.py
```diff
@@ -4,57 +4,90 @@
...
-    }
-    url = 'https://www.facebook.com/dialog/oauth?'
-    return http.HttpResponseRedirect('%s%s' % (url, urllib.urlencode(args)))
-
-def facebook_logout(request):
+
+    permanent = False
+    url = 'https://www.facebook.com/dialog/oauth'
+
+    def get(self, request, *args, **kwargs):
```
mitar: We said ... not do this in `get` method, but `get_redirect_url`.

piplmesh/context_processors.py
```diff
@@ -0,0 +1,8 @@
+from django.conf import settings
+
+def global_vars(request):
+    """
+    Adds global context variables to the context.
+    """
+
+    return {'SEARCH_ENGINE_UNIQUE_ID': settings.SEARCH_ENGINE_UNIQUE_ID}
```
mitar: Split into multiple lines, because there very probably be more values (so that we prepare for that, otherwise, for only one value, it would be OK like this.)

piplmesh/account/views.py
```diff
@@ -4,57 +4,90 @@
...
\ No newline at end of file
+
+    permanent = False
+    url = settings.FACEBOOK_LOGIN_REDIRECT
+  
+    def get(self, request, *args, **kwargs):
+        try:
+            token = request.GET['code']
+        except KeyError:
+            # If the user clicks cancel, they will be redirected back without logging in
```
mitar: _they_ -> _she_

mitar: Wouldn't something like this be simpler:

```
if 'code' in request.GET:
    user = auth.authenticate(token=request.GET['code'], request=request)
    auth.login(request, user)
    # message user that she has been logged in (maybe this will be already in auth.login once we move to MongoDB
else:
    # message user that she has not been logged in because she canceled the Facebook
return super(FacebookCallbackView, self).get(request, *args, **kwargs)
```

piplmesh/account/views.py
```diff
@@ -4,57 +4,90 @@
...
-    code = request.GET['code']
-    user = auth.authenticate(token=code, request=request)
-    auth.login(request, user)
-    return http.HttpResponseRedirect(settings.FACEBOOK_LOGIN_REDIRECT)
\ No newline at end of file
+
+    permanent = False
+    url = settings.FACEBOOK_LOGIN_REDIRECT
+  
+    def get(self, request, *args, **kwargs):
```
mitar, line 79: Do you remember when we said that GET requests should never change anything? So you see that here we do not do this. So it is wrong. So this should be `post` and we should not use simple links (GET request) for Facebook login, but POST form button (which can be rendered as a button, link, whatever) - the important things are twice: we get POST request and it is checked for the CSRF token.

piplmesh/account/views.py
Irrialite: If we don't know the gender, we have to use "they". Proper English :p

piplmesh/account/views.py
```diff
@@ -4,57 +4,90 @@
...
     """ 
     Log user out of Facebook and redirect to FACEBOOK_LOGOUT_REDIRECT. 
     """
+
+    permanent = False
+    url = settings.FACEBOOK_LOGOUT_REDIRECT
     
-    auth.logout(request)
-    return http.HttpResponseRedirect(settings.FACEBOOK_LOGOUT_REDIRECT)
+    def get(self, request, *args, **kwargs):
```
mitar, line 64: Same as bellow. GET requests shouldn't change anything.

piplmesh/account/views.py
Irrialite, line 21: Do I put 
pip install svn+http://code.djangoproject.com/svn/django/trunk/#egg=django
in requirements.txt file? I'm not sure how to do this.

piplmesh/account/views.py
mitar: Mnja mnja mnja. Please provide links for this argument. And please also search for links for my "she" argument. ;-) I think both goes, no?

piplmesh/account/views.py
mitar, line 21: You can always install with pip into virtualenv and then use pip freeze to get the requirements.txt. ;-)

I would not go for the latest trunk, but for release candidate. But also latest trunk would be good in other cases, where release candidate would not be already available.

So the idea is that we freeze the exact trunk version in requirements.txt, not just have the latest on time of install, because then we all will not have the same environment while they will develop Django further.

Irrialite, line 21: svn+http://code.djangoproject.com/svn/django/trunk/#egg=django-trunk
in file, that's just pip install thingy.

mitar, line 21: Please freeze the version.

Irrialite, line 21: So I download the tar gz file for RC then unzip and pip freeze?

mitar, line 21: No. Install through pip and then see what pip freeze says?

mitar, line 21: Or Google how you can freeze that.

piplmesh/account/views.py
Irrialite: http://en.wikipedia.org/wiki/Singular_they

mitar: Thank you. Please grep over whole source code and change any occurrence of _she_ to _they_.

piplmesh/account/views.py
```diff
@@ -3,58 +3,85 @@
...
-    auth.logout(request)
-    return http.HttpResponseRedirect(settings.FACEBOOK_LOGOUT_REDIRECT)
 
-def facebook_callback(request):
+    permanent = False
+    url = settings.FACEBOOK_LOGOUT_REDIRECT
+
+    # TODO
+    # use post instead of get
+    # use csrf tokens
```
mitar, line 63: This is checked automatically. You have to use CSRF in the form, though.

Write `TODO` comments like:

```
# TODO: Use POST instead of GET
# TODO: Use CSRF token in the form
```

In this way IDEs can parse them and display them.

piplmesh/context_processors.py
```diff
@@ -0,0 +1,10 @@
+from django.conf import settings
+
+def global_vars(request):
+    """
+    Adds global context variables to the context.
+    """
+
+    return {
+        'SEARCH_ENGINE_UNIQUE_ID': settings.SEARCH_ENGINE_UNIQUE_ID,
+        }
```
mitar, line 10: Unindent the last `}` one level to the left.

mitar:
Still waiting for the rest. ;-)

mitar:
Ping.

Irrialite:
Don't know how to convert gets to posts since they communicate between different sites.

mitar:
Ha? No? You make a POST request to our site, which then redirect you to Facebook.

Irrialite:
Also for some reason when I push to our branch it doesn't update here anymore.

Irrialite:
Closing because the repository where this branch resides changed. Made a new pull request with the same content.

Issue #82, opened 2012-03-22
XHTML validity


mitar:
We have some things which are invalid in XHTML. That should be fixed.

mitar:
Done with #83.

mitar:
There are still some warnings.

PR #83, opened 2012-03-23
XHTML validity (team 3)

ghost:
Fixed invalid things in XHTML. (#82)

### DIFF
--- a/piplmesh/frontend/templates/header.html
+++ b/piplmesh/frontend/templates/header.html
@@ -7,15 +7,16 @@
             </a>
         </div>
         <div id="search">
-            <div id="cse_search_form"></div>
-            <script src="https://www.google.com/jsapi" type="text/javascript"></script>
-            <script type="text/javascript">
-                var language_code = '{{ LANGUAGE_CODE }}';
-                var search_engine_unique_id = '{{ SEARCH_ENGINE_UNIQUE_ID }}';
-                var search_page_url = '{% url search %}';
-            </script>
+            <div id="cse_search_form">
+                <script src="https://www.google.com/jsapi" type="text/javascript"></script>
+                <script type="text/javascript">
+                    var language_code = '{{ LANGUAGE_CODE }}';
+                    var search_engine_unique_id = '{{ SEARCH_ENGINE_UNIQUE_ID }}';
+                    var search_page_url = '{% url search %}';
+                </script>
+            </div>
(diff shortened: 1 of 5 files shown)


piplmesh/frontend/templates/header.html
```diff
@@ -7,15 +7,16 @@
...
         </div>
         <div id="search">
-            <div id="cse_search_form"></div>
-            <script src="https://www.google.com/jsapi" type="text/javascript"></script>
-            <script type="text/javascript">
-                var language_code = '{{ LANGUAGE_CODE }}';
-                var search_engine_unique_id = '{{ SEARCH_ENGINE_UNIQUE_ID }}';
-                var search_page_url = '{% url search %}';
-            </script>
+            <div id="cse_search_form">
```
ghost, line 10: Please put this line back as it was, it was like that for a reason.

mitar:
In registration form I am getting some warnings.

piplmesh/frontend/templates/header.html
ghost, line 10: Ok, I'll change this back. Just out of curiosity - what's the reason?

mitar, line 10: Probably placeholder for the content which is then filled from JavaScript.

ghost, line 10: So there is no difference, because I just moved the <//div> tag  at the end of JS code.

ghost:
I don't really know how to fix these warnings. The problem is that the label for Gender "needs" 2 different "for" atributes, because of id_gender_0 and id_gender_1 "id" atributes of radio buttons.

piplmesh/frontend/templates/header.html
mitar, line 10: Exactly this is the difference. Once it is an empty placeholder where JavaScript code replaces the content with new one. In your case there is something inside which get replaced with new content and we lose JavaScript. I don't know what browser does in such case, but I do not want to disk if it is not necessary.

ghost, line 10: Ok, changing  back.

mitar:
Read here:

http://www.w3.org/TR/WCAG20-TECHS/H44.html#H44-ex3
http://www.w3.org/TR/WCAG20-TECHS/H71.html

So you will have to change template that for multiple fields it uses fieldset instead of label. Or: maybe you could always use fieldset for each row and labels only for those fields consisted of multiple inputs.

mitar:
Can you also fix the warnings please?

Issue #81, opened 2012-03-22
Structure CSS file


mitar:
We had a very nice CSS file structure, indented in accordance with DOM structure. Currently CSS file is not anymore likes this. It should be made again.

premik91:
Is this what you meant?

mitar:
I am not sure. ;-) But I think it has to follow also DOM structure, (`li` is inside the `ul`) not selector specificity. (The later is sometimes, but not always, related.)

premik91:
This is now done as in the first request.

mitar:
Pull request please.

mitar:
I give up on this ticket. Just everybody try to make a readable CSS while you change it.

PR #84, opened 2012-03-23
Issue #81, team 5

premik91:
pre-structured CSS file

### DIFF
--- a/piplmesh/frontend/static/piplmesh/css/style.css
+++ b/piplmesh/frontend/static/piplmesh/css/style.css
@@ -73,18 +73,18 @@ ul.errorlist {
 }
 
 .login_form {
-    text-align:right;
-    display:inline-block;
+    text-align: right;
+    display: inline-block;
 }
 
 #other_login_options {
-    margin-left:50px;
-    display:inline-block;
-    text-align:left;
+    margin-left: 50px;
+    display: inline-block;
+    text-align: left;
 }
 
 #other_login_options li {
-    padding-top:15px;
+    padding-top: 15px;
     list-style-type: none;
 }
 


piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -78,41 +83,41 @@ ul.errorlist {
...
-    width: 300px;
-    text-align: left;
-}
+        .registration .form_field .form_textbox {
+            width: 300px;
+            text-align: left;
+        }
 
-/* buttons */
+/* buttons -> all buttons should have same design */
```
mitar, line 123: I do not understand this comment?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -12,6 +12,10 @@ body {
     visibility: hidden;
 }
 
+.shown {
```
mitar, line 15: This was not defined before?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -78,41 +83,41 @@ ul.errorlist {
...
+    .button ul {
+        list-style-type: none;
+        overflow: hidden;
+    }
 
-.button li {
-    float: left;
-    padding-right: 4px;
-}
+    .button li {
```
mitar, line 139: Shouldn't this be indented one more?

piplmesh/frontend/static/piplmesh/css/style.css
mitar, line 123: Why then defining design for class and not directly for HTML elements? If we want all to have the same design?

mitar:
This pull request cannot be automatically merged. Please update to the HEAD from main repository.

piplmesh/frontend/static/piplmesh/css/style.css
premik91, line 123: Of course you are right. I don't even know why I added that.

piplmesh/frontend/static/piplmesh/css/style.css
premik91, line 15: No I just thought if someone wanted to make a visible element from hidden, he could change the class. But it is much easier to just change the atribute. I will delete.

mitar, line 15: Often you use classes for initial state. Afterwards you simply show and hide with jQuery.

mitar, line 15: Also don't add new features in the middle of some other commit. If you are just fixing the CSS file structure, then just fix that. And if you want to add another class, do another commit for that.

premik91:
Ok I suppose this is it :)

piplmesh/frontend/static/piplmesh/css/style.css
mitar, line 15: TODO?

premik91, line 15: https://github.com/frendo11/PiplMesh/commit/ec73cae43a0998f4e3a9e58a9d21956808b491c6
I did fix everything?

mitar, line 15: Have you updated the pull request? Here you see that it is still visible in the diff.

PR #85, opened 2012-03-25
Issue #81, team 5

premik91:
Fixed CSS indentation, restructured


mitar:
I am not sure if this is OK now. It is differently structured than #84 and it does not really follow the DOM but selectivity of the CSS?

premik91:
My mistake, I was told (by a friend) this is it. I am gonna "read" DOM how to do it correctly.

premik91:
Nothing much changed, just some code reduced. Am I going in the right direction?

mitar:
No. First do only CSS structure change. Then change also the styling. Don't mix features!

premik91:
I didn't change the styling, except aligning #main to center(and logout, but I can delete that). What do you mean by structure change than? All I did was restructuring classes, of whom some were unused or repeated. 
I thought DOM structure expects exactly that, hmm, I am a bit lost sorry.

mitar:
I see that you changed also HTML. So this is more than the tickets asks. Ticket asks only for CSS structure. No other changes. But you are welcome to cleanup/change/improve the design/HTML/CSS. But this should be another pull request.

premik91:
Now only restructuring to CSS remains. When I was looking at CSS I noticed that some classes referenced in the HTML don't exist or are never used. That't why I updated the HTML, and since changes to CSS reflects on HTML, I thought I should update them as well. I will do it in another pull request then.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -17,13 +17,18 @@ a {
     text-decoration: none;
 }
 
-.gravatar {
-    border: 0px;
+/* Styling lists */
+ul {
```
mitar, line 21: What? All lists you want to make non-list? Why?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -117,52 +114,36 @@ ul.errorlist {
 
 /* ------------------ REGISTRATION ---------------------------- */
 .registration {
-    width: 960px;
-    margin: auto;
     margin-top: 50px;
 }
 
-.registration .form_field ul {
```
mitar, line 125: What is now this? You are changing the CSS content, not just structure!

mitar:
This is not good. Please redo everything again. Just structure the CSS. If the class is not used anymore, you can remove it. But otherwise do not change/optimize things. For that (and other design changes) make another pull request.

piplmesh/frontend/static/piplmesh/css/style.css
premik91, line 21: Because all our list up to date, have "list-style: none;", and DOM structure (as I understand) also wants that all classes are used as much as possible(and repetitions reduced if possible). And all our list styles have/had list-style set to none. Such repetition is unnecessary. And if one in three of lists wants different list-style, that one will overwrite this atribute. I thought this is more logical?

piplmesh/frontend/static/piplmesh/css/style.css
premik91, line 125: It lost importance, when I added the body ul {} class.

premik91:
Ok, but as I understand now you want something like the first pull request? Just the DOM like indentation?

piplmesh/frontend/static/piplmesh/css/style.css
mitar, line 21: No, this is bad reason. There will be definitely also regular lists on our page. You should change default style of regular lists only if you want to design them in some way. Not for those special cases when we use list structure, but design them as no-list.

premik91, line 21: These were not special cases, all our lists have this atributes(for now). But ok, I understand what you mean.

PR #87, opened 2012-03-26
Changed the cse_search_form div back, small CSS fix on login page.

ghost:


piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -83,6 +83,10 @@ ul.errorlist {
     text-align: left;
 }
 
+#other_login_options ul {
+    margin: 0;
+    padding: 0;
+}
```
mitar, line 89: Add empty line after the block.

ghost, line 89: Fixed.

PR #88, opened 2012-03-27
Issue #81, team 5

premik91:
Indentation like DOM structure


piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -72,10 +72,10 @@ ul.errorlist {
...
-.login_form {
-    text-align: right;
-    display: inline-block;
-}
+    .login_form {
+        text-align: right;
+        display: inline-block;
+    }
 
 #other_login_options {
```
mitar: This is bellow #main, no?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -83,15 +83,15 @@ ul.errorlist {
...
-    padding-top: 15px;
-    list-style-type: none;
-}
+        #other_login_options li {
+            padding-top: 15px;
+            list-style-type: none;
+        }
 
 /* ---------------- FB LOGIN BUTTON ------------------- */
 .fb_login_button {
```
mitar: Also.

mitar: Of course those "sections" are not really compatible with the DOM structure approach.

mitar: Probably there should be one section for some general things, for home page, and then each for subpages.

piplmesh/frontend/static/piplmesh/css/style.css
premik91: I also indented .register classes, because I am planning to move them under the #main (I already did, but will put in another request).

piplmesh/frontend/static/piplmesh/css/style.css
premik91: I did something in that order. I also moved .button classes up to body, since they will be used on all buttons(so they don't really belong under registration)? Or should for the momment leave them where they were?

piplmesh/frontend/static/piplmesh/css/style.css
mitar: Ahhh. You really do not get the concept of one feature at the time? Now just restructure CSS according to current content. This was meant to be done very quickly at the beginning of the milestone so that others could build upon this. Now we are playing with this for days. When it is so simple. And now others will have problems.

mitar:
I give up.

piplmesh/frontend/static/piplmesh/css/style.css
premik91: I am giving up too, what was wrong now. I didn't change any other feature, just did what you said. And I don't see any harm done in making more features (if you don't like them thats a different story). How can it be simple, if whatever I did you said fix this and this, and when I fixed it, everything was wrong. I am afraid to start anything now. :)

mitar: Yes, sorry. There are some things which are hard to delegate because you spend more time delegating than doing yourself. This showed like one of those. Sorry for all the work you put in. Just do something where you have more free hands and can express your ideas freely. But still concentrate only on one feature. If you find some other thing to fix, do it in another branch, another pull request.

premik91: Ok.

PR #77, opened 2012-03-21
Issue#66 language selection enabled, team 5

ghost:
User can now select language based on his/her browser language settings.

PR #89, opened 2012-03-27
Updated views, team 2

Irrialite:
New pull request for class based views.


Irrialite:
Django 1.4 works it seems. Didn't work before.

piplmesh/account/templates/login.html
```diff
@@ -6,10 +6,10 @@
 
 {% block content %}
 <div class="main">
-    <form method="post" action="{% url login %}">
+    <form method="post" action="{% url login %}" id="login_form">
     {% csrf_token %}
     {{ form.as_p }}
-    <input type="submit" value="{% trans "Login" %}" />
+    <p><button type="submit" class="login_button" value="value"><span class="login_button_text">{% trans "Login" %}</span></button></p>
```
mitar: Why you need span around text? Why not simply style button element? Why additional element?

piplmesh/account/views.py
```diff
@@ -3,58 +3,75 @@
...
+
+    permanent = False
+
+    def get_redirect_url(self, **kwargs):
+        args = {
+            'client_id': settings.FACEBOOK_APP_ID,
+            'scope': settings.FACEBOOK_SCOPE,
+            'redirect_uri': self.request.build_absolute_uri(urlresolvers.reverse('facebook_callback')),
+        }
+        url = 'https://www.facebook.com/dialog/oauth' % kwargs
```
mitar: I am not sure what are you doing here with `kwargs`?

piplmesh/account/views.py
```diff
@@ -3,58 +3,75 @@
...
+        return "%s?%s" % (url, urllib.urlencode(args))
```
mitar: Can you do keyword-based string interpolation?

piplmesh/account/views.py
```diff
@@ -3,58 +3,75 @@
...
+        
+class FacebookLogoutView(generic_views.RedirectView):
     """ 
     Log user out of Facebook and redirect to FACEBOOK_LOGOUT_REDIRECT.
```
mitar: _This view logs the user ..._, be consistent ...

piplmesh/account/views.py
```diff
@@ -3,58 +3,75 @@
...
     """
-    
-    auth.logout(request)
-    return http.HttpResponseRedirect(settings.FACEBOOK_LOGOUT_REDIRECT)
 
-def facebook_callback(request):
+    permanent = False
+    url = settings.FACEBOOK_LOGOUT_REDIRECT
+
+    def get(self, request, *args, **kwargs):
```
mitar: It should be POST.

piplmesh/account/views.py
```diff
@@ -3,58 +3,75 @@
...
-    code = request.GET['code']
-    user = auth.authenticate(token=code, request=request)
-    auth.login(request, user)
-    return http.HttpResponseRedirect(settings.FACEBOOK_LOGIN_REDIRECT)
\ No newline at end of file
+
+    permanent = False
+    url = settings.FACEBOOK_LOGIN_REDIRECT
+
+    def get(self, request, *args, **kwargs):
```
mitar, line 56: It should be POST.

piplmesh/account/views.py
mitar: See how your friend is doing this in #77.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -65,6 +65,21 @@ ul.errorlist {
     padding-left: 110px;
     padding-bottom: 10px;
 }
+
+/* -------------------- LOGIN ------------------------ */
+.login_button {
+	background: none;
```
mitar: Indentation!

piplmesh/frontend/templates/header.html
```diff
@@ -31,7 +31,10 @@
             {% else %}
                 <p><a href="{% url registration %}">{% trans "Register" %}</a></p>
                 <p><a href="{% url login %}">{% trans "Log in" %}</a></p>
-                <p><a class="fb_login_button" href="{% url facebook_login %}"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></a></p>
+                <form method="post" action="{% url facebook_login %}" id="fb_login_form">
+                	{% csrf_token %}
```
mitar: Indentation!

piplmesh/frontend/templates/header.html
```diff
@@ -31,7 +31,10 @@
             {% else %}
                 <p><a href="{% url registration %}">{% trans "Register" %}</a></p>
                 <p><a href="{% url login %}">{% trans "Log in" %}</a></p>
-                <p><a class="fb_login_button" href="{% url facebook_login %}"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></a></p>
+                <form method="post" action="{% url facebook_login %}" id="fb_login_form">
```
mitar: So you do know how to do this. So do it also for logout.

piplmesh/account/views.py
mitar, line 56: Are you sure Facebook is doing GET callback? Or it can also do the POST callback?

piplmesh/account/views.py
```diff
@@ -3,58 +3,75 @@
...
+        if 'code' in request.GET:
+            user = auth.authenticate(token=request.GET['code'], request=request)
+            auth.login(request, user)
+            # message user that they have been logged in (maybe this will be already in auth.login once we move to MongoDB
+            return super(FacebookCallbackView, self).get(request, *args, **kwargs)
+        else:
+            # message user that they have not been logged in because they cancelled the Facebook
```
mitar: Add `TODO` at the beginning of the comment:

```
# TODO: Message user that they have not been logged in because they cancelled the Facebook
```

Also, use information provided from the Facebook why login was not successful.

piplmesh/account/views.py
```diff
@@ -3,58 +3,75 @@
...
\ No newline at end of file
+
+    permanent = False
+    url = settings.FACEBOOK_LOGIN_REDIRECT
+
+    def get(self, request, *args, **kwargs):
+        if 'code' in request.GET:
+            user = auth.authenticate(token=request.GET['code'], request=request)
+            auth.login(request, user)
+            # message user that they have been logged in (maybe this will be already in auth.login once we move to MongoDB
```
mitar: Add TODO.

piplmesh/account/views.py
mitar, line 56: OK, Facebook makes a redirect, so it is a GET.

piplmesh/account/views.py
Irrialite: This was copy paste from the previous version not done by me; will change.

piplmesh/account/views.py
Irrialite, line 56: That's why I was confused since there's no way to get post from an outside source if they don't post.

piplmesh/frontend/templates/header.html
Irrialite: Will do

piplmesh/account/views.py
Irrialite: Nothing, I kept it because the original method had it and I thought it would be a nice reminder that it is there in the original method even if it does nothing.

piplmesh/account/views.py
Irrialite: This class is actually never used. When a user logs out, they go to "logout" url not "facebook_logout".

piplmesh/account/views.py
mitar: I really like your argument and completely agree that we should just restructure the views and not change the code itself. But not too much. In this case it is something plain wrong. ;-) Or at least not explained. So I would fix it.

piplmesh/account/views.py
mitar: Very good observation. Later on we will combine this into one URL. But for now change the template so that Facebook logout URL is displayed if the user is logged in through Facebook, no?

piplmesh/account/views.py
mitar, line 56: OK. But elsewhere use POST.

You should probably also add some security. So that if somebody tricks user in opening this URL without him pressing login button before, nothing happens. And because login button will be secured with POST ...

piplmesh/account/views.py
Irrialite: I don't know if that makes much sense since they're essentially logged out only from our site - not facebook. A message like "You have been logged out of facebook" would make no sense, since they wouldn't be logged out of facebook (they can still go on facebook and do w/e) but only from our site that is linked to facebook. I don't think the distinction is necessary.

mitar: I am not saying that the user should see any difference? I am just saying that the logout URL to which user goes once it click the "Logout" link should be different based on the way the user has logged into our site, no? Why we then have the facebook logout view in the first place? It seem to me there is no need for that anyway? It does not do anything more than a normal logout view, no? So maybe we could simply remove this view and always use just the normal/standard logout view?

piplmesh/account/views.py
mitar: But you have changed the comment for the other view. I am saying just that if you change one, change also the other. Be consistent.

piplmesh/account/templates/login.html
```diff
@@ -6,10 +6,10 @@
 
 {% block content %}
 <div class="main">
-    <form method="post" action="{% url login %}">
+    <form method="post" action="{% url login %}" id="login_form">
     {% csrf_token %}
     {{ form.as_p }}
-    <input type="submit" value="{% trans "Login" %}" />
+    <p><button type="submit" class="auth_button" value="value"><span class="auth_button_text">{% trans "Login" %}</span></button></p>
```
mitar: Span is unnecessary.

piplmesh/account/views.py
Irrialite: Yep, keeping it is a waste of space at the moment. If we wanted to do something else depending on whether user was logged in through facebook then we could always add it back in.

piplmesh/account/views.py
```diff
@@ -3,58 +3,76 @@
...
+
+    permanent = False
+
+    def get_redirect_url(self, **kwargs):
+        args = {
+            'client_id': settings.FACEBOOK_APP_ID,
+            'scope': settings.FACEBOOK_SCOPE,
+            'redirect_uri': self.request.build_absolute_uri(urlresolvers.reverse('facebook_callback')),
+        }
+        url = 'https://www.facebook.com/dialog/oauth'
```
mitar: Why just do not move this to the string bellow?

`"https://www.facebook.com/dialog/oauth?%(args)s"`

piplmesh/account/views.py
```diff
@@ -3,58 +3,76 @@
...
+        return "%(url)s?%(args)s" % {'url':url, 'args':urllib.urlencode(args)}
```
mitar: Space after `:`.

piplmesh/account/views.py
```diff
@@ -3,58 +3,76 @@
...
+        
+class FacebookLogoutView(generic_views.RedirectView):
```
mitar: So this view is useless. We can just remove it. Also remove the setting.

piplmesh/account/views.py
mitar, line 56: So leave it for now like this. Just add a TODO comment that we should check if the callback is really called as a consequence of the user pressed Facebook login button before.

piplmesh/account/views.py
```diff
@@ -3,58 +3,76 @@
...
\ No newline at end of file
+
+    permanent = False
+    url = settings.FACEBOOK_LOGIN_REDIRECT
+
+    def get(self, request, *args, **kwargs):
+        if 'code' in request.GET:
+            user = auth.authenticate(token=request.GET['code'], request=request)
+            auth.login(request, user)
+            # TODO: message user that they have been logged in (maybe this will be already in auth.login once we move to MongoDB
```
mitar: Upper case the first letter of the TODO comment.

piplmesh/account/views.py
```diff
@@ -3,58 +3,76 @@
...
+            return super(FacebookCallbackView, self).get(request, *args, **kwargs)
+        else:
+            # TODO: message user that they have not been logged in because they cancelled the Facebook
```
mitar: Upper case the first letter of the TODO comment.

piplmesh/account/views.py
```diff
@@ -3,58 +3,76 @@
...
+            # TODO: use information provided from facebook as to why login was not successful
```
mitar: Upper case the first letter of the TODO comment.

mitar:
This is starting to look good. Please update to the latest from the official repository so that it can be automatically merged.

requirements.txt
```diff
@@ -1 +1,2 @@
-Django==1.3.1
+Django==1.4
+
```
mitar: Empty line?

piplmesh/frontend/templates/header.html
```diff
@@ -28,10 +28,17 @@
...
+                <form method="post" action="{% url logout %}" id="logout_form">
+                	  {% csrf_token %}
+                    <p><button type="submit" class="auth_button" value="value">{% trans "Log out" %}</button></p>
+                </form>
             {% else %}
                 <p><a href="{% url registration %}">{% trans "Register" %}</a></p>
                 <p><a href="{% url login %}">{% trans "Log in" %}</a></p>
+                <form method="post" action="{% url facebook_login %}" id="fb_login_form">
+                    {% csrf_token %}
+	                  <p><button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></button></p>
```
mitar: Why span here?

piplmesh/account/templates/login.html
```diff
@@ -11,12 +11,17 @@
...
+                <p><button type="submit" class="auth_button" value="value">{% trans "Login" %}</button></p>
             </form>
         </div>
         <div id="other_login_options">
             <ul>
-                <li><a class="fb_login_button" href="{% url facebook_login %}"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></a></li>
+                <li>
+                		<form method="post" action="{% url facebook_login %}" id="fb_login_form2">
+                    		{% csrf_token %}
+	                  		<button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></button>
```
mitar: Why span?

Is this XHTML valid?

piplmesh/frontend/templates/header.html
```diff
@@ -28,10 +28,17 @@
...
                 <p>{% trans "Username:" %}<strong>{{ user.username }}</strong> </p>
-                <p><a href="{% url logout %}">{% trans "Log out" %}</a></p>
+                <form method="post" action="{% url logout %}" id="logout_form">
+                	  {% csrf_token %}
+                    <p><button type="submit" class="auth_button" value="value">{% trans "Log out" %}</button></p>
+                </form>
             {% else %}
                 <p><a href="{% url registration %}">{% trans "Register" %}</a></p>
                 <p><a href="{% url login %}">{% trans "Log in" %}</a></p>
+                <form method="post" action="{% url facebook_login %}" id="fb_login_form">
```
mitar: Didn't we said that there will be no login with Facebook button on all pages but just on the login subpage?

mitar:
Please upgrade to the latest from main repository so that automatic merge can be done.

Irrialite:
It's upgraded to the latest.

piplmesh/account/templates/login.html
Irrialite: What else could be there, other than span?

mitar: Simply nothing? Why there should be span, when you already have `button` around? Just put all wanted style on `button`. No?

piplmesh/account/templates/login.html
```diff
@@ -11,12 +11,17 @@
...
+                <p><button type="submit" class="auth_button" value="value">{% trans "Login" %}</button></p>
             </form>
         </div>
         <div id="other_login_options">
             <ul>
-                <li><a class="fb_login_button" href="{% url facebook_login %}"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></a></li>
+                <li>
+                		<form method="post" action="{% url facebook_login %}" id="fb_login_form2">
+                    		{% csrf_token %}
+	                  		<p><button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></button></p>
```
mitar, line 22: Unnecessary span.

mitar:
It is still not updated to the latest (as I am being told by GitHub)?

Irrialite:
I merged with the main repository and removed conflicts, thought that was enough?

mitar:
But this was done last time 19 hours ago, no? There is newer stuff in the main repository. See the list of commits for this pull request.

Irrialite:
Oh, I might have forgotten to refresh.

piplmesh/account/templates/login.html
Irrialite: I think it needs 2 divs but I wasn't the one who made it so I'm not sure.

PR #93, opened 2012-03-30
Issue 61, Issue 62 rev-2 (team 4)

agiz:
Updated to latest master from main repository.
Removed test comments.


mitar:
Please update to the latest from the main repository.

agiz:
This was up to date 10 minutes ago. I mean I'd rather wait for you to finish all the merges else I'd have to keep updating ad infinitum.

mitar:
I am pulling requests as pull request are finalized. Currently all other are pending (some changes needed). But yes, you are right. First fix all my comments (I have to write them first) and then update the pull request just before I say that everything looks OK. There is no other reason to update with the latest from the main repository then to that it could be easier for you to update later on, if you update regularly.

piplmesh/account/backends.py
```diff
@@ -1,23 +1,31 @@
 import json, urlparse, urllib
 
 from django.conf import settings
-from django.contrib.auth import backends, models as auth_models
 from django.core import urlresolvers
 
+from mongoengine.django import auth as mongo_models
```
mitar: Don't rename module so much (just add a prefix to avoid conflicts, but leave it as it is). So `from foo.bar import xxx` is OK if `xxx` is unique, otherwise you prefix it with `bar`: `from foo.bar import xxx as bar_xxx`.

piplmesh/account/backends.py
```diff
@@ -1,23 +1,31 @@
...
+
 from piplmesh.account import models
 
-class CaseInsensitiveModelBackend(backends.ModelBackend):
-    """
-    This backend uses case-insensitive username authentication which is not supported by default.
-    """
-    
+class CustomUserModelBackend(mongo_models.MongoEngineBackend):
```
mitar: I like this. But I do not like the name. ;-) I would just leave it `MongoEngineBackend`. It tells everything. ;-) So:

```
class MongoEngineBackend(mongo_models.MongoEngineBackend):
```

This is also the reason why we are using prefixes when importing. ;-)

piplmesh/account/backends.py
```diff
@@ -1,23 +1,31 @@
...
+    supports_object_permissions = True
+    supports_anonymous_user = True
+    supports_inactive_user = False
```
mitar, line 16: Are we changing those values? Don't `mongo_models.MongoEngineBackend` already defines them in the way we like?

I would like that our user model supports all those.

agiz, line 16: The default values for those are False unless I am missing something. Source: https://github.com/hmarr/mongoengine/blob/master/mongoengine/django/auth.py

mitar, line 16: You are correct. But can you just turn on those things without support for it in the code? Have you coded that?

piplmesh/account/backends.py
```diff
@@ -1,23 +1,31 @@
...
+
     def authenticate(self, username=None, password=None):
-        try:
-            user = auth_models.User.objects.get(username__iexact=username)
-            if user.check_password(password):
+        user = self.user_class.objects(username__iexact=username).first()
```
mitar, line 19: Good.

piplmesh/account/backends.py
```diff
@@ -1,23 +1,31 @@
...
-        except auth_models.User.DoesNotExist:
-            return None
+        return None
+
+    def get_user(self, user_id):
+        return self.user_class.objects.with_id(user_id)
+
+    @property
+    def user_class(self):
+        self._user_class = models.CustomUser
```
mitar: I think this is unnecessary?

mitar: I would just return `models.CustomUser` (or `models.User` as I would call it) and leave to any possible future implementation to extend the method if wanted.

piplmesh/account/backends.py
```diff
@@ -47,25 +55,23 @@ def authenticate(self, token=None, request=None):
 
         try:
```
mitar: There was already before the problem with this approach: race condition. What if somebody creates an user object just between we check for existence and then go and try to create it? Can we do something about that now that we have moved to MongoDB? `get_or_create`?

piplmesh/account/backends.py
```diff
@@ -74,9 +80,14 @@ def get_user(self, user_id):
         """
     
         try:
-            return auth_models.User.objects.get(pk=user_id)
-        except auth_models.User.DoesNotExist:
+            return self.user_class.objects.with_id(user_id)
+        except self.user_class.DoesNotExist:
```
mitar: Why are you capturing the exception here and not also in the other backend?

piplmesh/account/backends.py
```diff
@@ -74,9 +80,14 @@ def get_user(self, user_id):
...
             return None
 
         supports_object_permissions = False
-        supports_anonymous_user = True
\ No newline at end of file
+        supports_anonymous_user = True
+
+    @property
+    def user_class(self):
+        self._user_class = models.CustomUser
```
mitar: Hm. What if we would just extend Facebook backend from the default one above? Then we would just have to change `authenticate` and we get everything else for free (like permissions)?

mitar: Or we could use this backend only for Facebook authentication and the above one for everything else (permissions and so on). Anyway, we should test and see.

mitar: BTW, this line is also unnecessary. Same as above. Memoization/caching is useful if you have some expensive computation to do and present it as a property.

piplmesh/account/backends.py
```diff
@@ -1,23 +1,31 @@
...
+
+    def get_user(self, user_id):
+        return self.user_class.objects.with_id(user_id)
+
+    @property
+    def user_class(self):
+        self._user_class = models.CustomUser
+        return self._user_class
 
 class FacebookBackend:
```
mitar: This is wrong. All classes in Python should extend `object`. Otherwise they are old-style classes.

piplmesh/account/backends.py
```diff
@@ -47,25 +55,23 @@ def authenticate(self, token=None, request=None):
 
         try:
             # Check if user profile exists
-            profile = models.UserProfile.objects.get(facebook_id=fb['id'])
-            user = profile.user
+            user = self.user_class.objects(facebook_id=fb['id']).first()
```
mitar: Use maybe `fb.get('id')` so that it does not throw an exception (and 500) if it does not exist?

piplmesh/account/backends.py
```diff
@@ -47,25 +55,23 @@ def authenticate(self, token=None, request=None):
...
 
             # Update access token
-            profile.token = access_token
-            profile.save()
+            user.token = access_token
```
mitar: _token_ -> _facebook_token_?

agiz:
Can I update this pull? I will just fix things you commented on for now (ie. will not yet update with main repository).

mitar:
Of course you can update this pull. This is the idea. ;-) And also you do not need to make a new pull request if you update from the main repository. ;-)

piplmesh/account/backends.py
```diff
@@ -47,25 +55,23 @@ def authenticate(self, token=None, request=None):
...
+            user.save()
 
-        except models.UserProfile.DoesNotExist:
+        except self.user_class.DoesNotExist:
             # User profile doesn't exist, create new user
             username = fb.get('username', fb['id'])
```
mitar: What if user with such username already exists? Add TODO comment for that. (Or fix it.)

piplmesh/account/backends.py
```diff
@@ -44,39 +55,9 @@ def authenticate(self, token=None, request=None):
...
-
-        try:
-            # Check if user profile exists
-            profile = models.UserProfile.objects.get(facebook_id=fb['id'])
-            user = profile.user
-
-            # Update access token
-            profile.token = access_token
-            profile.save()
-
```
mitar, line 56: Do we always get e-mail?

piplmesh/account/backends.py
agiz: Fixed

piplmesh/account/backends.py
```diff
@@ -44,39 +55,9 @@ def authenticate(self, token=None, request=None):
...
-        except models.UserProfile.DoesNotExist:
-            # User profile doesn't exist, create new user
-            username = fb.get('username', fb['id'])
-            user = auth_models.User.objects.create_user(username=username, email=fb['email'])
-            user.first_name = fb['first_name']
-            user.last_name = fb['last_name']
```
mitar, line 62: Are you sure we can just simply assign this value? It is compatible?

piplmesh/account/backends.py
agiz: Changed to MongoEngineBackend(mongo_auth.MongoEngineBackend):.

piplmesh/account/backends.py
agiz, line 16: Changed supports_inactive_user to True. I am not quite sure where else shall we change those things.

piplmesh/account/backends.py
agiz: Fixed.

piplmesh/account/forms.py
```diff
@@ -1,14 +1,14 @@
+from account import models
```
mitar: Don't import like this. What if there is some Python package/module in Python path with name `account`? Use absolute/relative imports. Or `from piplmesh.account import models`.

Also order this import bellow. First come system modules, then Django, then other libraries, then our imports.

piplmesh/account/backends.py
agiz: try/catch replaced with get_or_create().

piplmesh/account/backends.py
agiz: Fixed.

piplmesh/account/forms.py
```diff
@@ -1,14 +1,14 @@
...
 from django import forms
 from django.conf import settings
-from django.contrib.auth import forms as auth_forms, models as auth_models
+from django.contrib.auth import forms as auth_forms
 from django.forms.extras import widgets
 from django.utils import safestring
 from django.utils.translation import ugettext_lazy as _
 
-from piplmesh.account import models, fields
-
```
mitar, line 11: Don't rename module like that.

piplmesh/account/forms.py
```diff
@@ -32,7 +32,7 @@ class RegistrationForm(auth_forms.UserCreationForm):
 
     # Required data
     username = forms.CharField(label=_("Username"))
-    email = forms.EmailField(label=_("E-mail"))
+    email = forms.EmailField(label=_("Email"))
```
mitar: I think we want _E-mail_. No?

piplmesh/account/backends.py
agiz: FacebookBackend now extends MongoEngineBackend (previous CustomUserModelBackend).

piplmesh/account/forms.py
```diff
@@ -1,14 +1,14 @@
...
-from django.contrib.auth import forms as auth_forms, models as auth_models
+from django.contrib.auth import forms as auth_forms
 from django.forms.extras import widgets
 from django.utils import safestring
 from django.utils.translation import ugettext_lazy as _
 
-from piplmesh.account import models, fields
-
 # Form settings
 GENDER_CHOICES = (
```
mitar: This comment is obsolete?

piplmesh/account/backends.py
agiz: Fixed.

piplmesh/account/forms.py
```diff
@@ -53,25 +53,23 @@ def clean_username(self):
...
     def save(self):	
         # We first have to save user to database
-        new_user = auth_models.User(
+        new_user = models.User(
             username=self.cleaned_data['username'],
             first_name=self.cleaned_data['first_name'],
             last_name=self.cleaned_data['last_name'],
             email=self.cleaned_data['email'],
+            gender=self.cleaned_data['gender'],
+            birthdate=self.cleaned_data['birthdate'],
```
mitar, line 61: `language` is missing (or you have not yet pulled this in)?

piplmesh/account/backends.py
agiz: Changed fb['id'] to fb.get('id'). Also applied this style to all other fb variables.

piplmesh/account/backends.py
agiz: Renamed.

piplmesh/account/models.py
```diff
@@ -1,21 +1,111 @@
-from django.contrib.auth import models as auth_models
-from django.db import models as django_models
+from mongoengine import *
```
mitar: Order order order. First system, Django, other libraries, then ours ...

Never import everything. Do only `import mongoeinge`.

piplmesh/account/models.py
```diff
@@ -1,21 +1,111 @@
...
 
-from piplmesh.account import fields
+from django.utils.encoding import smart_str
```
mitar: Import only modules, not directly their contents.

piplmesh/account/models.py
```diff
@@ -1,21 +1,111 @@
...
+from django.contrib.auth.models import AnonymousUser
+from django.contrib.auth.hashers import check_password, make_password
+from django.utils.translation import ugettext_lazy as _
 
-class UserProfile(django_models.Model):
-    """ 
-    Class used for storing additional information about user.
+import datetime
```
mitar, line 1: System imports come first.

piplmesh/account/models.py
```diff
@@ -1,21 +1,111 @@
...
+
+REDIRECT_FIELD_NAME = 'next'
```
mitar: Why don't simply use the one from `django.contrib.auth`?

piplmesh/account/models.py
```diff
@@ -1,21 +1,111 @@
...
+
+class User(Document):
```
mitar: Why you don't extend the one provided by mongoengine?

piplmesh/account/models.py
```diff
@@ -1,21 +1,111 @@
...
+    """A User document that aims to mirror most of the API specified by Django
```
mitar: Use reST markup for comments (and links).

piplmesh/account/backends.py
mitar: Of course the idea is not always to hide the fact that there are missing values. But to behave in a smart way. If the value is missing and it is required, then there should be some error (just not unhandled exception which throws 500).

piplmesh/account/models.py
```diff
@@ -1,21 +1,111 @@
...
+    def check_password(self, raw_password):
+        """Checks the user's password against a provided password - always use
+        this rather than directly comparing to
+        :attr:`~mongoengine.django.auth.User.password` as the password is
+        hashed before storage.
+        """
+        return check_password(raw_password, self.password)
+
+    @classmethod
+    def create_user(cls, username, password, email=None):
```
mitar: I am not sure if this goes into model. I think it goes into [queryset manager](http://mongoengine.org/docs/v0.5/guide/querying.html#custom-querysets)?

piplmesh/account/models.py
agiz: I am not sure extending mongoengine User is this trivial. I've spent sever days on this matter and tried numerous combinations and at the end chose this one. 

The whole User from models.py is just a copy/paste from mongoengine auth.py (with some custom fields added) which seems to me plausible short to mid-term solution. In the long run I expect django to implement non-relational sources as well. 

I don't know what to do here... Do you want me to implement your notations in (slightly changed mongoengine) User because as I said this is just a ripoff of mongoengine auth.py.

piplmesh/account/models.py
mitar: I think you could steal this method directly from the `django.contrib.auth.models.UserManager`. :-) You just have to make sure that `self.model` points to our user document. :-) Soo, something like:

```
class UserQuerySet(mongoengine.queryset.QuerySet):
       super(UserQuerySet, self).__init__(*args, **kwargs)
       self.model = self._document

    create_user = django.contrib.auth.models.UserManager.create_user
```

:-)

mitar: Never copy-paste code. ;-)

piplmesh/account/models.py
```diff
@@ -1,21 +1,111 @@
...
-    gender = django_models.CharField(max_length=6, blank=True)
-    facebook_id = django_models.BigIntegerField(verbose_name=u'Facebook ID', null=True, blank=True)
-    token = django_models.CharField(max_length=150)
-    language = fields.LanguageField(verbose_name=u'language')
-    
+    birthdate = DateTimeField()
+    gender = StringField(max_length=6)
+    facebook_id = IntField()
+    facebook_token = StringField(max_length=150)
+
```
mitar: One empty line too many.

piplmesh/settings.py
```diff
@@ -3,9 +3,13 @@
 # Development Django settings for PiplMesh project.
 
 import os.path
+import mongoengine
+
 settings_dir = os.path.abspath(os.path.dirname(__file__))
 database_file = os.path.join(settings_dir, 'db.sqlite')
```
mitar: At the end we would like to get completely rid of relation database. So this can go out. No?

piplmesh/settings.py
```diff
@@ -3,9 +3,13 @@
...
 
+mongoengine.connect('project')
```
mitar: The name of the database should probably be a setting?

piplmesh/settings.py
```diff
@@ -3,9 +3,13 @@
 # Development Django settings for PiplMesh project.
 
 import os.path
+import mongoengine
```
mitar, line 9: Move this bellow just before the `mongoengine.connect`.

piplmesh/settings.py
```diff
@@ -189,10 +191,12 @@
 
 LOGIN_REDIRECT_URL = '/'
 
-AUTH_PROFILE_MODULE = 'account.UserProfile'
+SESSION_ENGINE = 'mongoengine.django.sessions'
+
+CUSTOM_USER_MODEL = 'accounts.User'
```
mitar: This is not used anywhere?

piplmesh/account/backends.py
agiz: Fixed automagically with get_or_create().

piplmesh/settings.py
```diff
@@ -147,8 +151,6 @@
     os.path.join(settings_dir, 'templates'),
 )
 
-MESSAGE_STORAGE = 'django.contrib.messages.storage.session.SessionStorage'
```
mitar, line 151: Why removed this? We want messages to be stored in the session? And session in MongoDB?

piplmesh/account/backends.py
agiz, line 56: I am not sure what this is pointed at. I've updated this pull and lines got crossed?

piplmesh/account/forms.py
agiz: Fixed.
Edit: but not yet updated.
Edit 2: now updated.

piplmesh/account/models.py
```diff
@@ -1,21 +1,111 @@
...
+                pass
+            else:
+                email = '@'.join([email_name, domain_part.lower()])
+
+        user = cls(username=username, email=email, date_joined=now)
+        user.set_password(password)
+        user.save()
+        return user
+
+    def get_and_delete_messages(self):
```
mitar: Why? We want support for messages? They do not work on MongoDB?

piplmesh/account/backends.py
```diff
@@ -1,25 +1,36 @@
...
+
 from piplmesh.account import models
 
-class CaseInsensitiveModelBackend(backends.ModelBackend):
-    """
-    This backend uses case-insensitive username authentication which is not supported by default.
-    """
-    
+class MongoEngineBackend(mongo_auth.MongoEngineBackend):
+    supports_object_permissions = True
```
mitar: But you have to provide the code for all this things you are saying we are supporting!

piplmesh/account/backends.py
agiz, line 62: Fixed automagically with get_or_create().

mitar, line 62: This goes for gender.

piplmesh/account/backends.py
mitar, line 56: To `auth_models.User.objects.create_user(username=username, email=fb['email'])`.

piplmesh/settings.py
agiz: Yes and no respectively. Database ENGINE has to be defined even if it is not used.

mitar: Why? What happens if it is not? We do not want to use any Django application which requires relational database.

mitar: (You should probably remove some from `INSTALLED_APPS` setting.)

piplmesh/account/models.py
mitar: I do not understand why is copy-pasting necessary? You ought to be able to just extend the User document from mongoengine `auth.py` and add additional fields and methods. Why is this not working?

mitar: This is the whole idea of using mongoengine so that we can use extensibility of classes.

agiz: I stand very weak here. From what I surmised you cannot extend mongoengine Documents(?). The farthest I came in this direction is to add new Class and referenced a user there to mongoengine User, but then there are other issues.

Also I want to know if do you think it can be done (extending mongoengine User) or know it can be done? If you know it can be done, please give some advice, because I'm running out of resources on this matter.

piplmesh/settings.py
agiz: django.core.exceptions.ImproperlyConfigured: settings.DATABASES is improperly configured. Please supply the ENGINE value. Check settings documentation for more details.

or 

django.core.exceptions.ImproperlyConfigured: You haven't set the DATABASE_ENGINE setting yet.

agiz: Tried removing various combinations of apps in INSTALLED_APPS to no success.

piplmesh/account/models.py
mitar: I know it can be done. It is even in the [tutorial](http://mongoengine.org/docs/v0.5/tutorial.html#posts). This is one of the main features of mongoengine. Because MongoDB is schemaless you can store there anything you want. Those fields are just so that mongoengine knows how to map values in MongoDB to Python object values and to do validation and such.

So you should really be able to simply extend provided User document with additional fields and make our own. And then add additional methods (you can simply hijack original methods) there.

piplmesh/account/models.py
mitar: You can also hijack all other "useful" methods from Django User object in this way. ;-)

piplmesh/settings.py
mitar: You also have to remove `django.middleware.transaction.TransactionMiddleware`!

mitar: I am using:

```
INSTALLED_APPS = (
    'piplmesh.frontend',
    'piplmesh.account',

    'django.contrib.sessions',
    'django.contrib.messages',
    'django.contrib.staticfiles',
)
```

piplmesh/account/models.py
agiz: But I cannot extend mongoengine User. By extending I am thinking of ExtendedMongoUser(mongo.User). 

I can add another class AdditionalUserData(Document) and reference a user there but I cannot return/pass that class in backends because it is not an instance of mongoengine User. I mean I can still add the data to AdditionalUserData but I can only return default mongoengine User then. 

What bothers me is how can we access these additional user data from a template? With current (I agree) crappy solution is as easy as user.gender.

piplmesh/settings.py
mitar: Of course this should happen only if you want to access relational database. I don't believe we need that anywhere. So you can remove that. And if you get above exceptions, this just means we are polluted somewhere with relational database code and should fix it. ;-)

agiz: Thanks, running without relational database.

piplmesh/account/models.py
mitar: I can extend? I just get a warning:

```
FutureWarning: User uses inheritance, the default for allow_inheritance is changing to off by default.  Please add it to the document meta.
```

The only thing I have done (for testing, you should be more precise) is to:

```

class User(auth.User):
    birthdate = DateTimeField()
    gender = StringField(max_length=6)
    facebook_id = IntField()
    facebook_token = StringField(max_length=150)
```

So removed all parent fields. And I could register?

agiz: You can register and login? I cannot :/. For starters it raises "settings.DATABASES is improperly configured. Please supply the ENGINE value. Check settings documentation for more details." error again

mitar: No. I can register with extended User document and having DATABASES setting. But I can also open the webpage if I remove the DATABASES setting. But registration does not yet work with removed DATABASES setting. This is something you have to fix. (Check the traceback of the exception and you will see who wants to access the database.)

mitar: https://github.com/hmarr/mongoengine/issues/476

agiz: Heh, neat. At least we are on the same boat now :). 

Can you check https://github.com/mitar/PiplMesh/blob/master/piplmesh/account/views.py RegistrationView. Why did this got implemented together with https://github.com/mitar/PiplMesh/blob/master/piplmesh/account/forms.py RegistrationForm? Previously def registration_view(request): redirected request to RegistrationForm but is RegistrationForm still needed? 
All that is left is traces in form_class = forms.RegistrationForm (https://github.com/mitar/PiplMesh/blob/master/piplmesh/account/views.py) which seems redundant now. I do not know which form to fix to get it working with mongoengine again.

mitar: I don't understand the problem? Something is a form and something is a view? Those are two different things. ;-)

Now we are using class-based views so it is enough just to tell which form to use and everything else is done automatically.

mitar: The problem is that our `RegistrationForm` extends `auth_forms.UserCreationForm`. And that the later is using relation database internally.

mitar: _What bothers me is how can we access these additional user data from a template? With current (I agree) crappy solution is as easy as user.gender._ << Also when you extend the provided User document with ours you can access this in the same manner. This mongoengine makes sure that all fields are simply accessible.

agiz: ad 1:
I cannot understand how RegistrationView with form_class = forms.RegistrationForm is doing all this:

```
     form = forms.RegistrationForm(request.POST)
     if form.is_valid():
         username, password = form.save()
         new_user = auth.authenticate(username=username, password=password)
         auth.login(request, new_user)
```

ad 2: thanks for the tip

ad 3: Yeah with that kind of implementation (the kind I could not get to work with ME 0.6.2(?)) it is all clear. I was referring that question to implementation when there is a separate class Foo(Document) that stores additional user data but my question is now irrelevant.

mitar: re ad 1: I think you should read discussion on that pull request we introduced [class-based views](https://docs.djangoproject.com/en/dev/ref/class-based-views/#formmixin). ;-) #89 #79

re ad 3: Yes, but the main idea of going to MongoDB was to have only one object/model/class for users and not user + profile. ;-)

agiz: It is working :). Will update in a minute.

mitar: Will see ... I want that you still extend things! ;-)

piplmesh/account/backends.py
agiz, line 56: Changed. Also masked as fb.get('email', fb.get('id')).

piplmesh/account/backends.py
agiz, line 62: Hmm... I did not code that (but that is not an excuse) so hopefully this is already tested well. Cannot do tests myself as I do not have FB.

piplmesh/account/backends.py
mitar, line 56: Why is this good?

mitar, line 56: id != email?

piplmesh/account/forms.py
agiz, line 11: Is this meant for from piplmesh.account import models?

piplmesh/account/forms.py
agiz: Fixed.

piplmesh/account/forms.py
agiz: Removed  # Form settings comment.

piplmesh/account/forms.py
agiz, line 61: language is still missing in this pull, sorry.

piplmesh/account/backends.py
mitar, line 62: OK. Then leave it as it was. So let it throw exceptions. Will somebody else check/test this. Then do not convert to `get`. Or at least make smart decisions when to do get and when not.

piplmesh/account/models.py
agiz: Fixed.

piplmesh/account/models.py
agiz: User in models.py now extends mongoengine.django.auth.User.

piplmesh/settings.py
agiz: Changed to settings.

piplmesh/settings.py
agiz, line 9: Done.

piplmesh/settings.py
agiz: Removed CUSTOM_USER_MODEL = 'accounts.User'.

piplmesh/settings.py
agiz, line 151: Restored to default.

piplmesh/account/forms.py
mitar, line 11: No. ;-) For import of mongoengine.django.auth.

piplmesh/account/forms.py
mitar, line 61: You will have to merge it sooner or later. I don't care. ;-)

piplmesh/account/backends.py
agiz: Not sure what to do here. Changed all these variables to False and added #TODO: comments.

piplmesh/account/backends.py
agiz, line 56: I saw fb.get('username', fb.get('id') and just followed the principle. 

fb.get('email', fb.get('email')) is ok?

piplmesh/account/backends.py
mitar: You should read documentation and read official implementation. ;-)

But OK, put TODO comments.

piplmesh/account/backends.py
mitar, line 56: Now I will ask you to think a bit before ask stupid questions. So please think what code does. And what are you doing.

piplmesh/account/forms.py
agiz, line 61: This pull is already merged with current revision, although language is still missing :)

mitar, line 61: Then language is there, no?

mitar, line 61: Or you have removed it?

agiz, line 61: language is where exactly? It is certainly not in the current forms.py https://github.com/mitar/PiplMesh/blob/master/piplmesh/account/forms.py.

Edit: Yes, language is in models.py. But I have changed this file.. Update soon to follow.

mitar, line 61: OK. It is not in the form. Just in the model.

mitar, line 61: (This has not been finished because we were waiting for change to MongoDB.)

piplmesh/account/backends.py
agiz, line 56: That took me way too much time to figure that out. Hope no one reads this.

piplmesh/account/forms.py
agiz, line 61: Shall I add language field in models.py? MongoDB field type should be StringField, no?

piplmesh/account/backends.py
```diff
@@ -1,25 +1,39 @@
 import json, urlparse, urllib
 
 from django.conf import settings
-from django.contrib.auth import backends, models as auth_models
 from django.core import urlresolvers
 
+from mongoengine.django import auth as mongo_auth
```
mitar: Just leave it `auth`.

piplmesh/account/backends.py
```diff
@@ -1,25 +1,39 @@
...
 
-class CaseInsensitiveModelBackend(backends.ModelBackend):
-    """
-    This backend uses case-insensitive username authentication which is not supported by default.
-    """
-    
+class MongoEngineBackend(mongo_auth.MongoEngineBackend):
+    # TODO:
+    supports_object_permissions = False
+    # TODO: implement anonymous user backend
```
mitar: Start with the upper case letter.

piplmesh/account/backends.py
```diff
@@ -1,25 +1,39 @@
...
+
 from piplmesh.account import models
 
-class CaseInsensitiveModelBackend(backends.ModelBackend):
-    """
-    This backend uses case-insensitive username authentication which is not supported by default.
-    """
-    
+class MongoEngineBackend(mongo_auth.MongoEngineBackend):
+    # TODO:
```
mitar: Missing TODO description.

piplmesh/account/forms.py
mitar, line 61: Yes. You should make sure all fields from models are included in the new user document. You can create your own field for language, similar to how it is done for relation database.

piplmesh/account/backends.py
mitar, line 56: We will show it on Wednesday, don't worry. ;-)

piplmesh/account/forms.py
agiz, line 61: Added language field.

piplmesh/account/backends.py
agiz: Changed to just auth.

piplmesh/account/backends.py
agiz: implement -> Implement

piplmesh/account/backends.py
agiz: Added TODO description.

piplmesh/account/forms.py
```diff
@@ -25,10 +23,11 @@ class HorizontalRadioRenderer(forms.RadioSelect.renderer):
     def render(self):
         return safestring.mark_safe(u'\n'.join([u'%s\n' % widget for widget in self]))
 
-class RegistrationForm(auth_forms.UserCreationForm):
+class RegistrationForm(forms.Form):
     """
     Class with user registration form.
     """
+    print "RegistrationForm"
```
mitar: Remove.

piplmesh/account/backends.py
```diff
@@ -44,39 +58,9 @@ def authenticate(self, token=None, request=None):
...
-            user = auth_models.User.objects.create_user(username=username, email=fb['email'])
-            user.first_name = fb['first_name']
-            user.last_name = fb['last_name']
-            user.save()
-
-            # Create and save account user
-            profile = models.UserProfile.objects.create(user=user, token=access_token, facebook_id=fb['id'], gender=fb['gender'])
-            profile.save()
+        
+        user, created = self.user_class.objects.get_or_create(facebook_id=fb.get('id'), defaults={'username': fb.get('username'), 'first_name': fb.get('first_name'), 'last_name': fb.get('last_name'), 'email': fb.get('email'), 'gender': fb.get('gender')})
```
mitar, line 64: I think we should move this to User document or manager? So that all code for creating users is in one place and not that we have API for creating users but we then do it manually somewhere.

piplmesh/account/models.py
```diff
@@ -1,21 +1,11 @@
...
-    language = fields.LanguageField(verbose_name=u'language')
-    
-    def __unicode__(self):
-        return u'%s' % (self.user)
+class User(auth.User):
+    birthdate = mongoengine.DateTimeField()
+    gender = mongoengine.StringField(max_length=6)
+    facebook_id = mongoengine.IntField()
+    facebook_token = mongoengine.StringField(max_length=150)
+    language = mongoengine.StringField(max_length=5)
```
mitar: Define our own field type which is limited by [choices](http://mongoengine.org/docs/v0.4/guide/defining-documents.html#field-arguments).

piplmesh/settings.py
```diff
@@ -13,20 +16,9 @@
...
-    'default': {
-        'ENGINE': 'django.db.backends.sqlite3', # Add 'postgresql_psycopg2', 'postgresql', 'mysql', 'sqlite3' or 'oracle'.
-        'NAME': database_file,                  # Or path to database file if using sqlite3.
-        'USER': '',                      # Not used with sqlite3.
-        'PASSWORD': '',                  # Not used with sqlite3.
-        'HOST': '',                      # Set to empty string for localhost. Not used with sqlite3.
-        'PORT': '',                      # Set to empty string for default. Not used with sqlite3.
-    }
-}
+#ADMINS = ()
```
mitar: Why commented this out? Leave it in.

piplmesh/settings.py
```diff
@@ -13,20 +16,9 @@
...
+
+#MANAGERS = ADMINS
```
mitar: Same.

piplmesh/settings.py
```diff
@@ -3,8 +3,11 @@
 # Development Django settings for PiplMesh project.
 
 import os.path
+
+import mongoengine
+mongoengine.connect('settings')
```
mitar: ???

I think there was some misunderstanding here. ;-)

piplmesh/account/forms.py
```diff
@@ -25,10 +23,11 @@ class HorizontalRadioRenderer(forms.RadioSelect.renderer):
     def render(self):
         return safestring.mark_safe(u'\n'.join([u'%s\n' % widget for widget in self]))
 
-class RegistrationForm(auth_forms.UserCreationForm):
+class RegistrationForm(forms.Form):
```
mitar: Be careful. Then you have to implement all cleaning/checking methods. I would still extend from UserCreationForm just check in the code which methods access database and override them with our own implementation.

piplmesh/account/models.py
```diff
@@ -1,21 +1,11 @@
...
-    gender = django_models.CharField(max_length=6, blank=True)
-    facebook_id = django_models.BigIntegerField(verbose_name=u'Facebook ID', null=True, blank=True)
-    token = django_models.CharField(max_length=150)
-    language = fields.LanguageField(verbose_name=u'language')
-    
-    def __unicode__(self):
-        return u'%s' % (self.user)
+class User(auth.User):
+    birthdate = mongoengine.DateTimeField()
+    gender = mongoengine.StringField(max_length=6)
```
mitar: Same. It should be limited to possible values. You could also do our own field type too.

piplmesh/account/models.py
```diff
@@ -1,21 +1,11 @@
...
-    birthdate = django_models.DateField(blank=True, null=True)
-    gender = django_models.CharField(max_length=6, blank=True)
-    facebook_id = django_models.BigIntegerField(verbose_name=u'Facebook ID', null=True, blank=True)
-    token = django_models.CharField(max_length=150)
-    language = fields.LanguageField(verbose_name=u'language')
-    
-    def __unicode__(self):
-        return u'%s' % (self.user)
+class User(auth.User):
+    birthdate = mongoengine.DateTimeField()
```
mitar: Some checking could be done here? Like that it cannot be in the future?

piplmesh/account/models.py
```diff
@@ -1,21 +1,11 @@
...
+    gender = mongoengine.StringField(max_length=6)
+    facebook_id = mongoengine.IntField()
+    facebook_token = mongoengine.StringField(max_length=150)
```
mitar, line 14: Please order them in the end. So like profile fields are first and then special fields are last. (You can also put one empty line inbetween.)

piplmesh/settings.py
```diff
@@ -190,10 +176,10 @@
 
 LOGIN_REDIRECT_URL = '/'
 
-AUTH_PROFILE_MODULE = 'account.UserProfile'
+SESSION_ENGINE = 'mongoengine.django.sessions'
 
 AUTHENTICATION_BACKENDS = (
```
mitar, line 183: Where has the lazy signup disappeared?

piplmesh/account/models.py
agiz: I guess I have to use fields.py and extend mongoengine.StringField? Or do you mean to just add field arguments to mongoengine.StringField in models.py?

mitar: First. Like I had done for `LanguageField` for relation database.

piplmesh/settings.py
agiz, line 183: That wasn't me? https://github.com/mitar/PiplMesh/blob/master/piplmesh/settings.py

mitar, line 183: Strange, but then we still have to develop that. ;-) In some other pull request. Leave it as it is and fix other issues.

piplmesh/account/models.py
agiz: Ok, but I am stuck at this obstacle:

```
super(mongoengine.StringField, self).__init__(*args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'max_length'
```

I really have no idea how to fix this.

mitar: Can you give me more code?

agiz: everything is the same as in current fields.py except:

```
class LanguageField(mongoengine.StringField):
```

and

```
super(mongoengine.StringField, self).__init__(*args, **kwargs)
```

and get_internal type is set to return "StringField"

in models.py

```
```

Edit: add internal_type comment

mitar: Can you please copy exactly as it is? It is much easier to debug when I can see everything together. Especially with "strange bugs".

agiz: Ok, I will update pull request.

mitar: Eh, no. You can just write here in the comments. It is just few lines. I am interested only in the `LanguageField` definition.

agiz: Too late :(

also:

```
import mongoengine

def get_initial_language(request=None):
    return settings.LANGUAGE_CODE

class LanguageField(mongoengine.StringField):
        kwargs.setdefault('max_length', 5)
        kwargs.setdefault('choices', settings.LANGUAGES)
        kwargs.setdefault('default', get_initial_language)

        super(mongoengine.StringField, self).__init__(*args, **kwargs)

    def get_internal_type(self):
        return "StringField"
```

piplmesh/account/fields.py
```diff
@@ -1,16 +1,17 @@
...
 
-class LanguageField(fields.CharField):
+class LanguageField(mongoengine.StringField):
     def __init__(self, *args, **kwargs):
         kwargs.setdefault('max_length', 5)
         kwargs.setdefault('choices', settings.LANGUAGES)
         kwargs.setdefault('default', get_initial_language)
         
-        super(fields.CharField, self).__init__(*args, **kwargs)
+        super(mongoengine.StringField, self).__init__(*args, **kwargs)
```
mitar:
```
super(LanguageField, self).__init__(*args, **kwargs)
```

It seems I had the bug in original code. ;-)

piplmesh/account/fields.py
```diff
@@ -1,16 +1,17 @@
...
         
     def get_internal_type(self):
```
mitar: This does not exist for mongoengine fields.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,17 @@
+import mongoengine
```
mitar: Order!!!

agiz: I thought import prefix imports go first.
Edit: Fixed now.

piplmesh/settings.py
agiz: The it still is?

I am confused do you want it in this order:

```
import mongoengine
mongoengine.connect('settings')

import os.path
```

piplmesh/settings.py
agiz: Uncommented.

piplmesh/settings.py
agiz: Uncommented.

piplmesh/account/fields.py
mitar: No. This is not important (the syntax). The content is important. First comes built-in libraries (os, datatime, ...), then Django, then other libraries (like mongoengine), then ours.

piplmesh/settings.py
mitar: No. Order is OK. Just `'settings'` is not. :-)

I thought that you would do:

```
MONGODB_DATABASE = 'PiplMesh'

import mongo
mongoengine.connect(MONGODB_DATABASE)
```

But this is still not the best (we would like a way to override the setting before the connection is made), but for now it will be OK.

agiz: I can move mongoengine.connect(MONGODB_DATABASE) at the end if you wish.

```
import os.path

import mongoengine

MONGODB_DATABASE = 'PiplMesh'

mongoengine.connect(MONGODB_DATABASE)

settings_dir = os.path.abspath(os.path.dirname(__file__))
```

piplmesh/account/models.py
agiz: Added TODO for now. This cannot be done with "just" mongoengine constraints right?

piplmesh/account/models.py
agiz, line 14: Reordered fields and added new line.

piplmesh/account/fields.py
agiz: Removed.

piplmesh/settings.py
mitar: This would not help. We will talk about this when we will be preparing production settings.

piplmesh/account/forms.py
mitar: You can move this to fields.py and use it for gender field choices and then use it also for form.

piplmesh/account/models.py
mitar: You just extend `validate` method (first call parent, then continue with range checks). You can also create maybe a generic field which can take upper date as argument and checks against that and then you set this parameter to current day, for example.

For such things it is always the best to check how existing (provided/official) fields are defined.

mitar: https://code.djangoproject.com/ticket/18034

agiz: Added fields.DateTimeField stub in fields.py.

piplmesh/account/models.py
agiz: Changed to fields.GenderField, implemented in fields.py.

agiz:
This pull is broken.

```
'super' object has no attribute 'clean_username'

return super(RegistrationView, self).dispatch(request, *args, **kwargs) - views.py in dispatch
```

I can only go as far as to change

```
username = super(RegistrationForm, self).clean_username()
```

in forms.py to

```
username = self.cleaned_data['username']
```

that leads to:

```
Errors encountered validating document
```

in froms.py

```
username, password = form.save() 
```

I guess this is because I am extending RegistrationForm(forms.Form) (and not django.contrib.auth.forms. UserCreationForm)? If so it's weird because this did work yesterday.

piplmesh/account/backends.py
mitar, line 64: This is not good. If  there is no `id` in `fb` there should be some error, not just silently try to store `None` (the default) into the database. And same for others.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,34 @@
...
         kwargs.setdefault('default', get_initial_language)
         
-        super(fields.CharField, self).__init__(*args, **kwargs)
+        super(LanguageField, self).__init__(*args, **kwargs)
+
+class GenderField(mongoengine.StringField):
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('max_length', 6)
+        kwargs.setdefault('choices', GENDER_CHOICES)
+        kwargs.setdefault('default', None)
```
mitar: This is not needed.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,34 @@
...
+        
+        super(GenderField, self).__init__(*args, **kwargs)
+
+class DateTimeField(mongoengine.DateTimeField):
```
mitar: Rename to `LimitedDateTimeField`.

piplmesh/account/fields.py
agiz: Hardcoded to

```
kwargs.setdefault('default', GENDER_CHOICES[1])
```

not sure if this is ok.

piplmesh/account/fields.py
agiz: Renamed.

piplmesh/account/backends.py
agiz, line 64: Added TODO. Before fixing this, you will have to explain first part to me like I am 5. I am not sure where and how shall we move this.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,34 @@
...
         kwargs.setdefault('default', get_initial_language)
         
-        super(fields.CharField, self).__init__(*args, **kwargs)
+        super(LanguageField, self).__init__(*args, **kwargs)
+
+class GenderField(mongoengine.StringField):
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('max_length', 6)
+        kwargs.setdefault('choices', GENDER_CHOICES)
+        kwargs.setdefault('default', GENDER_CHOICES[1])
```
mitar: Probably not. ;-) `GENDER_CHOICES[0][0]` would be more correct. ;-)

But maybe it is not even needed. Try without and we will see how it goes.

agiz: I cannot. This pull got broken somewhere in the meantime.

https://github.com/mitar/PiplMesh/pull/93#issuecomment-4858365

piplmesh/account/backends.py
mitar, line 64: OK, leave as TODO for now.

mitar:
Yes. Of course `forms.Form` do not have `clean_username`. Please read official code. This is the best way to learn things and also understand things. This is why open source is great. Because you can really understand everything. So as I said, you should extend that special form. (If it is possible.) Check how it is implemented and override as necessary.

agiz:
Well I will and already did some peeking at django forms but it's hard to persuade myself to look at the implementation of django forms. Also I love this kind of concept, documentation and source code and have zero problems learning things that are in my interest. However becoming django form expert is not one of them :-)

mitar:
But your interest is getting to become a better programmer: and reading code of experts is a good way. And it is not that you are becoming Django form expert, more Python expert. I also do not know everything about Django forms, but I know how to find what I need.

mitar:
More you limit yourself ("this I will never need, I do not need to learn") more you are limiting your potential. Because you never know how this knowledge will be useful. And it is not so hard to gain the knowledge once you are get used to check things when they fall in front of you (in a form of a traceback, for example).

agiz:
But still... Forms*? :)

*I know it's not about the forms but it's not funny without them.

mitar:
He he. Yes. If you are more comfortable: this will be one or rare forms we will have. We we use REST for the rest. ;-)

But more you know Django forms, more you know that things can be automated _a lot_.

mitar:
So, stop complaining, go back to work! I can assure you that my problems are greater than yours! :-)

agiz:
Pull is working once again. I will take a (not so well deserved) break now.

piplmesh/account/forms.py
agiz: Registration form now extends auth_forms.UserCreationForm yet again.

piplmesh/account/models.py
agiz: Added constraints for future dates.

piplmesh/account/backends.py
```diff
@@ -44,39 +58,10 @@ def authenticate(self, token=None, request=None):
...
-            user = auth_models.User.objects.create_user(username=username, email=fb['email'])
-            user.first_name = fb['first_name']
-            user.last_name = fb['last_name']
-            user.save()
-
-            # Create and save account user
-            profile = models.UserProfile.objects.create(user=user, token=access_token, facebook_id=fb['id'], gender=fb['gender'])
-            profile.save()
+        
+        # TODO: Check if id is returned
```
mitar: And check also for other fields.

piplmesh/account/backends.py
```diff
@@ -44,39 +58,10 @@ def authenticate(self, token=None, request=None):
...
+        user, created = self.user_class.objects.get_or_create(facebook_id=fb.get('id'), defaults={'username': fb.get('username'), 'first_name': fb.get('first_name'), 'last_name': fb.get('last_name'), 'email': fb.get('email'), 'gender': fb.get('gender')})
```
mitar, line 64: Add TODO to move this user creation (or get) into the User document/manager.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,41 @@
...
+
+class LimitedDateTimeField(mongoengine.DateTimeField):
+    def validate(self, value):
+        super(LimitedDateTimeField, self).validate(value)
         
-    def get_internal_type(self):
-        return "CharField"
\ No newline at end of file
+        try:
+            futureDate = value > datetime.datetime.today() - datetime.timedelta(settings.FUTUREDATE)
```
mitar: No, this should be an argument for `LimitedDateTimeField` at initalization. I do not understand the need of `FUTUREDATE` setting?

piplmesh/account/fields.py
```diff
@@ -1,16 +1,41 @@
...
+        except:
```
mitar: Never catch all exceptions because this could hide possible programming errors.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,41 @@
...
+            futureDate = value > datetime.date.today() - datetime.timedelta(settings.FUTUREDATE)
```
mitar: I am not sure why you repeat this again?

piplmesh/account/fields.py
```diff
@@ -1,16 +1,41 @@
...
+        if futureDate:
+            self.error(u'User born on "%s" is too young to register' % value)
```
mitar: Error message should be generic: like "Value is out of bounds"

piplmesh/account/forms.py
```diff
@@ -2,18 +2,13 @@
...
 from django.utils.translation import ugettext_lazy as _
 
-from piplmesh.account import models, fields
-
-# Form settings
-GENDER_CHOICES = (
-    ('male', _('Male')),
-    ('female', _('Female'))
-)
+from piplmesh.account import fields
```
mitar: Combine into one import line.

piplmesh/account/forms.py
```diff
@@ -51,27 +46,28 @@ def clean_password2(self):
...
-        # Then we asign profile to this user
-        profile = models.UserProfile(user=new_user, gender=self.cleaned_data['gender'], birthdate=self.cleaned_data['birthdate'])
-        profile.save()
 
-        return self.cleaned_data['username'], self.cleaned_data['password2']
\ No newline at end of file
+        return self.cleaned_data['username'], self.cleaned_data['password2']
+        
+    def validate_unique(self):
+        # TODO: check for errors
```
mitar: First letter upper case in TODO.

piplmesh/account/models.py
```diff
@@ -1,21 +1,12 @@
...
-    user = django_models.OneToOneField(auth_models.User)
-
-    # Custom fields
-    birthdate = django_models.DateField(blank=True, null=True)
-    gender = django_models.CharField(max_length=6, blank=True)
-    facebook_id = django_models.BigIntegerField(verbose_name=u'Facebook ID', null=True, blank=True)
-    token = django_models.CharField(max_length=150)
-    language = fields.LanguageField(verbose_name=u'language')
+class User(auth.User):
+    birthdate = fields.LimitedDateTimeField()
```
mitar: You should pass here the limits for the date.

piplmesh/account/fields.py
agiz: It's not repeated datetime.date.today() != datetime.datetime.today().

As I tested, value is type of datetime.date. Can I assume it is always that type and never datetime.datetime?

mitar: No, you cannot. Check the parent `validate`. You will see that there it allows both.

Now I see what you want to do. Don't use exceptions for normal program flow. In this case, simply check the type of the value and do one thing or another. Without exceptions.

You do not need to complicate with `FUTUREDATE` offset. We just want some preliminary check on the birthday.

agiz: I am not sure I understand completely what to do here. If I pass the variable for `FUTUREDATE` offset through LimitedDateTimeField in models.py I have to override `__init__` in LimitedDateTimeField right and then somehow clean that value off of kwargs, store it in a class variable so `validate()` method can access it, and pass the rest to the parent class.

mitar: Exactly, you will have to override `__init__`. But you do not need to pass `FUTUREDATE`. We do not need this. Just pass the `upper_limit` value, which is some date. And you can do simply:

```
def __init__(self, upper_limit=None, lower_limit=None, *args, **kwargs):
   self.upper_limit = upper_limit
   self.lower_lmit = lower_limit
   super(LimitedDateTimeField, self).__init__(*args, **kwargs)
```

piplmesh/account/backends.py
agiz: Added TODOs.

piplmesh/account/backends.py
agiz, line 64: Added TODO.

piplmesh/account/fields.py
agiz: Reimplemented.

piplmesh/account/fields.py
agiz: Implemented.

piplmesh/account/fields.py
agiz: Fixed.

piplmesh/account/forms.py
agiz: Fixed.

piplmesh/account/forms.py
agiz: Fixed.

piplmesh/account/models.py
agiz: LimitedDateTimeField now takes limiting dates arguments. Not sure if this is ok...

piplmesh/account/fields.py
```diff
@@ -1,16 +1,54 @@
...
+class LimitedDateTimeField(mongoengine.DateTimeField):
+    def __init__(self, upper_limit=None, lower_limit=None, *args, **kwargs):
+        self.upper_limit = upper_limit
+        self.lower_limit = lower_limit
+        super(LimitedDateTimeField, self).__init__(*args, **kwargs)
+   
+    def validate(self, value):
+        super(LimitedDateTimeField, self).validate(value)
+        
+        # TODO: This is not needed right?
```
mitar: You should check validity of the arguments in `__init__` so that they are checked only once and not for every `validate` run. Also exceptions will be thrown correctly at field initialization and not field validation.

Arguments can of course be `None`.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,54 @@
...
+        if self.upper_limit and not isinstance(self.upper_limit, (datetime.datetime, datetime.date)):
+            self.error(u'Invalid limit.')
+        if self.lower_limit and not isinstance(self.lower_limit, (datetime.datetime, datetime.date)):
+            self.error(u'Invalid limit.')
         
-    def get_internal_type(self):
-        return "CharField"
\ No newline at end of file
+        if isinstance(value, datetime.datetime):
+            futureDate = value > self.upper_limit
```
mitar: `futureDate` is probably badly named now? Where do you initialize it, BTW? And why do you use it? Simply call `self.error` when you notice an error condition (it throws an exception).

piplmesh/account/fields.py
```diff
@@ -1,16 +1,54 @@
...
+            futureDate |= value < self.lower_limit
+        elif isinstance(value, datetime.date):
+            futureDate = value > self.upper_limit.date()
```
mitar: Check only limits if they are defined. Does `date()` work both on `datetime.datetime` and `datetime.date` objects?

piplmesh/account/forms.py
```diff
@@ -51,27 +45,28 @@ def clean_password2(self):
...
     def clean_username(self):
         # This method checks whether the username exists in case-insensitive manner
-        username = super(RegistrationForm, self).clean_username()
-        try:
-            auth_models.User.objects.get(username__iexact=username)
-        except auth_models.User.DoesNotExist:
-            return username
-        raise forms.ValidationError(_("A user with that username already exists."))
+        username = self.cleaned_data['username']
+        if models.User.objects(username__iexact=username).first():
```
mitar: Maybe `.count()` is better to use if you want only to check for existence?

piplmesh/account/fields.py
agiz: Unfortunately not.

mitar: I thought so. :-) So you cannot use this like you are doing. :-)

agiz: I can, upper_limit is instance of datetime.datetime.

mitar: Not always? You are allowing both `date` and `datetime`:

```
if self.upper_limit and not isinstance(self.upper_limit, (datetime.datetime, datetime.date)):
```

agiz: Sorry for derail :). My mistake.

piplmesh/account/forms.py
agiz: Fixed.

piplmesh/account/fields.py
agiz: Moved bounds checking to `__init__`.

piplmesh/account/fields.py
agiz: Hm... This is still not good. 

Is this a bug?

```
a = datetime.datetime.today()
isinstance(a, datetime.datetime) # True
isinstance(a, datetime.date)     # True
a > datetime.date.today()        # TypeError: can't compare datetime.datetime to datetime.date
```

Shouldn't inherited  instances be comparable?

Edit: considered fixed?

piplmesh/account/fields.py
agiz: Reimplemented.

piplmesh/account/fields.py
mitar: Seems not. :-)

agiz: As in not fixed?

mitar: Don't edit your comment and you will easier understand what I am talking about. :-)

piplmesh/account/fields.py
```diff
@@ -1,16 +1,53 @@
+import datetime
+
 from django.conf import settings
-from django.db.models import fields
+from django.utils.translation import ugettext_lazy as _
+
+import mongoengine
+
+GENDER_CHOICES = (
+    ('male', _('Male')),
```
mitar: Ah, btw, this could be lowercased. Because upper case is often then done in the template, if wanted.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,53 @@
...
+        kwargs.setdefault('default', GENDER_CHOICES[1][0])
+        
+        super(GenderField, self).__init__(*args, **kwargs)
+
+class LimitedDateTimeField(mongoengine.DateTimeField):
+    def __init__(self, upper_limit=None, lower_limit=None, *args, **kwargs):
+        self.upper_limit = upper_limit
+        self.lower_limit = lower_limit
+
+        # TODO: This is not needed right?
```
mitar: Remove TODO?

piplmesh/account/fields.py
```diff
@@ -1,16 +1,53 @@
...
+        if self.upper_limit and not isinstance(self.upper_limit, datetime.datetime):
```
mitar: Allow also `datetime.date`?

piplmesh/account/fields.py
```diff
@@ -1,16 +1,53 @@
...
+            self.error(u'Invalid limit.')
```
mitar: Be more verbose: _Invalid upper_limit arguments._

piplmesh/account/fields.py
```diff
@@ -1,16 +1,53 @@
...
+        super(LimitedDateTimeField, self).__init__(*args, **kwargs)
+   
+    def validate(self, value):
+        super(LimitedDateTimeField, self).validate(value)
         
-    def get_internal_type(self):
-        return "CharField"
\ No newline at end of file
+        if isinstance(value, datetime.datetime)
+            value = value.date()
```
mitar: This is not good. You are then comparing only dates. But the field can be also with time. We want to support a general use case, not just ours which is for birthday and thus only a date.

piplmesh/account/fields.py
agiz: Then I should only check for `datetime.date`. Because `datetime.datetime` is an instance of `datetime.date` and hence only checking against `datetime.date` is needed.

`isinstance(self.upper_limit, (datetime.datetime, datetime.date))` is no different from `isinstance(self.upper_limit, datetime.date)`. I am only allowing `datetime.datetime` because that way I am sure what type of object I am dealing with.

piplmesh/account/fields.py
agiz: Relevant with https://github.com/mitar/PiplMesh/pull/93#r637870.

piplmesh/account/fields.py
agiz: Removed.

piplmesh/account/fields.py
agiz: Changed to lowercase.

piplmesh/account/fields.py
mitar: I would still suggest you check for both because then it is clear to the reader of the code (who might don't know about inheritance, as you didn't know) what you are allowing/expecting and what you don't. But comment the reason (code readability) so that later on somebody will not "optimize".

Or, you could add a comment explaining that `datetime.datetime` is a subclass of `datetime.date` so it is enough to check only for the later.

piplmesh/account/fields.py
mitar: Still this is not good. If you convert to date then only dates (and not hours, minutes, ...) are compared.

agiz: Yes, I am only comparing date-wise. Maybe I took this too lightly. Is hours and minutes comparing really important?

piplmesh/account/fields.py
agiz: Rewritten.

piplmesh/account/fields.py
mitar: If the user specifies limit as `datetime`, we should respect that.

agiz: What you want is compare with hours and minutes if the `value` is type of `datetime.datetime` and compare days, months, etc, if the `value`is type of `datetime.date`?

mitar: I just want to compare with all the information available. If limit is datetime and value is datetime, then also hours and minutes should be compared. Otherwise only date.

agiz: This will get ugly... There are so many cases then. What is one limit is type of `datetime.datetime` and the other `datetime.date`... I am trying to figure out how to write this as cleanly as possible.

agiz: Is this implementation now ok?

piplmesh/account/fields.py
agiz: Fixed.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,65 @@
...
+        
+        super(GenderField, self).__init__(*args, **kwargs)
+
+class LimitedDateTimeField(mongoengine.DateTimeField):
+    def __init__(self, upper_limit=None, lower_limit=None, *args, **kwargs):
+        self.upper_limit = upper_limit
+        self.lower_limit = lower_limit
+
+        if self.upper_limit and not isinstance(self.upper_limit, (datetime.datetime, datetime.date)):
+                self.error(u'Invalid upper_limit argument.')
```
mitar: Indentation.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,65 @@
...
+        super(LimitedDateTimeField, self).validate(value)
         
-    def get_internal_type(self):
-        return "CharField"
\ No newline at end of file
+        if self.upper_limit:
+            tmp_value = value
+            if not isinstance(value, datetime.datetime) or not isinstance(self.upper_limit, datetime.datetime):
+                if isinstance(self.upper_limit, datetime.datetime):
+                    self.upper_limit = self.upper_limit.date()
```
mitar: Don't change passed in argument. Use some local variable.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,65 @@
...
+                elif isinstance(value, datetime.datetime):
```
mitar, line 53: Shouldn't here be:

```
if isinstance(tmp_value, datetime.datetime):
```

agiz, line 53: Can you explain a bit why... Both values are still the same at this point, however if you move down to `lower_limit` checks, tmp_value cannot be used since it is potentially dirty.

agiz, line 53: Err... it is not, because I re-set it. But still, what is the issue here?

piplmesh/account/fields.py
agiz: Using local variables now.

piplmesh/account/fields.py
agiz: Fixed, sorry.

piplmesh/account/fields.py
mitar, line 53: The point it is that the code is more readable. It is more local. If you assign some variable then use that variable, not use original, because it is easier to follow one variable than two. So once you assign, it is easier to remember that from now on we are dealing with just that one temporary copy and this is it.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,67 @@
...
-    def get_internal_type(self):
-        return "CharField"
\ No newline at end of file
+        if self.upper_limit:
+            tmp_value = value
+            tmp_upper_limit = self.upper_limit
+            if not isinstance(value, datetime.datetime) or not isinstance(self.upper_limit, datetime.datetime):
+                if isinstance(self.upper_limit, datetime.datetime):
+                    tmp_upper_limit = self.upper_limit.date()
+                elif isinstance(value, datetime.datetime):
```
mitar, line 53: Please fix both, use only `tmp_*`. It is OK to use `elif`, forget about my previous comment.

piplmesh/account/models.py
```diff
@@ -1,21 +1,14 @@
...
+from piplmesh.account import fields
 
-    # Custom fields
-    birthdate = django_models.DateField(blank=True, null=True)
-    gender = django_models.CharField(max_length=6, blank=True)
-    facebook_id = django_models.BigIntegerField(verbose_name=u'Facebook ID', null=True, blank=True)
-    token = django_models.CharField(max_length=150)
-    language = fields.LanguageField(verbose_name=u'language')
+class User(auth.User):
+    birthdate = fields.LimitedDateTimeField(upper_limit=datetime.datetime.today(), lower_limit=datetime.datetime.today() - datetime.timedelta(366 * 120))
```
mitar, line 9: For lower limit create a constant at the top of this module, and just hard-code the year, like 1900. Not years.

piplmesh/account/forms.py
```diff
@@ -2,18 +2,12 @@
```
mitar, line 2: OK, you will also have to sync form fields with model fields (like default value, limits, and so on).

What is `current_date` doing in this file? Useless.

PR #96, opened 2012-04-01
Issue#94

NejcZupec:
Team 1. Updates in requirements.txt


mitar:
I believe we are not using lazy signup anymore. And furthermore we want to get rid of it in favor of our own MongoDB-based version.

Issue #64, opened 2012-03-21
After logout, redirect user back


mitar:
After user logouts, she should be redirected back to the page she came from. Display to the user a message that logout was successful.

mitar:
Done with #101.

PR #98, opened 2012-04-03
Redirect user after logout

domenc:
Add JS for redirect user to previous page after logout - issue #64


mitar:
Not good. The idea is that to user immediate page is never displayed, but they are immediately redirected back to the page when they come from, with a message displayed on that page that they have been logged in. See how it works in [nodewatcher](https://nodes.wlan-si.net/).

PR #102, opened 2012-04-06
Issue 61, Issue 62 rev-3 (team 4)

agiz:


piplmesh/account/formfields.py
```diff
@@ -0,0 +1,46 @@
+import datetime
+
+from django import forms
+
+class FormLimitedDateTimeField(forms.DateTimeField):
```
mitar: It can be just `LimitedDateTimeField`. We are always importing just modules, so we exactly know what we have (form field for model field).

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, upper_limit=None, lower_limit=None, input_formats=None, *args, **kwargs):
+        self.upper_limit = upper_limit
+        self.lower_limit = lower_limit
+
+        if self.upper_limit and not isinstance(self.upper_limit, (datetime.datetime, datetime.date)):
+            self.error(u'Invalid upper_limit argument.')
```
mitar: Hm, form fields do not have `error` method. This is something from mongoengine.

You should probably do something like `raise ValidationError(self.error_messages['bounds'])`, where you define the `bounds` error message in `default_error_messages` (extend/append new one).

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,46 @@
...
+        if self.lower_limit and not isinstance(self.lower_limit, (datetime.datetime, datetime.date)):
+            self.error(u'Invalid lower_limit argument.')
+
+        super(FormLimitedDateTimeField, self).__init__(input_formats=None, *args, **kwargs)
+   
+    def clean(self, value):
```
mitar, line 25: `clean`: https://docs.djangoproject.com/en/dev/ref/forms/fields/#django.forms.Field.clean

So you should return the value, if everything is OK.

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,46 @@
...
+        super(FormLimitedDateTimeField, self).clean(value)
```
mitar: And capture the value from parent's `clean`.

```
value = super(FormLimitedDateTimeField, self).clean(value)
```

Not `birthdate`! We are having general form field here!

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,46 @@
...
+        
+        try:
```
mitar: This is not needed then.

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,46 @@
...
+            birthdate = datetime.datetime.strptime(value, '%Y-%m-%dT%H:%M:%S')
+        except ValueError:
+            birthdate = datetime.datetime.strptime(value, '%Y-%m-%d')
+            
+        
+        if self.upper_limit:
+            tmp_value = birthdate
```
mitar: `tmp_value = value`

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,46 @@
...
+        if self.lower_limit:
+            tmp_value = birthdate
+            tmp_lower_limit = self.lower_limit
+            if not isinstance(birthdate, datetime.datetime) or not isinstance(self.lower_limit, datetime.datetime):
+                if isinstance(self.lower_limit, datetime.datetime):
+                    tmp_lower_limit = self.lower_limit.date()
+                elif isinstance(tmp_value, datetime.datetime):
+                    tmp_value = birthdate.date()
+            if tmp_value < tmp_lower_limit:
+                self.error(u'Value is out of bounds.')
```
mitar: At the end you should then return `value`.

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,46 @@
...
+        if self.upper_limit:
+            tmp_value = birthdate
+            tmp_upper_limit = self.upper_limit
+            if not isinstance(birthdate, datetime.datetime) or not isinstance(self.upper_limit, datetime.datetime):
+                if isinstance(self.upper_limit, datetime.datetime):
+                    tmp_upper_limit = self.upper_limit.date()
+                elif isinstance(tmp_value, datetime.datetime):
+                    tmp_value = birthdate.date()
+            if tmp_value > tmp_upper_limit:
+                self.error(u'Value is out of bounds.')
```
mitar: No `error` method. Have you even tried forms with out of the bounds values?

piplmesh/account/formfields.py
mitar: The other way with Django fields is that you just add additional [validator](https://docs.djangoproject.com/en/dev/ref/validators/) when creating new field. But it is OK to override also `clean`. Everything depends on the use case. If you have multiple locations when you want to use new field, then it is probably better to define a new field, if you just want to limit value in some way only once, then you could just create a simple method for that.

Leave it as it is here. Just to tell you.

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,46 @@
...
+    def clean(self, value):
+        super(FormLimitedDateTimeField, self).clean(value)
+        
+        try:
+            birthdate = datetime.datetime.strptime(value, '%Y-%m-%dT%H:%M:%S')
+        except ValueError:
+            birthdate = datetime.datetime.strptime(value, '%Y-%m-%d')
+            
+        
+        if self.upper_limit:
```
mitar: Anyway, I do not like that we are duplicating code here. No copy-paste please. Can you create a top-level function in models:

```
def limit_date(value, lower_limit, upper_limit, error):
```

which does everything you do bellow. On error, it calls `error` argument like `error()`. And then you can define `clean` like:

```
def clean(self, value):
    value = super(LimitedDateTimeField, self).clean(value)

    def error():
        raise ValidationError(self.error_messages['bounds'])

    limit_date(value, self.lower_limit, self.upper_limit, error):

    return value
```

piplmesh/account/formfields.py
agiz: Err.. Should I use `default_error_messages` or `error_messages` or are these two somehow related?
Also, what about translatables here?

mitar: It takes exactly 1.854 seconds to check this in the code.

piplmesh/account/formfields.py
agiz: Almost asked how to handle mongo/django form field errors then but by some freak coincidence figure it out just in time... You are testing me way too hard here. :)

agiz:
If module A requires module B and module B requires modul A I am in a deadlock situation right?
I moved limit_date together with GENDER_CHOICES and error_messages to models.py and I am trapped in above scenario. 

What is the next best place to move upper constants to?

mitar:
OK. Put then all this into `fields.py`. :-)

piplmesh/account/formfields.py
agiz: Renamed.

piplmesh/account/formfields.py
agiz: Fixed and made translatable.

piplmesh/account/formfields.py
agiz: Implemented as suggested above.

mitar:
Please write a comment when you finish, because I do not get notification otherwise.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,79 @@
...
+        
+        super(GenderField, self).__init__(*args, **kwargs)
+
+class LimitedDateTimeField(mongoengine.DateTimeField):
+    def __init__(self, upper_limit=None, lower_limit=None, *args, **kwargs):
+        self.upper_limit = upper_limit
+        self.lower_limit = lower_limit
+
+        if self.upper_limit and not isinstance(self.upper_limit, (datetime.datetime, datetime.date)):
+            self.error(error_messages['syntax_upper'])
```
mitar: This error is displayed only in the case of programming error. I would hard-code it and _not_ make it translatable (it is really hard to Google for errors).

So I would have only `bounds` up there.

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,28 @@
...
+        self.upper_limit = upper_limit
+        self.lower_limit = lower_limit
+
+        if self.upper_limit and not isinstance(self.upper_limit, (datetime.datetime, datetime.date)):
+            raise forms.ValidationError(fields.error_messages['syntax_upper'])
+        if self.lower_limit and not isinstance(self.lower_limit, (datetime.datetime, datetime.date)):
+            raise forms.ValidationError(fields.error_messages['syntax_lower'])
+
+        super(LimitedDateTimeField, self).__init__(input_formats=None, *args, **kwargs)
+
```
mitar: Unnecessary empty line.

piplmesh/account/forms.py
```diff
@@ -31,17 +25,13 @@ class RegistrationForm(auth_forms.UserCreationForm):
...
-    password2 = forms.CharField(widget=forms.PasswordInput, label=_("Repeat password"))
     first_name = forms.CharField(label=_("First name"))
     last_name = forms.CharField(label=_("Last name"))
     
     # Additional information
-    gender = forms.ChoiceField(label=_("Gender"), required=False, choices=GENDER_CHOICES, widget=forms.RadioSelect(renderer=HorizontalRadioRenderer))    
-    current_date = datetime.now()
-    birthdate = forms.DateField(label=_("Birth date"), required=False, widget=widgets.SelectDateWidget(years=[y for y in range(current_date.year, 1900, -1)]))
+    gender = forms.ChoiceField(label=_("Gender"), required=False, choices=fields.GENDER_CHOICES, widget=forms.RadioSelect(renderer=HorizontalRadioRenderer))    
+    birthdate = formfields.LimitedDateTimeField(upper_limit=datetime.datetime.today(), lower_limit=datetime.datetime.today() - datetime.timedelta(models.LOWER_DATE_LIMIT), label=_("Birth date"), required=False, widget=widgets.SelectDateWidget(years=[y for y in range(datetime.datetime.today().year, (datetime.datetime.today() - datetime.timedelta(models.LOWER_DATE_LIMIT)).year, -1)]))
```
mitar: Can you split those two fields into multiline passing of arguments? So

```
foo = Bar(
   bla=blu,
   ...,
)
```

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,28 @@
...
+            raise forms.ValidationError(fields.error_messages['syntax_lower'])
+
+        super(LimitedDateTimeField, self).__init__(input_formats=None, *args, **kwargs)
+   
+
+    def clean(self, value):
+        value = super(LimitedDateTimeField, self).clean(value)
+
+        def error():
+            raise forms.ValidationError(fields.error_messages['bounds'])
```
mitar: No, it should be:

```
raise forms.ValidationError(self.error_messages['bounds'])
```

And now make it work like this. :-) (Hint: Set `default_error_messages` attribute of `LimitedDateTimeField`.)

piplmesh/account/formfields.py
agiz: Removed.

piplmesh/account/fields.py
agiz: I am not sure I understand what exactly to do here... Is this ok?

piplmesh/account/forms.py
agiz: Done.

piplmesh/account/formfields.py
agiz: Done, but this was implemented the way it was only to share errors with fields.py and since fields.py errors are now hard-coded, I can use the correct way of reporting errors here.

agiz:
Did some work on forms.py, field.py and formfields.py.

piplmesh/account/backends.py
```diff
@@ -44,39 +58,11 @@ def authenticate(self, token=None, request=None):
...
-            user.last_name = fb['last_name']
-            user.save()
-
-            # Create and save account user
-            profile = models.UserProfile.objects.create(user=user, token=access_token, facebook_id=fb['id'], gender=fb['gender'])
-            profile.save()
+        
+        # TODO: Check if id and other fields are returned
+        # TODO: Move user retrieval/creation to User document/manager
+        user, created = self.user_class.objects.get_or_create(facebook_id=fb.get('id'), defaults={'username': fb.get('username'), 'first_name': fb.get('first_name'), 'last_name': fb.get('last_name'), 'email': fb.get('email'), 'gender': fb.get('gender')})
```
mitar: Can you split also defaults into multiple lines?

piplmesh/account/fields.py
```diff
@@ -1,16 +1,77 @@
...
+    error_messages = {
+        'bounds': (u"Value is out of bounds."),
+    }
+
+    def __init__(self, upper_limit=None, lower_limit=None, *args, **kwargs):
+        self.upper_limit = upper_limit
+        self.lower_limit = lower_limit
+
+        if self.upper_limit and not isinstance(self.upper_limit, (datetime.datetime, datetime.date)):
+            self.error(self.error_messages['bounds'])
```
mitar: No? Here should be an error about arguments.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,77 @@
...
+        if self.lower_limit and not isinstance(self.lower_limit, (datetime.datetime, datetime.date)):
+            self.error(self.error_messages['bounds'])
```
mitar: No? Here should be an error about arguments.

piplmesh/account/fields.py
```diff
@@ -1,16 +1,77 @@
...
+
+        super(LimitedDateTimeField, self).__init__(*args, **kwargs)
+   
+    def validate(self, value):
+        super(LimitedDateTimeField, self).validate(value)
+
+        def error():
+            self.error(self.error_messages['bounds'])
```
mitar: Just hard-code it here.

mitar: Those are two different thins, mongoengine fields and Django fields.

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,34 @@
+import datetime
+
+from django import forms
+from django.utils.translation import ugettext_lazy as _
+
+from piplmesh.account import fields
+
+class LimitedDateTimeField(forms.DateTimeField):
+    default_error_messages = {
```
mitar: Not good, you have overridden parent's default messages. You have to add to them (what if `super(LimitedDateTimeField, self).clean(value)` wants to use some message)?

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,34 @@
...
+        'bounds': _(u"Value is out of bounds."),
+        'syntax_lower': _(u"Invalid lower_limit argument."),
+        'syntax_upper': _(u"Invalid upper_limit argument."),
```
mitar: Those two do not got to default messages!

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    def __init__(self, upper_limit=None, lower_limit=None, input_formats=None, *args, **kwargs):
+        self.upper_limit = upper_limit
+        self.lower_limit = lower_limit
+
+        if self.upper_limit and not isinstance(self.upper_limit, (datetime.datetime, datetime.date)):
+            raise forms.ValidationError(self.error_messages['syntax_upper'])
```
mitar: Just hard-code. I would not raise ValidationError here but some other (like AttributeError).

piplmesh/account/formfields.py
mitar: You could of course do that even with shared error messages. For your current code:

```
default_error_messages = fields.error_messages
```

:-)

piplmesh/account/formfields.py
agiz: Fixed.

piplmesh/account/formfields.py
agiz: Removed from `default_error_messages`.

piplmesh/account/formfields.py
agiz: Added `bounds` to `default_error_messages` without overriding.

piplmesh/account/fields.py
agiz: Changed `self.error(self.error_messages['bounds'])` to `self.error(u"Value is out of bounds.")`.

piplmesh/account/fields.py
agiz: Changed `self.error(self.error_messages['bounds'])`to `self.error(u"Invalid upper_limit argument.")`.

piplmesh/account/fields.py
agiz: Changed `self.error(self.error_messages['bounds'])`to `self.error(u"Invalid lower_limit argument.")`.

piplmesh/account/backends.py
agiz: Done.

agiz:
Updated backends.py, fields.py, formfields.py.

piplmesh/account/formfields.py
agiz: I have to say, I try to just blindly follow here. `syntax_upper` and `syntax_lower` errors do not go to `default_error_messages` because these errors do not get displayed to end user?

piplmesh/account/formfields.py
agiz: There are not many examples of `default_error_messages` on the web - the ones I found all override defaults(?).

piplmesh/account/models.py
```diff
@@ -1,21 +1,17 @@
...
+from mongoengine.django import auth
 
-class UserProfile(django_models.Model):
-    """ 
-    Class used for storing additional information about user.
-    """
+from piplmesh.account import fields
 
-    user = django_models.OneToOneField(auth_models.User)
+# rough lower limit date
```
mitar: Start with upper case.

mitar: Or just remove the comment. :-)

piplmesh/account/forms.py
```diff
@@ -31,17 +25,25 @@ class RegistrationForm(auth_forms.UserCreationForm):
...
     
     # Additional information
-    gender = forms.ChoiceField(label=_("Gender"), required=False, choices=GENDER_CHOICES, widget=forms.RadioSelect(renderer=HorizontalRadioRenderer))    
-    current_date = datetime.now()
-    birthdate = forms.DateField(label=_("Birth date"), required=False, widget=widgets.SelectDateWidget(years=[y for y in range(current_date.year, 1900, -1)]))
+    gender = forms.ChoiceField(
+        label=_("Gender"),
+        required=False,
+        choices=fields.GENDER_CHOICES,
+        widget=forms.RadioSelect(renderer=HorizontalRadioRenderer)
```
mitar: Comma at the end.

piplmesh/account/forms.py
```diff
@@ -31,17 +25,25 @@ class RegistrationForm(auth_forms.UserCreationForm):
...
+    )    
+    birthdate = formfields.LimitedDateTimeField(
+        upper_limit=datetime.datetime.today(),
+        lower_limit=datetime.datetime.today() - datetime.timedelta(models.LOWER_DATE_LIMIT),
+        label=_("Birth date"),
+        required=False,
+        widget=widgets.SelectDateWidget(years=[y for y in range(datetime.datetime.today().year,
+        (datetime.datetime.today() - datetime.timedelta(models.LOWER_DATE_LIMIT)).year, -1)])
```
mitar: This is not wrapped in best way. If you want, you can wrap (and indent) arguments to SelectDateWidget again. But not like this.

Also do not forget comma at the end.

piplmesh/account/formfields.py
mitar: Yes.

piplmesh/account/forms.py
agiz:
```
    widget=widgets.SelectDateWidget(
        years=[
            y for y in range(datetime.datetime.today().year,
            (datetime.datetime.today() - datetime.timedelta(models.LOWER_DATE_LIMIT)).year,
            -1),
        ]
    ),
```

This is too much right?

piplmesh/account/formfields.py
```diff
@@ -0,0 +1,30 @@
+import datetime
+
+from django import forms
+from django.utils.translation import ugettext_lazy as _
+
+from piplmesh.account import fields
+
+class LimitedDateTimeField(forms.DateTimeField):
+    forms.DateTimeField.default_error_messages['bounds'] = _(u"Value is out of bounds.")
```
mitar: Not good. You change `default_error_messages` in `forms.DateTimeField` here! Don't do that!

You can try something like:

```
default_error_messages = forms.DateTimeField.default_error_messages.copy()
default_error_messages.update({
    'bounds': _(u"Value is out of bounds."),
})
```

piplmesh/account/forms.py
mitar: No, but not correct. `range` is again a function call, so you should indent it accordingly.

mitar: And comma after the `]`.

piplmesh/account/models.py
agiz: Removed.

piplmesh/account/forms.py
agiz: Added.

piplmesh/account/forms.py
agiz: Corrected indentation.

mitar:
Good. :-)

mitar:
I hope it works! You have tested it, yes? :-)

(We do not yet have unit tests!)

mitar:
(This could be your next pull request.)

piplmesh/account/formfields.py
agiz: Thanks. Did you really think I could come up with that? :) Even now that I see it I vaguely know the princip of `default_error_messages`.

mitar: Of course? You should trust into yourself! :-) The only limit to you is yourself.

If something is not clear about `default_error_messages`, check the code. How it is copied to `self.error_messages`.

agiz:
I have tested it numerous times but just simple things. I tested upper and lower bounds couple of pulls ago, but tested upper limits every pull. I tested registration and login thoroughly when figuring this part out. i just tested valid registration and login and it works.

Issue #70, opened 2012-03-21
Define MongoDB model for posts and comments


mitar:
Define MongoDB model for posts and comments as [MongoEngine](http://mongoengine.org/) document types. We will support link, photo and video post types.

Maybe check [Facebook model definition](https://developers.facebook.com/docs/reference/api/post/) for ideas on what all can be there.

mitar:
Done with #104.

PR #104, opened 2012-04-09
Issue #70, #team2

maticko:
Definition of Post and Comment models - issue #70

### DIFF
--- a/piplmesh/account/models.py
+++ b/piplmesh/account/models.py
@@ -8,7 +8,7 @@ from piplmesh.account import fields
 
 LOWER_DATE_LIMIT = 366 * 120
 
-TEXT_POST_MAX_LENGTH = 500
+MESSAGE_MAX_LENGTH = 500
 COMMENT_MAX_LENGTH = 300
 
 class User(auth.User):
@@ -30,31 +30,31 @@ class Comment(mongoengine.EmbeddedDocument):
 
 class Post(mongoengine.Document):
     """
-    This class defines document type for storing post on our wall.
+    This class defines document type for posts on a wall.
     """
 
     author = mongoengine.ReferenceField('User', required=True)
     created_time = mongoengine.DateTimeField(default=datetime.datetime.now()) 
-    edited_time = mongoengine.DateTimeField()
+    updated_time = mongoengine.DateTimeField()
     comments = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Comment))
 
 class TextPost(Post):
     """
-    This class defines support for posting text in wall posts.
+    This class defines document type for text posts.
     """
 
-    text = mongoengine.StringField(max_length=TEXT_POST_MAX_LENGTH)
+    message = mongoengine.StringField(max_length=MESSAGE_MAX_LENGTH)
 
 class ImagePost(Post):
     """
-    This class defines support for posting images in wall posts.
+    This class defines document type for image posts.
     """
 
     image_path = mongoengine.StringField()
 
 class LinkPost(Post):
     """
-    This class defines support for posting links in wall posts.
+    This class defines document type for link posts.
     """
 
     link_url = mongoengine.URLField()  
--- a/piplmesh/account/models.py
+++ b/piplmesh/account/models.py
@@ -8,8 +8,8 @@ from piplmesh.account import fields
 
 LOWER_DATE_LIMIT = 366 * 120
 
-MESSAGE_MAX_LENGTH = 500
-COMMENT_MAX_LENGTH = 300
+POST_MESSAGE_MAX_LENGTH = 500
+COMMENT_MESSAGE_MAX_LENGTH = 300
 
 class User(auth.User):
     birthdate = fields.LimitedDateTimeField(upper_limit=datetime.datetime.today(), lower_limit=datetime.datetime.today() - datetime.timedelta(LOWER_DATE_LIMIT))
@@ -26,7 +26,7 @@ class Comment(mongoengine.EmbeddedDocument):
 
     created_time = mongoengine.DateTimeField(default=datetime.datetime.now())
     author = mongoengine.ReferenceField(User, required=True)
-    comment = mongoengine.StringField(max_length=COMMENT_MAX_LENGTH)
+    message = mongoengine.StringField(max_length=COMMENT_MESSAGE_MAX_LENGTH)
 
 class Post(mongoengine.Document):
     """
@@ -43,7 +43,7 @@ class TextPost(Post):
     This class defines document type for text posts.
     """
 
-    message = mongoengine.StringField(max_length=MESSAGE_MAX_LENGTH)
+    message = mongoengine.StringField(max_length=POST_MESSAGE_MAX_LENGTH)
 
 class ImagePost(Post):
     """
@@ -58,3 +58,5 @@ class LinkPost(Post):
     """
 
     link_url = mongoengine.URLField()  
+    link_caption = mongoengine.StringField()
+    link_description = mongoengine.StringField()
--- a/piplmesh/account/models.py
+++ b/piplmesh/account/models.py
@@ -23,7 +23,7 @@ class Comment(mongoengine.EmbeddedDocument):
     This class defines document type for comments on wall posts.
     """
 
-    created_time = mongoengine.DateTimeField(default=lambda:datetime.datetime.now())
+    created_time = mongoengine.DateTimeField(default=lambda: datetime.datetime.now())
     author = mongoengine.ReferenceField(User, required=True)
     message = mongoengine.StringField(max_length=COMMENT_MESSAGE_MAX_LENGTH)
 
@@ -33,7 +33,7 @@ class Post(mongoengine.Document):
     """
 
     author = mongoengine.ReferenceField(User, required=True)
-    created_time = mongoengine.DateTimeField(default=lambda:datetime.datetime.now())
+    created_time = mongoengine.DateTimeField(default=lambda: datetime.datetime.now())
     updated_time = mongoengine.DateTimeField()
     comments = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Comment))
 
--- a/piplmesh/account/models.py
+++ b/piplmesh/account/models.py
@@ -24,7 +24,7 @@ class Post(mongoengine.Document):
         This class defines document type for storing post on our wall.
     '''
 
-    author = fields.ReferenceField(User, required=True)
+    author = fields.ReferenceField(User, required=True, reverese_delete_rule=CASCADE)
     created_time = fields.DateTimeField(auto_now_add=True) 
     edited_time = fields.DateTimeField()
 
@@ -54,11 +54,11 @@ class PostComments(Post):
         Adds support for storing comments under each post.
     '''
 
-    comments = fields.ListFields()
+    comments = fields.ListFields(EmbeddedDocumentField(Comment))
 
-class Comment(mongoengine.Document):
+class Comment(mongoengine.EmbeddedDocument):
     '''
-        This class defines document type for comments on wall posts. .
+        This class defines embedded Document type for comments on wall posts. .
     '''
 
     created_time = fields.DateTimeField(auto_now_add=True)
--- a/piplmesh/account/models.py
+++ b/piplmesh/account/models.py
@@ -2,7 +2,6 @@ import datetime
 
 import mongoengine
 from mongoengine.django import auth
-#from mongoengine import Document
 
 from piplmesh.account import fields
 
--- a/piplmesh/account/models.py
+++ b/piplmesh/account/models.py
@@ -30,30 +30,30 @@ class Post(mongoengine.Document):
 
 class TextPost(Post):
     """
-    This class adds support for posting text in wall posts.
+    This class defines support for posting text in wall posts.
     """
 
-    message = fields.StringField(max_langth=TEXT_POST_MAX_LENGTH)
+    text = fields.StringField(max_langth=TEXT_POST_MAX_LENGTH)
 
 class ImagePost(Post):
     """
-    This class adds support for posting images in wall posts. For now we don't support uploading photos.
+    This class defines support for posting images in wall posts.
     """
 
     image_path = fields.StringField()
 
 class LinkPost(Post):
     """
-    This class adds support for posting links in wall posts.
+    This class defines support for posting links in wall posts.
     """
 
     link_url = fields.URLField()  
 
 class Comment(mongoengine.EmbeddedDocument):
     """
-    This class defines embedded Document type for comments on wall posts. .
+    This class defines document type for comments on wall posts.
     """
 
     created_time = fields.DateTimeField(auto_now_add=True)
     author = fields.ReferencedField(User, required=True)
-    text = fields.StringField(max_length = COMMENT_MAX_LENGTH)
+    comment = fields.StringField(max_length=COMMENT_MAX_LENGTH)
--- a/piplmesh/account/models.py
+++ b/piplmesh/account/models.py
@@ -23,16 +23,17 @@ class Post(mongoengine.Document):
     This class defines document type for storing post on our wall.
     """
 
-    author = fields.ReferenceField(User, required=True, reverese_delete_rule=CASCADE)
+    author = mongoengine.ReferenceField('User', required=True)
     created_time = fields.DateTimeField(auto_now_add=True) 
     edited_time = fields.DateTimeField()
+    comments = fields.ListFields(EmbeddedDocumentField(Comment))
 
 class TextPost(Post):
     """
     This class adds support for posting text in wall posts.
     """
 
-    message = fields.StringField(max_langth=MAX_USER_POST_MESSAGE_LENGTH)
+    message = fields.StringField(max_langth=TEXT_POST_MAX_LENGTH)
 
 class ImagePost(Post):
     """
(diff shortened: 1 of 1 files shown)


piplmesh/account/models.py
```diff
@@ -7,10 +7,60 @@
 
 LOWER_DATE_LIMIT = 366 * 120
 
+#Defined maximum post and comment length
```
mitar: I think this comment is useless: constants should be named so that from the name itself it is clear what they do.

Furthermore, we put space after `#`.

piplmesh/account/models.py
```diff
@@ -7,10 +7,60 @@
...
+MAX_USER_POST_MESSAGE_LENGTH = 500
```
mitar: `MAX_USER_POST_MESSAGE_LENGTH` -> `TEXT_POST_MAX_LENGTH` ?

To be consistent.

piplmesh/account/models.py
```diff
@@ -7,10 +7,60 @@
...
     gender = fields.GenderField()
     language = fields.LanguageField()
     
     facebook_id = mongoengine.IntField()
-    facebook_token = mongoengine.StringField(max_length=150)
\ No newline at end of file
+    facebook_token = mongoengine.StringField(max_length=150)
+
+class Post(mongoengine.Document):
+    '''
```
mitar: Use `"""`.

piplmesh/account/models.py
```diff
@@ -7,10 +7,60 @@
...
+        This class defines document type for storing post on our wall.
```
mitar: Don't indent docstring.

piplmesh/account/models.py
maticko: Makes sense so i changed it.

piplmesh/account/models.py
maticko: done

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
maticko: Deleted comment.

piplmesh/account/models.py
```diff
@@ -7,10 +7,59 @@
...
+
+class Post(mongoengine.Document):
+    """
+    This class defines document type for storing post on our wall.
+    """
+
+    author = fields.ReferenceField(User, required=True, reverese_delete_rule=CASCADE)
+    created_time = fields.DateTimeField(auto_now_add=True) 
+    edited_time = fields.DateTimeField()
+
```
mitar: Will we allow posts to be edited?

piplmesh/account/models.py
```diff
@@ -7,10 +7,59 @@
...
+    image_path = fields.StringField()
+
+class LinkPost(Post):
+    """
+    This class adds support for posting links in wall posts.
+    """
+
+    link_url = fields.URLField()
+
+class PostComments(Post):
```
mitar: This is not good. Comments will not be posts. I would put comments as list of embeded documents of `Post` document.

mitar: And all types of posts can have comments.

piplmesh/account/models.py
```diff
@@ -7,10 +7,59 @@
...
\ No newline at end of file
+    facebook_token = mongoengine.StringField(max_length=150)
+
+class Post(mongoengine.Document):
+    """
+    This class defines document type for storing post on our wall.
+    """
+
+    author = fields.ReferenceField(User, required=True, reverese_delete_rule=CASCADE)
+    created_time = fields.DateTimeField(auto_now_add=True)
```
mitar: _created_time_ or _create_time_?

piplmesh/account/models.py
```diff
@@ -7,10 +7,59 @@
...
+class PostComments(Post):
+    """
+    Adds support for storing comments under each post.
+    """
+
+    comments = fields.ListFields(EmbeddedDocumentField(Comment))
+
+class Comment(mongoengine.EmbeddedDocument):
+    """
+    This class defines embedded Document type for comments on wall posts. .
```
mitar: Just _This class defines document type for comments on wall posts._

piplmesh/account/models.py
```diff
@@ -7,10 +7,59 @@
...
+    """
+
+    created_time = fields.DateTimeField(auto_now_add=True)
+    author = fields.ReferencedField(User, required=True)
+    text = fields.StringField(max_length = COMMENT_MAX_LENGTH)
```
mitar: No space around `=`.

mitar: _text_ -> _comment_?

piplmesh/account/models.py
```diff
@@ -7,10 +7,59 @@
...
+    author = fields.ReferenceField(User, required=True, reverese_delete_rule=CASCADE)
+    created_time = fields.DateTimeField(auto_now_add=True) 
+    edited_time = fields.DateTimeField()
+
+class TextPost(Post):
+    """
+    This class adds support for posting text in wall posts.
+    """
+
+    message = fields.StringField(max_langth=MAX_USER_POST_MESSAGE_LENGTH)
```
mitar: _message_ -> _text_?

piplmesh/account/models.py
```diff
@@ -7,10 +7,59 @@
...
+
+class ImagePost(Post):
+    """
+    This class adds support for posting images in wall posts. For now we don't support uploading photos.
```
mitar: The second sentence is unnecessary. Put that into TODO if you want. :-)

piplmesh/account/models.py
```diff
@@ -7,10 +7,59 @@
...
+    """
+
+    image_path = fields.StringField()
+
+class LinkPost(Post):
+    """
+    This class adds support for posting links in wall posts.
```
mitar: _adds support_? Or defines?

piplmesh/account/models.py
maticko: After first look of Post fields which facebook uses they have something similar, but I named it wrong. They have updated_time (The time of the last comment on this post) and that might be useful. Should I delete it or rename?

piplmesh/account/models.py
maticko: created_time  ... because it means "The time the post was initially published"

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
```diff
@@ -7,10 +7,53 @@
...
+
+class Post(mongoengine.Document):
+    """
+    This class defines document type for storing post on our wall.
+    """
+
+    author = mongoengine.ReferenceField('User', required=True)
+    created_time = fields.DateTimeField(auto_now_add=True) 
+    edited_time = fields.DateTimeField()
+    comments = fields.ListFields(EmbeddedDocumentField(Comment))
```
mitar: This works? Because `Comment` is defined bellow?

mitar: Shouldn't be `fields.EmbeddedDocumentField` instead of `EmbeddedDocumentField`?

piplmesh/account/models.py
mitar: Yes, `updated_time` is a good idea.

piplmesh/account/models.py
```diff
@@ -7,10 +7,53 @@
...
+    link_url = fields.URLField()  
+
+class Comment(mongoengine.EmbeddedDocument):
+    """
+    This class defines document type for comments on wall posts.
+    """
+
+    created_time = fields.DateTimeField(auto_now_add=True)
+    author = fields.ReferencedField(User, required=True)
+    comment = fields.StringField(max_length=COMMENT_MAX_LENGTH)
```
mitar: Sorry. Let it be `message`, as it is in Facebook.

piplmesh/account/models.py
```diff
@@ -7,10 +7,53 @@
...
+    created_time = fields.DateTimeField(auto_now_add=True) 
+    edited_time = fields.DateTimeField()
+    comments = fields.ListFields(EmbeddedDocumentField(Comment))
+
+class TextPost(Post):
+    """
+    This class defines support for posting text in wall posts.
+    """
+
+    text = fields.StringField(max_langth=TEXT_POST_MAX_LENGTH)
```
mitar: `message`?

piplmesh/account/models.py
```diff
@@ -7,10 +7,53 @@
...
     language = fields.LanguageField()
     
     facebook_id = mongoengine.IntField()
-    facebook_token = mongoengine.StringField(max_length=150)
\ No newline at end of file
+    facebook_token = mongoengine.StringField(max_length=150)
+
+class Post(mongoengine.Document):
+    """
+    This class defines document type for storing post on our wall.
```
mitar: _This class defines document type for posts on a wall._

piplmesh/account/models.py
```diff
@@ -7,10 +7,53 @@
...
+    """
+
+    author = mongoengine.ReferenceField('User', required=True)
+    created_time = fields.DateTimeField(auto_now_add=True) 
+    edited_time = fields.DateTimeField()
+    comments = fields.ListFields(EmbeddedDocumentField(Comment))
+
+class TextPost(Post):
+    """
+    This class defines support for posting text in wall posts.
```
mitar: _This class defines document type for text posts._

piplmesh/account/models.py
```diff
@@ -7,10 +7,53 @@
...
+class ImagePost(Post):
+    """
+    This class defines support for posting images in wall posts.
+    """
+
+    image_path = fields.StringField()
+
+class LinkPost(Post):
+    """
+    This class defines support for posting links in wall posts.
```
mitar: _This class defines document type for link posts._

piplmesh/account/models.py
```diff
@@ -7,10 +7,53 @@
...
+class TextPost(Post):
+    """
+    This class defines support for posting text in wall posts.
+    """
+
+    text = fields.StringField(max_langth=TEXT_POST_MAX_LENGTH)
+
+class ImagePost(Post):
+    """
+    This class defines support for posting images in wall posts.
```
mitar: _This class defines document type for image posts._

mitar:
Just minor things. Otherwise good.

piplmesh/account/models.py
```diff
@@ -7,10 +7,53 @@
...
+    """
+
+    image_path = fields.StringField()
+
+class LinkPost(Post):
+    """
+    This class defines support for posting links in wall posts.
+    """
+
+    link_url = fields.URLField()
```
mitar: Add also `link_caption`, `link_description`?

maticko: Done.

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
maticko: Fixed.

piplmesh/account/models.py
maticko: changed to _message_ , and i also changed constant name to _MESSAGE_MAX_LENGTH_ to be consistent.

piplmesh/account/models.py
maticko: Here i would keep it as it is... so we can distinguish them from posts.

mitar: The problem is, that you will probably do something like:

```
comment = Comment.objects.get(...)[0]
```

and it is really strange to do then:

```
comment.comment
```

No?

piplmesh/account/models.py
mitar: _POST_MESSAGE_MAX_LENGTH_?

piplmesh/account/models.py
maticko: Fixed. But i used mongoengine.EmbeddedDocumentField.

mitar: Why? It is defined in `fields`, no?

piplmesh/account/models.py
maticko: Ok but then I should change COMMENT_MAX_LENGTH  to COMMENT_MESSAGE_MAX-LENGTH ?

mitar: _COMMENT_MESSAGE_MAX_LENGTH_ you mean? OK, could be.

piplmesh/account/models.py
maticko: comments = fields.ListField(mongoengine.EmbeddedDocumentField(Comment))
AttributeError: 'module' object has no attribute 'ListField

and i gat bunch of errors for other Fields if i use fields.

mitar: OK, because `fields` are not `mongoengine.fields`. :-) Yes, for mongoengine fields use `mongoengine`.

Have you checked if all this really works? `Comment` being referenced before the definition?

piplmesh/account/models.py
```diff
@@ -2,15 +2,61 @@
 
 import mongoengine
 from mongoengine.django import auth
+#from mongoengine import Document
```
mitar: Remove this?

piplmesh/account/models.py
```diff
@@ -2,15 +2,61 @@
...
+    created_time = mongoengine.DateTimeField(default=datetime.datetime.now())
+    author = mongoengine.ReferenceField(User, required=True)
+    message = mongoengine.StringField(max_length=COMMENT_MESSAGE_MAX_LENGTH)
+
+class Post(mongoengine.Document):
+    """
+    This class defines document type for posts on a wall.
+    """
+
+    author = mongoengine.ReferenceField('User', required=True)
```
mitar: Decide, `'User'` or `User`. You use both in `ReferenceField`s.

piplmesh/account/models.py
```diff
@@ -2,15 +2,61 @@
...
-    facebook_token = mongoengine.StringField(max_length=150)
\ No newline at end of file
+    facebook_token = mongoengine.StringField(max_length=150)
+
+class Comment(mongoengine.EmbeddedDocument):
+    """
+    This class defines document type for comments on wall posts.
+    """
+
+    created_time = mongoengine.DateTimeField(default=datetime.datetime.now())
```
mitar: I am not sure if this is OK default. Because it is evaluated on module loading and not when field is used (which can be also multiple times).

maticko: Ok i will check this later.

mitar: TODO.

maticko: this is not working properly... you were right it's time from module loading and not when field is used...
so i will remove this out. We will have to fill this when we are saving post in db...

mitar: This is not true. Read documentation about default values.

mitar: (We do not have to remove it, you just have to fix it.)

maticko: can i do it with custom save fucntion? Otherwise I currently have no idea, how to implement this...

mitar: See those examples:

http://mongoengine.org/docs/v0.4/guide/defining-documents.html#field-arguments

maticko: so just like this: \* created_time = mongoengine.DateTimeField(datetime.datetime.now())\* could be ok?

mitar: Why?

mitar: See examples for default attribute, of course. The one with lambda function. Do you know lambda function?

maticko: Huh I actually never used lambda before so I haven't taught about that. But it's good idea and it works. So i fixed that. What about updated_time?
How do you consider this idea: http://www.b-list.org/weblog/2006/nov/02/django-tips-auto-populated-fields/#c2650

piplmesh/account/models.py
```diff
@@ -7,10 +7,55 @@
...
-    facebook_token = mongoengine.StringField(max_length=150)
\ No newline at end of file
+    facebook_token = mongoengine.StringField(max_length=150)
+
+class Comment(mongoengine.EmbeddedDocument):
+    """
+    This class defines document type for comments on wall posts.
+    """
+
+    created_time = mongoengine.DateTimeField(default=lambda:datetime.datetime.now())
```
mitar: Space after `:`.

piplmesh/account/models.py
```diff
@@ -7,10 +7,55 @@
...
+    author = mongoengine.ReferenceField(User, required=True)
+    message = mongoengine.StringField(max_length=COMMENT_MESSAGE_MAX_LENGTH)
+
+class Post(mongoengine.Document):
+    """
+    This class defines document type for posts on a wall.
+    """
+
+    author = mongoengine.ReferenceField(User, required=True)
+    created_time = mongoengine.DateTimeField(default=lambda:datetime.datetime.now())
```
mitar: Same.

piplmesh/account/models.py
mitar: Leave `updated_time` as it is. We will anyway need code to update it on every change, so it can be also initial change.

mitar: The link above is for relational models. But yes, similar things work. But default is here nicer.

piplmesh/account/models.py
maticko: fixed

piplmesh/account/models.py
maticko: Fixed.

PR #105, opened 2012-04-09
Pull request for Issue#42, Team 1

NejcZupec:
Now when user login, he is redirected back from where he came.


piplmesh/account/templates/login.html
```diff
@@ -12,6 +12,7 @@
                 {% csrf_token %}
                 {{ form.as_p }}
                 <p><button type="submit" class="auth_button" value="value">{% trans "Login" %}</button></p>
+                <input type="hidden" name="next" value="{{ next }}" />
```
mitar, line 15: OK. You are right. login view populates this variable.

NejcZupec, line 15: :)

piplmesh/frontend/templates/header.html
```diff
@@ -36,7 +36,7 @@
                     </form>
                 </li>
             {% else %}
-                <li><a href="{% url login %}">{% trans "Log in" %}</a></li>
+                <li><a href="{% url login %}?next={{ request.get_full_path }}">{% trans "Log in" %}</a></li>
```
mitar, line 39: But you cannot hardcode `next` - you should use the value from `django.contrib.auth.REDIRECT_FIELD_NAME`.

piplmesh/settings.py
```diff
@@ -176,7 +176,7 @@
     }
 }
 
-LOGIN_REDIRECT_URL = '/'
+#LOGIN_REDIRECT_URL = '/'
```
mitar: Remove it or leave it. Not just comment it out (especially not without an explanation).

Probably there is nothing wrong with this left as it is, it is then a fallback URL to redirect.

piplmesh/frontend/templates/header.html
mitar, line 39: TODO.

piplmesh/account/templates/login.html
mitar, line 15: It is not XHTML valid. Also check in current main repository how to get better `next` value (called `next_url`).

mitar:
Update pull request to latest from main repository, please.

piplmesh/settings.py
NejcZupec: Ok, I will leave it.

mitar:
Reopen. :-)

mitar:
And finish your work!

Issue #86, opened 2012-03-25
Improve search


mitar:
Past search query string should be visible on result page. "See your results"  should be visible only on results page. That X on the right is unnecessary. Design the whole search input box as Firefox home search box. We should also have ads on the right side of the results?

mitar:
This is still not finished. ;-) Ads are missing.

PR #106, opened 2012-04-10
Part od Issue #86

domenc:
Remove X from search bar and display "See your results" only on search page

### DIFF
--- a/piplmesh/frontend/static/piplmesh/css/style.css
+++ b/piplmesh/frontend/static/piplmesh/css/style.css
@@ -90,6 +90,10 @@ ul.errorlist {
     display: inline-block;
 }
 
+td.gsc-clear-button {
+    display: none !important;
+}
+
 /* -------------------- MAIN ------------------------ */
 #main {
     width: 960px;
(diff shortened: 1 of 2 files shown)


domenc:
I have few questions:
- Past search query string should be visible on result page: How can I make this? With database history?
- Design the whole search input box as Firefox home search box: Add Image "Google" left od the search box and delete from input box?
- We should also have adds on the right side of the results?: Do you mean ads?

mitar:
> Past search query string should be visible on result page

No, this just means that you prefill the search box from `q` query parameter.

> Add Image "Google" left od the search box and delete from input box?

Yes, add "Google". What delete?

> Do you mean ads?

Yes. :-)

mitar:
BTW, this pull request was pulled in.

domenc:
OK, I will add Image and add query on result page, but no ads (this is bad).

mitar:
Why are ads bad? Would you rather pay with money for your services on the Internet? The payment on the Internet is: ads or privacy. Because you pay with that, you do not have to pay with money. You can always pay with money Gmail and do not get ads, you know that? Are you doing that? No.

Would it be possible to have this optional? Like a user value which would enable or disable this for given user? And then we could have like users who would get awarded and not be given ads?

PR #111, opened 2012-04-11
Issue #70 - fix

maticko:
Fixed problem with issue #70 because models were defined in wrong place.

### DIFF
--- a/piplmesh/api/models.py
+++ b/piplmesh/api/models.py
@@ -2,7 +2,7 @@ import datetime
 
 import mongoengine
 
-from piplmesh.models import User
+from piplmesh.account import models
 
 POST_MESSAGE_MAX_LENGTH = 500
 COMMENT_MESSAGE_MAX_LENGTH = 300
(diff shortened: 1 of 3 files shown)


piplmesh/api/models.py
```diff
@@ -1,3 +1,50 @@
-from django.db import models
+import datetime
 
-# Create your models here.
+import mongoengine
+
+from piplmesh.models import User
```
mitar: This is not correct. Wrong import. And import only modules!

maticko: Fixed?

mitar:
You should add `piplmesh.api` to `INSTALLED_APPS` and you will see if it works. :-)

maticko:
looks ok to me.

piplmesh/settings.py
```diff
@@ -145,11 +145,12 @@
 
 INSTALLED_APPS = (
     # Ours are first so that we can override default templates in other apps
-    'piplmesh.frontend',
     'piplmesh.account',
-    
+	'piplmesh.api',
```
mitar: Indentation!

maticko: Fixed.

piplmesh/settings.py
```diff
@@ -145,11 +145,12 @@
 
 INSTALLED_APPS = (
     # Ours are first so that we can override default templates in other apps
-    'piplmesh.frontend',
     'piplmesh.account',
-    
+    'piplmesh.api',
+    'piplmesh.frontend',
+
+	'django.contrib.messages',
```
mitar: Indentation! Still!

maticko:
This is fixed now....

PR #115, opened 2012-04-11
Add past query

domenc:
Add string: "Query result for: [STRING]" on search page

### DIFF
--- a/piplmesh/frontend/templates/plain.html
+++ b/piplmesh/frontend/templates/plain.html
@@ -1,8 +1,13 @@
 {% extends "base.html" %}
 
+{% load i18n %}
+
 {% block body %}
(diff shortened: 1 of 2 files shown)


piplmesh/frontend/templates/plain.html
```diff
@@ -1,8 +1,13 @@
 {% extends "base.html" %}
 
+{% load i18n %}
+
 {% block body %}
 {% include "header.html" %}
```
mitar, line 6: No, this should not into plain template. Make your own template for search results and extend plain there, adding what you need.

mitar:
So this pull request is obsolete? Why you opened a new pull request?

I am closing it now.

Issue #63, opened 2012-03-21
Create/update template tag for user avatar


mitar:
Create/update template tag for user avatar so that it first tries to get avatar from the user object (like from Facebook or OpenID) and if this fails it tries Gravatar (template tag for the later has already been written, so it should be reused/extended).

PR #116, opened 2012-04-11
Pull request for #63 #team1

matevzmihalic:
#63: Display correct user avatar


mitar:
Aren't you team 1?

matevzmihalic:
Oh, I keep making mistakes today. I am team 1.

mitar:
Thank you for making me stay attentive.

piplmesh/frontend/templatetags/gravatar.py
```diff
@@ -6,7 +6,7 @@
 
 register = template.Library()
 
-@register.inclusion_tag('gravatar.html', takes_context=True)
+@register.simple_tag(takes_context=True)
```
mitar: No, it is a very good practice to use external template. Really. This makes things more modular and easier to customize. You really do not want HTML in Python code. Python is done by programmers, HTML is done by designers. This is why we like MVC paradigm. It makes easier to split tasks. And also you do not need to give designers access to the code. You can be assured that they can break only design. :-)

piplmesh/frontend/templatetags/gravatar.py
```diff
@@ -38,9 +38,4 @@ def gravatar(context, user, size=50, default_avatar='unknown.png'):
         'default_avatar_url': urllib.quote(default_avatar_url),
     }
 
-    return {
-        'gravatar': {
-            'url': gravatar_url,
-            'size': size,
-        }
-    }
+    return '<img src="%s" width="%d" height="%d" class="gravatar" />' % (gravatar_url, size, size)
```
mitar: So no, this is ugly.

piplmesh/account/templatetags/user.py
```diff
@@ -0,0 +1,21 @@
+from django import template
+
+from piplmesh.account.templatetags import facebook
+from piplmesh.frontend.templatetags import gravatar
```
mitar: Why is something in account and something in frontend? This smells wrong. :-) Probably we could move all this to account, no?

piplmesh/account/templatetags/facebook.py
```diff
@@ -27,8 +27,9 @@ def facebook_picture(self, size):
     variable height) and "large" (200 pixels wide, variable height). 
     """
     
-    results = '%s?type=%s' % (utils.graph_api_url('%s/picture' % self), size)
-    return results
+    image_url = '%s?type=%s' % (utils.graph_api_url('%s/picture' % self), size)
+    
+    return '<img src="%s" class="facebook" />' % image_url
```
mitar: Ugly. No HTML in Python code, please.

piplmesh/account/templatetags/user.py
```diff
@@ -0,0 +1,21 @@
...
+def profile_image(context, user, size='square'):
+    """
+    Returns img tag with users avatar.
+    
+    Takes ``User`` object and optional size.
+    Sizes include "square" (50x50), "small" (50 pixels wide, variable height), 
+    "normal" (100 pixels wide, variable height) and "large" (200 pixels wide, variable height).
+    """
+    
+    if user.facebook_id:
```
mitar: I would organize this differently. I would put logic into User class. So User class can have some method which returns necessary data for avatar (like URL to the image, size). And then a generic template tag (inclusion tag) renders this into HTML.

mitar: Or a method which returns URL based on the given size. Or something.

So please make only a method to User class and one template tag for displaying that. No need to keep things separate.

mitar: Also for sizes, do Facebook has all sizes you are allowing above?

matevzmihalic: Facebook has all of the above sizes.

matevzmihalic: I've refactored the code as you wished.

mitar: Yes, but we need only squares. Gravatar has only squares, no? So common denominator is only 50x50, no? Aren't there also smaller? For chat/list of users? So which square avatars do Facebook support? Gravatar supports all?

piplmesh/account/models.py
```diff
@@ -1,9 +1,12 @@
-import datetime
+import datetime, hashlib, urllib
+
+from django.conf import settings
+from django.contrib.sites import models as sites_models
```
mitar, line 4: We are not using sites anymore - it requires relational database. But OK, because we have a fallback, let keep using it.

piplmesh/account/utils.py
```diff
@@ -1,15 +1,11 @@
...
 def graph_api_url(self, user=None, page=None, token=None):
     """ 
     Format Facebook Graph API URL. 
     """
     
     param = ''
-    if user:
-        profile = models.UserProfile.objects.get(user=user)
     if user and token:
         param = '?access_token=%s' % profile.token
```
mitar, line 10: `profile` is never define here?

piplmesh/account/models.py
```diff
@@ -17,6 +20,35 @@ class User(auth.User):
     
     facebook_id = mongoengine.IntField()
     facebook_token = mongoengine.StringField(max_length=150)
+    
+    def profile_image_url(self, size='square', request=None, is_secure=False, default_url=''):
+        """
+        Returns users profile image url.
```
mitar, line 26: _users_ -> _user's_, _url_ -> _URL_

piplmesh/account/models.py
```diff
@@ -17,6 +20,35 @@ class User(auth.User):
     
     facebook_id = mongoengine.IntField()
     facebook_token = mongoengine.StringField(max_length=150)
+    
+    def profile_image_url(self, size='square', request=None, is_secure=False, default_url=''):
```
mitar, line 24: I would define API as:

```
from django.contrib.staticfiles.storage import staticfiles_storage

DEFAULT_IMAGE_PATH = 'piplmesh/images/default.png'

def image_url(self, size='medium', default_url=staticfiles_storage.url(DEFAULT_AVATAR_PATH)):
```

We do not really need `request` anymore, you just always do secure requests to Facebook or Gravatar.

piplmesh/account/models.py
mitar, line 26: _Returns user's image URL._

piplmesh/account/models.py
```diff
@@ -17,6 +20,35 @@ class User(auth.User):
...
+            if sites_models.Site._meta.installed:
+                domain = sites_models.Site.objects.get_current().domain
+            else:
+                domain = sites_models.RequestSite(request).domain
+                
+            is_secure = request.is_secure()
+            default_url = '%s/%s/piplmesh/images/unknown.png' % (domain, settings.STATIC_URL)
+        
+        if self.facebook_id:
+            return '%s?type=%s' % (utils.graph_api_url('%s/picture' % self), size)
```
mitar, line 44: `% self`??? This depends on how we define `__unicode__`? This is really broken.

mitar, line 44: And also API for `graph_api_url` is strange - it requires called to know how to construct URLs for the API. I would add additional function into `utils` for user picture and call that. So that Facebook Graph API is defined all in one place and callers just give user object or username or something. For type I would always request square for Facebook.

piplmesh/account/models.py
```diff
@@ -17,6 +20,35 @@ class User(auth.User):
...
+        else:
+            return 'http%(secure)s://secure.gravatar.com/avatar/%(email_hash)s?s=%(size)s&d=http%(secure)s%(default_url)s' % {
+                'secure': 's' if is_secure else '',
+                'email_hash': hashlib.md5(self.email.lower()).hexdigest(),
+                'size': (50, 50, 100, 200)[('square', 'small', 'normal', 'large').index(size)],
+                'default_url': urllib.quote('://' + default_url),
```
mitar, line 50: I would just require that default is full URL.

piplmesh/account/templatetags/user.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Returns img tag with users profile image.
+    
+    Takes ``User`` object and optional size.
+    Sizes include "square" (50x50), "small" (50 pixels wide, variable height), 
+    "normal" (100 pixels wide, variable height) and "large" (200 pixels wide, variable height).
+    """
+    
+    return {
+        'profile_image_url': user.profile_image_url(size, context['request'])
```
mitar, line 16: Pass and use also size in the template.

piplmesh/account/templatetags/user.py
```diff
@@ -0,0 +1,17 @@
...
+register = template.Library()
+
+@register.inclusion_tag('tags/profile_image.html', takes_context=True)
+def profile_image(context, user, size='square'):
+    """
+    Returns img tag with users profile image.
+    
+    Takes ``User`` object and optional size.
+    Sizes include "square" (50x50), "small" (50 pixels wide, variable height), 
+    "normal" (100 pixels wide, variable height) and "large" (200 pixels wide, variable height).
```
mitar, line 12: Document example usage, in same way how it was for gravatar tag.

piplmesh/account/templatetags/user.py
```diff
@@ -0,0 +1,17 @@
+from django import template
+
+register = template.Library()
+
+@register.inclusion_tag('tags/profile_image.html', takes_context=True)
+def profile_image(context, user, size='square'):
```
mitar, line 6: _profile_image_ -> _user_image_

I would for size allow: _small_ (25x25), _medium_ (50x50) or number (of pixles). And then this size would be used in template to resize the image, whatever the `user.image_url` returns (but we pass it also there so that it can potentially optimize the URL based on that).

piplmesh/account/templatetags/user.py
```diff
@@ -0,0 +1,17 @@
```
mitar, line 0: I would name this file _account_, because _user_ is already in the template often user object. So you have some naming confusion.

piplmesh/account/models.py
```diff
@@ -17,6 +20,35 @@ class User(auth.User):
...
+            is_secure = request.is_secure()
+            default_url = '%s/%s/piplmesh/images/unknown.png' % (domain, settings.STATIC_URL)
+        
+        if self.facebook_id:
+            return '%s?type=%s' % (utils.graph_api_url('%s/picture' % self), size)
+        else:
+            return 'http%(secure)s://secure.gravatar.com/avatar/%(email_hash)s?s=%(size)s&d=http%(secure)s%(default_url)s' % {
+                'secure': 's' if is_secure else '',
+                'email_hash': hashlib.md5(self.email.lower()).hexdigest(),
+                'size': (50, 50, 100, 200)[('square', 'small', 'normal', 'large').index(size)],
```
mitar, line 49: It is a good question if we really want to use size argument. Because if we request just 50x50 always, then we will be able to cache that and then resize it to smaller in browser. Or we could always request exactly the requested size, but it is a question if this is reall better.

mitar:

mitar:
You have to update from main repository.

mitar:
You have to update from main repository.

piplmesh/account/templates/tags/profile_image.html
```diff
@@ -0,0 +1 @@
+<img src="{{ profile_image_url }}" class="profile_image" />
```
mitar, line 1: I would pass size to the template and set it as a CSS class on the image. So `small` would become `image_small` CSS class, `(50, 50)` would become `image_50_50` CSS class.

Issue #112, opened 2012-04-11
Keep documentation updated


maticko:
Keep all documentation up-to-date!

mitar:
This ticket can be done multiple times by different teams.

PR #118, opened 2012-04-12
Issue 112- updated doc files, #team2

maticko:
Update of instlation.rst  - issue #112


maticko:
Ehz this is not good...

Issue #120, opened 2012-04-12
Update project layout for Django 1.4


mitar:
In Django 1.4 they changed default [project layout](https://docs.djangoproject.com/en/dev/releases/1.4/#updated-default-project-layout-and-manage-py). Our project should be updated accordingly.

mitar:
Done with #121.

PR #121, opened 2012-04-13
Pull request for issue #120  #team2

ghost:
#120: Layout changed to default Django 1.4 layout.


mitar:
Hm. Wouldn't it be maybe better to simply move `manage.py` file one directory up?

mitar:
Often you have scripts in the top dir (like scripts to run tests, so now also to run the site).

ghost:
I don't know, I think that it is better if the website is separated from documentation?

mitar:
But website (code) is. Only scripts to run things can be in top-level. We will also have other such scripts later on there, you will see. This is also why they have changed the layout in this way. To remove this script from the code, so that one is code, the other is script to run it.

ghost:
Corrected.

Issue #65, opened 2012-03-21
List of currently logged in users


mitar:
On home page provide a list of currently logged in users. Maintain the list up-to-date also after the page has been opened. Do this using HTTP push for real-time updates.

For now use `onbeforeunload` JavaScript event handler to signal the server that user is closing the page (but be careful if the page is opened multiple times).

Use [django-pushserver](https://github.com/mitar/django-pushserver) app for HTTP push support.

agiz:
To start this topic... If one should use `onbeforeunload` as a signal for user parting, what shall be used for user joining? We could just assume users join whenever they connect to the site but there is an issue of having ghost users, that is, the users without javascript. So is it ok to send asynchronous request when user connects as a signal that she joined and has a browser with javascript support?

mitar:
Hmm. A very good point. I have not thought about that.

So we would really need to get information from nginx when user connects and when user disconnects. I will extend django-pushserver with this.

agiz:
Well, you can extend django-pushserver but this will make it less similar to nginx environment. As I mentioned in email correspondence I think there is no signal currently in nginx that detects connections terminated by client. 

For starters try connecting to nginx push server with Chrome (I am using Chromium) and abort the connections and you will see that nothing happens because it terminates silently. There is more :). You can still have an open socket and not even be on the website (ie if you close the site but have an image open from this site in another window).

I suggest we go back to pinging and timeout.

mitar:
In the worst case we will run django-pushserver in the production. :-)

mitar:
Test it out and give feedback. :-)

mitar:
I would recommend that you extend User document with two fields:
- list of open channels with subscription time
- time of last unsubscription

So if the first has any elements, then user is logged in. If it is empty and the second field is more then X seconds in the past, the user is logged out.

agiz:
What can be used to fire delayed events? I want to use something already implemented in django and not some external package. Also this delayed event must have an option to be canceled. Something similar to javascript's `setTimeout`.

mitar:
You cannot. Things are processed always inside of the request handler. This is synchronous operation. And you want that it is very quick (user waits longer for delay). Django signals are also synchronous signals. All this is by design. So the idea is that you just very quickly store stuff and then process things in the background. This is what we will be doing in the future with Celery. So probably then we will have some regular task which will check every 30 seconds for example status of users and send updates if things change.

mitar:
[...]

PR #122, opened 2012-04-14
Issue 65 rev 2 (team 4)

agiz:
Not up to date with repository
I was not able to do this http://grokbase.com/t/gg/mongodb-user/121jyejc7j/how-can-i-pull-one-array-element-not-removes-all-occurrences#20120118zhqusu6k7cdsqo23vpymu4em7y in mongoengine syntax.


agiz:
There are some `prints` and comments in code and some variable names are bad because this is not yet final release.

mitar:
You can do `unset` with:

```
models.User.objects(id=request.user.id, clist__http_if_none_match='...', clist__http_if_modified_since='...').update_one(unset__clist__S=1)
```

mitar:
Please define an embedded document for values you are storing into list. One with `http_if_none_match` and `http_if_modified_since` fields. And use that.

piplmesh/account/views.py
```diff
@@ -75,4 +79,49 @@ def logout(request):
         url = request.POST.get(auth.REDIRECT_FIELD_NAME)
         return auth_views.logout_then_login(request, url)
     else:
-        raise exceptions.PermissionDenied
\ No newline at end of file
+        raise exceptions.PermissionDenied
+
+class PushView(generic_views.TemplateView):
```
mitar: Why are you reimplementing this and not using signals provided by django-pushserver?

piplmesh/frontend/templates/base.html
```diff
@@ -8,6 +8,63 @@
...
         <link rel="stylesheet" href="https://www.google.com/cse/style/look/default.css" type="text/css" />
         <script src="{{ STATIC_URL }}piplmesh/jquery/jquery.js" type="text/javascript"></script>
+        {% load pushserver %}
+        <script type="text/javascript" src="{{ STATIC_URL }}pushserver/updates.js"></script>
+            
+        <script type="text/javascript">
+            var updates_url = '{% filter escapejs %}{% updates_url "a" %}{% endfilter %}';
+            
+            
+            $(document).ajaxSend(function(event, xhr, settings) {
```
mitar: This is probably not needed? Why have you added it? Are we doing POSTs in JavaScript anywhere now?

piplmesh/frontend/templates/plain.html
```diff
@@ -5,5 +5,6 @@
     <div id="content">
         {% block content %}{% endblock %}
     </div>
+    {% include "userlist.html" %}
```
mitar: Just move the content to home. Instead of that juicy content.

piplmesh/frontend/templates/base.html
agiz: Relict from the past or future, but yes, not now :).

piplmesh/account/views.py
agiz: I don't know how :/

piplmesh/frontend/templates/base.html
mitar: We will probably not even need it, because we will be doing REST requests. And those will be authenticated with OAuth one day. :-) (And I was lead to believe that then you do not have CSRF problems. I am not sure about that though.)

piplmesh/account/views.py
mitar: It is written in documentation of [django-pushserver](http://readthedocs.org/docs/django-pushserver/en/latest/usage.html):

```
from django import dispatch

from pushserver import signals

@dispatch.receiver(signals.channel_subscribe)
def process_channel_subscribe(sender, request, channel_id, **kwargs):
    print "Subscribed", request.user, channel_id

@dispatch.receiver(signals.channel_unsubscribe)
def process_channel_unsubscribe(sender, request, channel_id, **kwargs):
    print "Unsubscribed", request.user, channel_id
```

mitar: (Add that to `frontend.views` somewhere.)

mitar: Or probably better: signal handlers go to `account.models` (because they modify models). But Celery task goes to `frontend`, because it pushes things to the client.

agiz: I still don't understand this and believe me I have read this many times :)

I have put above in frontend.views and removed `url(r'^passthrough', account_views.PushView.as_view(), name='push')` from urls.py and I am getting `[14/Apr/2012 12:58:04] "POST /passthrough HTTP/1.1" 404 2830` now.

mitar: Have you not read installation documentation for push server? Or have you not listened at class? I showed this and explained.

agiz:
Moving on to Documents...
When user subscribes I create new document with `create()` and this method allows duplicate entries, yes?

mitar:
It depends how you define uniqueness on fields. :-) If none are unique, then multiple same documents can be stored.

But you are defining embedded document, no?

agiz:
I created new Document in models.py.

```
class UserTracking(mongoengine.Document):
    user = mongoengine.ReferenceField(User)
    http_if_none_match = mongoengine.IntField()
    http_if_modified_since = mongoengine.DateTimeField()
```

piplmesh/account/views.py
agiz: If I want to raise an error on exception in views.py - what is the way to do it?
I am doing it like this `return debug.technical_500_response(request, *sys.exc_info())`.

mitar:
Again, embedded document. Please see how we have defined comments for wall posts.

piplmesh/account/views.py
mitar: Views or signals? In views you just raise exception and with `DEBUG` set to `True` it shows you technical response.

In signals exceptions are not propagated (I configured like that in case of passthrough view).

agiz: In signals, because I test if `HTTP_IF_NONE_MATCH` is integer.

mitar: Why you test that? Just try to convert, if you fail, that request will be ignored.

agiz:
Like this?

```
class User(auth.User):
    birthdate = fields.LimitedDateTimeField(upper_limit=datetime.datetime.today(), lower_limit=datetime.datetime.today() - datetime.timedelta(LOWER_DATE_LIMIT))
    gender = fields.GenderField()
    usertracking = mongoengine.ListField(mongoengine.EmbeddedDocumentField(UserTracking))

    facebook_token = mongoengine.StringField(max_length=150)

class UserTracking(mongoengine.EmbeddedDocument):
    http_if_none_match = mongoengine.IntField()
    http_if_modified_since = mongoengine.DateTimeField()
```

piplmesh/account/views.py
agiz:
```
try:
    http_if_none_match = int(request.META['HTTP_IF_NONE_MATCH'])
except ValueError:
    #return debug.technical_500_response(request, *sys.exc_info())
    #what goes here?
```

agiz:
Help please.

I am adding like this that seems to be ok.
`request.user.update(push__usertracking3={'http_if_none_match':http_if_none_match, 'http_if_modified_since':request.META['HTTP_IF_MODIFIED_SINCE']})`

But I do not know how to unset one `usertracking3` value. 
This `models.User.objects(id=request.user.id, usertracking3__http_if_none_match=request.META['HTTP_IF_NONE_MATCH'], usertracking3__http_if_modified_since=request.META['HTTP_IF_MODIFIED_SINCE']).update_one(unset__usertracking3__S=1)` is not working - this does not even return user(s).

`models.User.objects(id=request.user.id, usertracking3={'http_if_none_match':request.META['HTTP_IF_NONE_MATCH'], 'http_if_modified_since':request.META['HTTP_IF_MODIFIED_SINCE']}).update_one(unset__usertracking3__S=1)` this is not working either - this messes up User document somehow.

Almost everything I do to pull an object from `usertracking` messes User document and is borderline annoying.

Mongoengine documentation is non-existant or high abstract at best usually just for one use case.

mitar:
I would rename _UserTracking_ to _Connection_, field _opened_connections_. Put it at the end of fields, after Facebook fields, in a new group. You also probably need to add channel ID field (currently we have fixed channels, but user could connect to multiple).

Otherwise yes, but probably you have to first define embedded document so that you can reference it later in the code. So the order of definitions is important.

mitar:
What are you reading for mongoengine documentation? [API reference](http://readthedocs.org/docs/mongoengine-odm/en/latest/apireference.html)? [User guide](http://readthedocs.org/docs/mongoengine-odm/en/latest/guide/index.html)?

mitar:
Also tests are good source of examples!

mitar:
BTW, why do you want to check for type of the values? I think you could store both values as strings and this is it. Otherwise, check also the date value. But I really do not think this is necessary, whatever the client sends, it sends.

mitar:
Please use `StringField` for both your fields and the first unset query you used above should work. Also, you have to remove all `None` values immediately from the list after that, otherwise MongoEngine is unable to query over objects (it fails with `AttributeError: 'NoneType' object has no attribute 'get'`).

mitar:
You do not have to run `save` after `request.user.update`. Check the code, it is the same as `models.User.objects(id=request.user.id).update_one`. So it runs directly on the database, not on the object. (You have to use [reload](http://mongoengine.org/docs/v0.4/apireference.html#mongoengine.Document.reload) to refresh the object if you need those changes there.)

agiz:
Thanks for suggestions. I merged open_connections to `str(request.META['HTTP_IF_NONE_MATCH'])+'_'+str(request.META['HTTP_IF_MODIFIED_SINCE'])`since it is both that is needed for identifying connections and thus having two fields is redundant.

I am facing another problem - I don't know how to make unique entry with mongoengine. I have documents defined as this:

```
class Channel(mongoengine.EmbeddedDocument):
    channel_id = mongoengine.StringField(primary_key=True)

class Connection(mongoengine.EmbeddedDocument):
    connection_id = mongoengine.StringField()

class User(auth.User):
    birthdate = fields.LimitedDateTimeField(upper_limit=datetime.datetime.today(), lower_limit=datetime.datetime.today() - datetime.timedelta(LOWER_DATE_LIMIT))
    gender = fields.GenderField()

    facebook_token = mongoengine.StringField(max_length=150)

    opened_connections = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Connection))
    channel = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Channel))
    last_access = mongoengine.DateTimeField()
```

I am updating like this:

```
request.user.update(
    push__opened_connections = {
        'connection_id': str(request.META['HTTP_IF_NONE_MATCH'])+'_'+str(request.META['HTTP_IF_MODIFIED_SINCE'])
    },
    set__last_access = datetime.datetime.now(),
    push__channel = {
        'channel_id': channel_id,
    }
)
```

`channel_id` is 'a', yet channel list keeps growing... I have tried many different arguments and different fields such as `ObjectIdField` to no avail.

mitar:
Please, use both fields. This makes things more readable. There is no need to "mangle" with data, you do not gain anything. Just lose readability.

mitar:
Please also do not try to solve all problems with this pull request. For now, we are trying just to find if user is online or offline. You do not need `Channel` document for that. You need only a `StringField` where storing channel ids. Please do it like that.

mitar:
Anyway, you are doing things your own way and I will let you have problems. (There are conceptually wrong - why have two lists, why have a list of channels?) Once you learn from your mistakes, come back.

agiz:
Ok, but `DictField` is better choice to store channels, agree?

mitar:
Why? You just have to store channel ID value for each `Connection` document you store? Why would you use channels for that?

agiz:
Because I think maintaining connection is different than subscribing to channels. Sure there is one default channel open but joining and leaving channels should take an active gesture.

agiz:
Oh, and I was thinking when suggested using `DictField` of having it in User document.

mitar:
And are we doing subscription to channels now or are we doing something else now?

agiz:
Something in between :).

mitar:
We are doing exactly what ticket #65 wants. Please stick to that.

agiz:
Ok, but on totally unrelated note... Can and if so how can I update `DictField` (set__channel) with one update?

```
request.user.update(
    push__opened_connections = {
        'connection_id': str(request.META['HTTP_IF_NONE_MATCH'])+'_'+str(request.META['HTTP_IF_MODIFIED_SINCE'])
    },
    set__last_access = datetime.datetime.now(),
    # TODO: This is not good(?) overrides 
    set__channel = {channel_id: True}
)
```

I could not figure that out. Please do tell, even if it seem irrelevant :)

mitar:
Do you read documentation?

http://www.mongodb.org/display/DOCS/Updating#Updating-%24set

You probably want:

http://www.mongodb.org/display/DOCS/Updating#Updating-%24addToSetand%24each

agiz:
Of course I do. I don't think this is it. I want to have channel_id stored as a key. Something I can do for example with `request.user.channel[channel_id] = '...'` but I want to be able to do it with an update.

mitar:
Read documentation again and again until you figure it out.

mitar:
What are you doing is `request.user.channel = {channel_id: True}`.

agiz:
I know, I know. I am overriding dict(), not good.

piplmesh/frontend/templates/plain.html
agiz: Added.

agiz:
Implemented recommended changes. 

I read documentation over and over and I am still clueless.

mitar:
Clueless about what?

piplmesh/account/models.py
```diff
@@ -7,10 +7,18 @@
...
     language = fields.LanguageField()
-    
+
     facebook_id = mongoengine.IntField()
-    facebook_token = mongoengine.StringField(max_length=150)
\ No newline at end of file
+    facebook_token = mongoengine.StringField(max_length=150)
+
+    opened_connections = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Connection))
+    channel = mongoengine.DictField()
```
mitar: We do not need this here.

piplmesh/account/models.py
```diff
@@ -7,10 +7,18 @@
 
 LOWER_DATE_LIMIT = 366 * 120
 
+class Connection(mongoengine.EmbeddedDocument):
+    http_if_none_match = mongoengine.StringField()
+    http_if_modified_since = mongoengine.StringField()
```
mitar, line 12: Add `channel_id` field.

agiz, line 12: But why here? It's no good.

agiz:
Clueless about https://github.com/wlanslovenija/PiplMesh/pull/122#issuecomment-5137738. 
You suggested: https://github.com/wlanslovenija/PiplMesh/pull/122#issuecomment-5137814

piplmesh/account/models.py
mitar, line 12: Exactly here is good. Those three values tell us which connection we are speaking about. (The cookie is handled automatically by Django.)

mitar:
You do not need that.

piplmesh/account/models.py
agiz, line 12: But channel is irrelevant here. This is just tracking if user is connected to the site. You do not need and much less want to track every channel so intensively. If you still insist, I will of course put channel here.

agiz:
Well I want to learn that anyway.

piplmesh/account/models.py
mitar, line 12: It is not irrelevant. It could happen that two different requests of two different channels have same `http_if_none_match` and `http_if_modified_since`. But they are completely unrelated. And so when you would be removing one from the list, you could remove too much.

mitar:
Then read documentation (more precisely: where in documentation it is written that your code should operate in they way you would like to? I do not see that, so why do you think it should?). But pull request is not a place for that. :-) Use forum for that.

piplmesh/account/models.py
agiz, line 12: And it does not matter. The way I see it, this is just for deducting if user is connected to the site. Channels do not matter here. I would track channels separately and only subscribe/unsubscribe user if she actively decided to do so.

mitar, line 12: Yes, and you could deduce wrong if you remove too much from the list of opened connections. You remove from the list something for other channel, while user is still connected there. You think she is offline, while the other channel is still open.

agiz, line 12: No :), because _other_ channel subscriptions are kept somewhere else.

mitar, line 12: Which other? I am talking about connections here. You can have multiple connections to the server at the same time. Each for one channel. So you have to keep track of them.

Look, it is very simple: if you know how to implement it your way, do it. If you do not, do it as I say.

agiz, line 12: It's hard to explain in just few lines. There is one _default_ connection (this is the one we are monitoring heavily) - think of it as status analogy of IRC. To join other channels some command must be send beforehand declaring "I want to join #foo". The same with unsubscribing "I want to leave #foo".  Just comment on that and I will do what you say next.

mitar, line 12: Why would there be one default connection? What happens in your scenario if user wants to leave default connection but keep others?

BTW, we are not doing IRC clone. All connections are equal.

mitar, line 12: That are two completely separate things. One is which tracking connections are opened from the browser. And the other is which walls user follows in real-time. Those are two separate things. And we are currently not yet doing the second one.

agiz, line 12: Where do you see the problem. If user closes default connection it does not matter. All connections are still recorded in Connection document, I just would not use these for channel subscription tracking. I see we won't go far here so I will implement it your way.

mitar, line 12: Yes, but connections which are recorded in Connection document could be from different channels and there is very small but existent chance that you would unsubscribe too much (also Connection documents from other channel).

agiz, line 12: Added.

agiz:
Added channel_id to Channel document. Removed ajax notifications.

piplmesh/account/models.py
```diff
@@ -7,10 +7,18 @@
...
     language = fields.LanguageField()
-    
+
     facebook_id = mongoengine.IntField()
-    facebook_token = mongoengine.StringField(max_length=150)
\ No newline at end of file
+    facebook_token = mongoengine.StringField(max_length=150)
+
+    opened_connections = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Connection))
+    last_access = mongoengine.DateTimeField()
```
mitar: I would store last unsubscribe time, no?

piplmesh/account/models.py
```diff
@@ -7,10 +7,18 @@
...
     gender = fields.GenderField()
     language = fields.LanguageField()
-    
+
     facebook_id = mongoengine.IntField()
-    facebook_token = mongoengine.StringField(max_length=150)
\ No newline at end of file
+    facebook_token = mongoengine.StringField(max_length=150)
+
+    opened_connections = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Connection))
```
mitar: _opened_connections_ -> _connections_

piplmesh/account/models.py
agiz: Umm... So just rename the variable or?

piplmesh/account/views.py
```diff
@@ -5,9 +5,13 @@
 from django.contrib import auth
 from django.contrib.auth import views as auth_views
 from django.core import exceptions, urlresolvers
+from django.utils.decorators import method_decorator
```
mitar: Import only modules.

piplmesh/account/views.py
```diff
@@ -5,9 +5,13 @@
...
 from django.views import generic as generic_views
+from django.views.decorators import csrf
```
mitar: Import only modules.

piplmesh/frontend/templates/base.html
```diff
@@ -8,6 +8,24 @@
         <link rel="stylesheet" href="{{ STATIC_URL }}piplmesh/css/style.css" type="text/css" media="screen" />
         <link rel="stylesheet" href="https://www.google.com/cse/style/look/default.css" type="text/css" />
         <script src="{{ STATIC_URL }}piplmesh/jquery/jquery.js" type="text/javascript"></script>
+        {% load pushserver %}
+        <script type="text/javascript" src="{{ STATIC_URL }}pushserver/updates.js"></script>
+            
+        <script type="text/javascript">
+            var updates_url = '{% filter escapejs %}{% updates_url "a" %}{% endfilter %}';
+
+            function updateAnswer(data) {
```
mitar: Move this function and registration bellow to external file.

piplmesh/frontend/templates/base.html
```diff
@@ -8,6 +8,24 @@
         <link rel="stylesheet" href="{{ STATIC_URL }}piplmesh/css/style.css" type="text/css" media="screen" />
         <link rel="stylesheet" href="https://www.google.com/cse/style/look/default.css" type="text/css" />
         <script src="{{ STATIC_URL }}piplmesh/jquery/jquery.js" type="text/javascript"></script>
+        {% load pushserver %}
```
mitar: Load tags all together at the beginning of the file.

piplmesh/frontend/templates/base.html
```diff
@@ -8,6 +8,24 @@
...
+            
+        <script type="text/javascript">
+            var updates_url = '{% filter escapejs %}{% updates_url "a" %}{% endfilter %}';
+
+            function updateAnswer(data) {
+                if (data.value.action == 'JOIN') {
+                    $('#userlist').append('<li>'+data.value.message+'</li>');
+                }
+                else if (data.value.action == 'PART') {
+                    $("ul > li").remove(":contains("+data.value.message+")");
```
mitar: Use `'` for constant strings.

piplmesh/frontend/templates/base.html
```diff
@@ -8,6 +8,24 @@
...
+                }
+            }
+            
+            registerUpdatesProcessor('answer', updateAnswer);
```
mitar: Of course type should be something else, also names. Also `message` should not be payload.

piplmesh/frontend/templates/base.html
mitar: Are you sure you can just concatenate like that strings without escaping?

mitar: Anyway, you could also do:

```
$('#userlist li:contains(' + data.value.message + ')').remove();
```

piplmesh/frontend/templates/base.html
```diff
@@ -8,6 +8,24 @@
         <link rel="stylesheet" href="{{ STATIC_URL }}piplmesh/css/style.css" type="text/css" media="screen" />
         <link rel="stylesheet" href="https://www.google.com/cse/style/look/default.css" type="text/css" />
         <script src="{{ STATIC_URL }}piplmesh/jquery/jquery.js" type="text/javascript"></script>
+        {% load pushserver %}
+        <script type="text/javascript" src="{{ STATIC_URL }}pushserver/updates.js"></script>
+
```
mitar: No need for extra lines.

piplmesh/frontend/templates/base.html
agiz: And contains probably is not good choice here either.

piplmesh/frontend/templates/userlist.html
```diff
@@ -0,0 +1,4 @@
+{% load i18n %}
```
mitar: What is this file for? Just include it above in `home.html`, no?

mitar: No need to load `i18n`.

piplmesh/frontend/templates/base.html
mitar: Yes, but we will deal with this in some other ticket. (You should leave something also for others.) Use `:contains` for now.

piplmesh/settings.py
```diff
@@ -2,6 +2,7 @@
 #
 # Development Django settings for PiplMesh project.
 
+import datetime
```
mitar: You can make this into one import line.

piplmesh/account/views.py
agiz: This module is not needed anymore.

piplmesh/account/views.py
agiz: This module is not needed anymore.

piplmesh/frontend/templates/base.html
agiz: processdata.js?

piplmesh/account/models.py
mitar: Also use it like that. So that after `connections` list becomes empty, you do not mark user as offline unless some time has gone since `connection_last_unsubscribe`.

piplmesh/frontend/templates/userlist.html
agiz: I removed userlist.html all-together. Is it still present?

piplmesh/frontend/templates/base.html
mitar: I would have file for all PiplMesh JavaScript. Except search. Maybe we could name it based on page view name? So `home.js`?

piplmesh/frontend/templates/userlist.html
mitar: Yes. See the "diff" tab of pull request.

But I like the idea of having that in a separate file. This is good to make things modular.

mitar: You just include that file. Check the include template tag.

piplmesh/settings.py
```diff
@@ -152,8 +153,62 @@
...
+            'passthrough': 'http://127.0.0.1:8000/passthrough',
+        },
+        {
+            'type': 'publisher',
+            'url': r'/send-update/([^/]+)',
+        },
+    ),
+}
+
+CELERY_RESULT_BACKEND = "mongodb"
```
mitar: Constant string.

But we will not use that backend later. :-) You probably had problems with that when you tried to cancel tasks. :-)

piplmesh/settings.py
```diff
@@ -152,8 +153,62 @@
...
+CELERY_MONGODB_BACKEND_SETTINGS = {
+    "host": "127.0.0.1",
```
mitar: Same.

piplmesh/settings.py
```diff
@@ -152,8 +153,62 @@
...
+    "port": 27017,
+    "database": "celery",
+    "taskmeta_collection": "my_taskmeta"
```
mitar: Comma at the end.

piplmesh/frontend/templates/userlist.html
agiz: I do not follow now :) I moved

```
<div>
    <ul id="userlist"></ul>
</div>
```

to the end of home.html. Good for now?

piplmesh/settings.py
```diff
@@ -152,8 +153,62 @@
...
+}
+
+BROKER_BACKEND = "mongodb"
+BROKER_HOST = "localhost"
+BROKER_PORT = 27017
+BROKER_USER = ""
+BROKER_PASSWORD = ""
+BROKER_VHOST = "celery"
+
+CELERY_IMPORTS = ('piplmesh.tasks', )
```
mitar: Split into multiple lines.

piplmesh/tasks.py
```diff
@@ -0,0 +1,31 @@
+import datetime
```
mitar: Move this file to `piplmesh.frontend`.

piplmesh/frontend/templates/userlist.html
mitar: I am saying that you still have a stand-alone file with that in pull request. And that it is better to include that file at the end of `home.html` then to have content directly in `home.html`.

piplmesh/account/models.py
agiz: I have already done so (the second part). Check tasks.py. Good?

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,56 @@
...
+
+@dispatch.receiver(signals.channel_subscribe)
+def process_channel_subscribe(sender, request, channel_id, **kwargs):
+    request.user.update(
+        push__opened_connections = {
+            'http_if_none_match': request.META['HTTP_IF_NONE_MATCH'],
+            'http_if_modified_since': request.META['HTTP_IF_MODIFIED_SINCE'],
+            'channel_id': channel_id
+        },
+        set__last_access = datetime.datetime.now()
```
mitar: You should keep track of last unsubscribe, not of last subscribe. (OK, last subscribe could be interesting, but last unsubscribe is what is necessary.)

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,56 @@
...
+
+from piplmesh.account import models
+
+@dispatch.receiver(signals.channel_subscribe)
+def process_channel_subscribe(sender, request, channel_id, **kwargs):
+    request.user.update(
+        push__opened_connections = {
+            'http_if_none_match': request.META['HTTP_IF_NONE_MATCH'],
+            'http_if_modified_since': request.META['HTTP_IF_MODIFIED_SINCE'],
+            'channel_id': channel_id
```
mitar: Comma at the end.

piplmesh/frontend/templates/base.html
agiz: Like this:

```
{% load pushserver %}
{% load i18n %}<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
```

or:

```
{% load pushserver %}{% load i18n %}<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
```

Does it matter here?

piplmesh/frontend/templates/base.html
agiz: 'answer' -> 'response'?

> also names

I do not understand this.

`message` -> `server`?

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,56 @@
...
+        },
+        set__last_access = datetime.datetime.now()
+    )
+
+    channel_id = 'a'
```
mitar: No, don't send anything in signal handler. Do both subscribe and unsubscribe in the task. Signal handler should be as lightweight as possible. What if many many users will be opening and closing connections? This means load will increase with number of users active on the site. We want to keep that as constant as possible.

Furthermore, you cannot send for each opening something to all other users! This will explode! Because for each send, all users unsubscribe and subscribe immediately afterwards.

piplmesh/frontend/views.py
agiz: I will just remove this from subscribe signal.

piplmesh/frontend/views.py
agiz: I ended up putting too much commas and it stopped working silently :)

piplmesh/frontend/templates/base.html
mitar:
```
{% load i18n pushserver %}<!DOCTYPE ...
```

There should be nothing before `<!DOCTYPE ...` in output, not ever newline!

piplmesh/frontend/templates/base.html
mitar: Type: `userlist`

`message` -> `username`?

The point is what are you sending. You could send simply:

```
{
    'type': 'userlist',
    'action': 'JOIN',
    'username': request.user.username,
}
```

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,56 @@
...
+
+@dispatch.receiver(signals.channel_unsubscribe)
+def process_channel_unsubscribe(sender, request, channel_id, **kwargs):
+    models.User.objects(
+        id = request.user.id,
+        opened_connections = {
+            'http_if_none_match': request.META['HTTP_IF_NONE_MATCH'],
+            'http_if_modified_since': request.META['HTTP_IF_MODIFIED_SINCE'],
+            'channel_id': channel_id
+        }
```
mitar: Comma at the end.

piplmesh/frontend/views.py
agiz: What if I put additional conditions here. I had `logged` variable here that cleared when task cleared user `connections`. This way there are no sequential subscriptions here. Putting this in tasks.py will cause delay. Is that really the best choice here?

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,56 @@
...
+                    'action': 'JOIN',
+                    'message': request.user.username
+                }
+            }
+        )
+
+@dispatch.receiver(signals.channel_unsubscribe)
+def process_channel_unsubscribe(sender, request, channel_id, **kwargs):
+    models.User.objects(
+        id = request.user.id,
```
mitar: OK, I would argue here that for the consistency sake we should not have space around `=`, same as we do not have them for inline arguments.

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,56 @@
...
+        opened_connections = {
+            'http_if_none_match': request.META['HTTP_IF_NONE_MATCH'],
+            'http_if_modified_since': request.META['HTTP_IF_MODIFIED_SINCE'],
+            'channel_id': channel_id
+        }
+    ).update_one(unset__opened_connections__S = 1)
+
+    # TODO: Race condition??
+    request.user.update(pull__opened_connections = None)
```
mitar: We are not having spaces around `=` elsewhere, do we?

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,56 @@
...
+    models.User.objects(
+        id = request.user.id,
+        opened_connections = {
+            'http_if_none_match': request.META['HTTP_IF_NONE_MATCH'],
+            'http_if_modified_since': request.META['HTTP_IF_MODIFIED_SINCE'],
+            'channel_id': channel_id
+        }
+    ).update_one(unset__opened_connections__S = 1)
+
+    # TODO: Race condition??
```
mitar: No, there is no race condition here. It does not matter if this command or some other command clears all `None` out.

piplmesh/frontend/templates/base.html
agiz: Ok, but I do not get the rest then. Do we still have use one `registerUpdatesProcessor` or does each type has its own `registerUpdatesProcessor`?

piplmesh/frontend/views.py
agiz: Ok, will fix that but I thought we are using spaces that is why I added them. And what about colons?

piplmesh/frontend/views.py
agiz: Even if tasks gets scheduled in between?

agiz: I mean isn't there some command in mongo that fails when there are None types in an array? I saw it a lot during testing but I am not sure what caused it.

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,56 @@
...
+    request.user.update(pull__opened_connections = None)
+    request.user.update(set__last_access = datetime.datetime.now())
```
mitar: Only here race condition could happen, but it does not matter, because even if some timestamp overwrites one later, it is still good enough for us.

piplmesh/frontend/views.py
mitar: No, MongoEngine could fail. When it tries to create an embededobject out of it. MongoDB does not care for `None`.

But even with MongoEngine we are just adding to the list or checking the length, not really reading the values in there. It should work. But we will see.

mitar: In any case, than just that task will fail. If we see that happens often, we could wrap some exception handler so that only that one user who has `None` there is skipped. But let us see if this will be happening.

piplmesh/frontend/views.py
mitar: We have space after the colon, but not before.

piplmesh/frontend/templates/base.html
mitar: You register update handler with `registerUpdatesProcessor` function. For each type you can register different function. And then my code calls that function (or even more of them) for each type.

piplmesh/frontend/templates/base.html
mitar: Rename also to _updateUserlist_.

piplmesh/frontend/views.py
mitar: Delay is not a problem. Other users do not care or know that we are 10 seconds late (at worst) with information when user logged in. And for the user herself we will of course always show her as logged in. ;-) (Except when she will hide herself manually.)

mitar: So no, nothing more then updating the database here. Making a HTTP request to push server takes time. And it is a blocking call. What if there are some problems in connectivity. Really, no code which requires external services should be called in request handler. This should be done asynchronously. (The exception is database of course.)

piplmesh/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+
+from celery.task import task
+
+from pushserver.utils import updates
+
+from piplmesh.account import models
+
+@task
+def prune_users():
```
mitar: So this task should be then changed so that it takes care both of new users and users who left. And inform (sync) this information with everybody. How do you propose to do that?

piplmesh/frontend/views.py
agiz: So she is prevented to enter any text for 10 seconds (at worst)? Users will see her posting, but not yet as logged in.

agiz: And also what if tasks interval changes? Isn't 10 seconds pretty intensive?

mitar: No, we can always have in JavaScript code that if a post comes from an user for which we believe is offline, we switch the user to online. :-)

And yes, if you check Facebook, you will see things are sometimes inconsistent. But users do not really care. Much better is that things work!

mitar: Much less intensive then doing that on every request (if we will have many requests). :-)

The main thing is that it is much more predictable. And not tied to request processing. (Request processing means that Django process cannot handle others at the same time, this is why you have other processes doing that, even on other servers.)

agiz: And I still need one more variable `logged` in User document to prevent duplicate subscriptions?

mitar: I don't understand the question?

But yes, you should probably have some flag to store whether we believe user is logged in or not. And you update that flag in the task. And on change (atomic!) you update everybody of this change.

piplmesh/frontend/templates/base.html
agiz: I guess I moved too much. Do I leave `updates_url` here? Otherwise there is trouble.

agiz:
Updated pull request. Did not test it as I get
`TypeError("'int' object is not iterable",)` in tasks.py, line 16.

agiz:
I am out now :-)

piplmesh/frontend/templates/base.html
mitar: Of course. :-)

And include external file _after_ this block.

mitar: (Or register handler in jQuery document ready, which is probably even better.) So include together at top, but register in document ready.

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+                channel_id,
+                {
+                    'type': 'userlist',
+                    'action': 'PART',
+                    'username': user.username,
+                }
+            )
+        elif len(user.connections) > 0 and user.is_online == False:
+            user.update(
+                set__is_online=True,
```
mitar: This is not good. You should set to `True` only if it is still set to `False`. :-)

mitar: There are special atomic operations for such things. And you continue only if you succeeded.

mitar:
Check what the value of `user.connections` is. :-)

agiz:
Updates: fixed views.py, user counting should be ok now.

piplmesh/frontend/tasks.py
agiz: I read about these, did some tries but did not go far. `models.User.objects(__raw__={'getlasterror':1})` always returns `[]`. I used `save()` operation instead of `update()`.

agiz:
I think we are overcomplicating here. Why is it not good enough to send random `ETAGs`? By random I mean random SHA-1 hash for example. This way there is only  10^-42 chance of two connections colliding.

mitar:
Values of etags are defined by push protocol we are using. We cannot change that. We could add another random value for each connection we are making, but this I think is complicating at this stage. So current solution with checking three values is the best compromise in my opinion.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -0,0 +1,10 @@
+function updateUserlist(data) {
+    if (data.action == 'JOIN') {
+        $('#userlist').append('<li>' + escape(data.username) + '</li>');
```
mitar: BTW, better (cleaner) way to do this is:

```
$('<li/>').text(data.username).appendTo('#userlist');
```

mitar: `.text()` also automatically HTML escapes.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -0,0 +1,10 @@
...
+    }
+    else if (data.action == 'PART') {
+        $('#userlist li:contains(' + escape(data.username) + ')').remove();
```
mitar, line 7: Sadly, I do not know of a better way to do this. But are you aware, that this does not escape in proper way? It escapes only HTML, not also JavaScript or even jQuery selector syntax.

For now, I would leave `escape`, but add TODO for other types of escaping.

piplmesh/frontend/templates/userlist.html
```diff
@@ -0,0 +1,3 @@
+    <div>
```
mitar: You don't need to indent. Only indent based on levels in the file, not on how it looks in the output (we can die trying to achieve consistency there).

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,40 @@
+import datetime
+
+from django import dispatch
 from django.conf import settings
-from django.views import generic as generic_views
+from django.views import debug, generic as generic_views
```
mitar: `debug` is probably unnecessary?

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,40 @@
...
+        }
+    )
+
+@dispatch.receiver(signals.channel_unsubscribe)
+def process_channel_unsubscribe(sender, request, channel_id, **kwargs):
+    models.User.objects(
+        id=request.user.id,
+        connections__http_if_none_match=request.META['HTTP_IF_NONE_MATCH'],
+        connections__http_if_modified_since=request.META['HTTP_IF_MODIFIED_SINCE'],
+        connections__channel_id=channel_id
```
mitar: Comma at the end.

piplmesh/frontend/views.py
```diff
@@ -1,8 +1,40 @@
...
+    ).update_one(unset__connections__S=1)
+
+    request.user.update(
+        pull__connections=None,
+        set__connection_last_unsubscribe=datetime.datetime.now()
```
mitar: Comma at the end.

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
+import datetime
+
+from celery.task import task
+
+from pushserver.utils import updates
+
+from piplmesh.account import models
+
+@task
+def prune_users():
```
mitar: Probably task should be renamed now?

mitar:
You should merge with main repository, currently you have quite outdated version.

You should also add to documentation (to section of running daemons in separate terminals, before pushserver):

```
./manage.py celeryd --loglevel=info --concurrency=1 --maxtasksperchild=10 --beat
```

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+    chan_users = models.User.objects()
```
mitar: Do we really need this variable? If, then it should be called `users`. :-)

piplmesh/settings.py
```diff
@@ -152,8 +152,64 @@
...
+BROKER_VHOST = 'celery'
+
+CELERY_IMPORTS = (
+    'piplmesh.frontend.tasks',
 )
 
+CELERYBEAT_SCHEDULE = {
+    'runs-every-10-seconds': {
+        'task': 'piplmesh.frontend.tasks.prune_users',
+        'schedule': datetime.timedelta(seconds=10),
```
mitar: Move `10` to some settings constant (`CHECK_ONLINE_USERS_INTERVAL`?).

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+    for user in chan_users:
+        # TODO: Implement joining/parting subsystem
+        channel_id = 'a'
+        if len(user.connections) == 0 and user.connection_last_unsubscribe and datetime.datetime.now() > user.connection_last_unsubscribe + datetime.timedelta(seconds=30):
```
mitar: Move `30` to some top-level constant(`CHECK_ONLINE_USERS_RECONNECT_TIMEOUT`?), value based on `settings.CHECK_ONLINE_USERS_INTERVAL` (like `2 * settings.CHECK_ONLINE_USERS_INTERVAL`).

piplmesh/frontend/tasks.py
mitar: `check_online_users`?

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+            user.update(
+                set__connections=[],
```
mitar: Hm, but you know that `user.connections` is `[]`? Why setting it again?

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+                unset__connection_last_unsubscribe=1,
```
mitar: Why unsetting this?

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+                set__is_online=False,
```
mitar: Yes, I would set only this. But also here you have the same issue, maybe somebody already set it to `False`.

What I would do is check those conditions above in MongoDB, so make an atomic query which sets `is_online` to `False`, if `is_online` is `True`, `connections` is `[]` and `connection_last_unsubscribe` is old enough. If atomic query succeeds, you send an update to users.

If not, you try another atomic query, bellow.

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    'type': 'userlist',
+                    'action': 'PART',
+                    'username': user.username,
+                }
+            )
+        elif len(user.connections) > 0 and user.is_online == False:
+            # TODO: Check if this command succeeds and only then proceed to send_update
+            user.update(
+                set__is_online=True,
```
mitar: Also this could be a simple atomic query, you set `is_online` to `True`, if `is_online` is `False` and if `user.connections` is non equal `[]`. If query succeeds, you send an update.

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+
+from pushserver.utils import updates
+
+from piplmesh.account import models
+
+@task
+def prune_users():
+    chan_users = models.User.objects()
+    for user in chan_users:
+        # TODO: Implement joining/parting subsystem
```
mitar: I don't understand this TODO? I think this does not have a place here now?

piplmesh/settings.py
```diff
@@ -152,8 +152,64 @@
...
+BROKER_USER = ''
+BROKER_PASSWORD = ''
+BROKER_VHOST = 'celery'
+
+CELERY_IMPORTS = (
+    'piplmesh.frontend.tasks',
 )
 
+CELERYBEAT_SCHEDULE = {
+    'runs-every-10-seconds': {
```
mitar: You can rename this to `check_online_users` (same as task), or something, no?

mitar:
Otherwise it looks like it works. Great! Can you fix above comments so this can be pulled in?

piplmesh/frontend/views.py
agiz: Yes it is unnecessary.

piplmesh/frontend/tasks.py
agiz: So that upper `if` does not get triggered again.

piplmesh/frontend/static/piplmesh/js/home.js
agiz, line 7: It's just javascript, we will figure it out anyhow :). Plus `contains` is inappropriate here.

piplmesh/frontend/templates/base.html
agiz: That is still TODO. Is external file home.js ok? I only wrapped

```
$(function() {
    registerUpdatesProcessor('userlist', updateUserlist);
}
```

piplmesh/frontend/tasks.py
agiz: Hmm yes, this is bigger I think. I only know that user.connections _was_ `[]` somewhere in the near past and hope and assume it still is. This is not ideal and yes I agree, it is even less ideal if I am unsetting `connections`.

piplmesh/frontend/tasks.py
mitar: But that upper `if` will not get triggered again because `is_online` is already `False`.

mitar: (In my approach.)

piplmesh/frontend/static/piplmesh/js/home.js
mitar, line 7: Yes, just add TODO.

piplmesh/frontend/templates/base.html
mitar: Ah, yes, I didn't notice that you haven't yet done this. Use `$(document).ready(function () {` syntax, I prefer it over above because it is more verbose and even people unfamiliar with jQuery will understand what are you doing. Otherwise good, just wrap.

piplmesh/frontend/tasks.py
mitar: Yes, I have explained latter how I would do it. (Instead of `if` in the Python code, an atomic query checking for all those conditions.)

piplmesh/frontend/templates/base.html
agiz: Is this wrapping ok?

piplmesh/frontend/static/piplmesh/js/home.js
agiz: Changed.

piplmesh/frontend/static/piplmesh/js/home.js
agiz, line 7: Added.

piplmesh/frontend/views.py
agiz: Removed.

piplmesh/frontend/views.py
agiz: Added.

piplmesh/frontend/views.py
agiz: Added.

piplmesh/frontend/tasks.py
agiz: Changed.

piplmesh/frontend/tasks.py
agiz: No, we do not need that.

piplmesh/settings.py
agiz: Made into constant.

piplmesh/frontend/tasks.py
agiz: Made into constant.

piplmesh/frontend/tasks.py
agiz: Is it ok syntax-wise and style-wise?

piplmesh/frontend/tasks.py
agiz: Removed.

piplmesh/settings.py
agiz: Renamed.

agiz:
Pull updated. Merged with up to date repository version, made suggested changes.

piplmesh/frontend/templates/base.html
mitar: Perfect.

piplmesh/settings.py
```diff
@@ -152,8 +152,66 @@
...
+}
+
+CHECK_ONLINE_USERS_INTERVAL = 10
+
+CELERY_RESULT_BACKEND = 'mongodb'
+CELERY_MONGODB_BACKEND_SETTINGS = {
+    'host': '127.0.0.1',
+    'port': 27017,
+    'database': 'celery',
+    'taskmeta_collection': 'my_taskmeta',
```
mitar: Could just have `celery_taskmeta` ([default](http://ask.github.com/celery/configuration.html#celery-mongodb-backend-settings)) here, no?

mitar:
Great stuff. Checking it now.

piplmesh/settings.py
agiz: Changed.

docs/installation.rst
```diff
@@ -62,6 +62,7 @@ Running
 PiplMesh consist of many components, so multiple daemons should be running. Run
 the following in separate terminals::
 
+    ./manage.py celeryd --loglevel=info --concurrency=1 --maxtasksperchild=10 --beat
```
mitar, line 65: Ah, now I remembered why I have `--concurrency=1` in my development setup, because more does not work SQLite (which does not support multiple connections at the same time). But we do not use SQLite anymore.

Anyway, leave it in. Our development web server is still one-threaded. :-)

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
+import datetime
+
+from django.conf import settings
+
+from celery.task import task
```
mitar: Import only modules.

piplmesh/frontend/templates/base.html
```diff
@@ -7,6 +7,11 @@
         <link type="image/x-icon" href="{{ STATIC_URL }}piplmesh/images/favicon.ico" rel="icon" />
         <link rel="stylesheet" href="{{ STATIC_URL }}piplmesh/css/style.css" type="text/css" media="screen" />
         <script src="{{ STATIC_URL }}piplmesh/jquery/jquery.js" type="text/javascript"></script>
+        <script type="text/javascript" src="{{ STATIC_URL }}pushserver/updates.js"></script>
+        <script src="{{ STATIC_URL }}piplmesh/js/home.js" type="text/javascript"></script>
+        <script type="text/javascript">
+            var updates_url = '{% filter escapejs %}{% updates_url "a" %}{% endfilter %}';
```
mitar: OK, just this channel ID is a bit ugly currently. So we will probably have some general channel for "home" view (independent of walls), so you could name it home, no? (Also elsewhere in the code.)

mitar: Probably also make it some top-level constant (in `piplmesh.frontend.tasks`?) you then add to template context and use here?

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+
+from piplmesh.account import models
+
+CHECK_ONLINE_USERS_RECONNECT_TIMEOUT = 2 * settings.CHECK_ONLINE_USERS_INTERVAL
+
+@task
+def check_online_users():
+    for user in models.User.objects():
+        channel_id = 'a'
+        if models.User.objects( \
```
mitar: No need for `\`?

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+            id=user.id, \
+            is_online=True, \
+            connections=[], \
+            connection_last_unsubscribe__lt=datetime.datetime.now() - datetime.timedelta(seconds=CHECK_ONLINE_USERS_RECONNECT_TIMEOUT), \
+        ).update(set__is_online=False):
```
mitar, line 22: Probably `safe_update=True` argument should be used?

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+from pushserver.utils import updates
+
+from piplmesh.account import models
+
+CHECK_ONLINE_USERS_RECONNECT_TIMEOUT = 2 * settings.CHECK_ONLINE_USERS_INTERVAL
+
+@task
+def check_online_users():
+    for user in models.User.objects():
+        channel_id = 'a'
```
mitar: Move to top-level constant (`HOME_CHANNEL_ID`).

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+                    'type': 'userlist',
+                    'action': 'PART',
+                    'username': user.username,
+                }
+            )
+        elif models.User.objects( \
+            id=user.id, \
+            is_online=False, \
+            connections__ne=[], \
+        ).update(set__is_online=True):
```
mitar, line 35: Same.

mitar:
Looks very good now. Almost ready for pulling it in!

piplmesh/frontend/tasks.py
agiz, line 22: Isn't this `True` by default? http://readthedocs.org/docs/mongoengine-odm/en/latest/apireference.html

agiz, line 22: I am looking at section: Querying.

agiz:
This one took way longer than it should. Wait, still stuff todo.

piplmesh/frontend/tasks.py
mitar: To `home`. Hm, maybe it would be better to move this constant to `views.py`. (We will not use it much there, but it is somehow connected to views.)

piplmesh/frontend/templates/base.html
agiz: Huh, this is going to take time to figure that out.

piplmesh/frontend/tasks.py
mitar, line 22: True. From the description I got a feeling that you have to set it.

OK, then leave as it is.

piplmesh/frontend/tasks.py
agiz: Sorry, I deleted the question because I saw you already answered my question above.

Deleted comment was:
What should I rename default channel_id to?

piplmesh/frontend/templates/base.html
mitar: No, this is already done with other similar constants, so just check how it is done.

piplmesh/frontend/tasks.py
mitar: I wrote you above, to `home`.

piplmesh/frontend/templates/base.html
agiz: I know, but all of these are settings variables. I need BIG hint here. Do I have to change TemplateView to DetailView in HomeView class? Or something like this http://reinout.vanrees.org/weblog/2011/08/24/class-based-views-walkthrough.html maybe?

mitar: What? What then if they are settings variables? (And no, not all of them are setting variables.) So grep after one settings variable you see that we are using in templates and see how we are getting it into the template. There is no magic there for settings variables.

agiz: I know how are we getting settings variable in:
`{{ foo }}`

I know it's not `{{ piplmesh.views.frontend.HOME_CHANNEL_ID }}`. Help! This is only making me look stupid :/

mitar: No, you don't understand my question. How we are getting settings variable to be available in the template in the first place? Can you find one such variable and grep our source code?

agiz: Seems like with this `TEMPLATE_CONTEXT_PROCESSORS`.

mitar: Close.

agiz: No need to response anymore. After 2+ hours I brute forced my way out.

agiz:
Updated pull request: implemented default channel constant `HOME_CHANNEL_ID`, removed settings module from urls.py, removed back-slash line continuator from tasks.py.

piplmesh/frontend/tasks.py
agiz: This is funny ;)

piplmesh/frontend/tasks.py
agiz: Removed.

piplmesh/frontend/tasks.py
agiz: Fixed.

piplmesh/frontend/context_processors.py
```diff
@@ -11,4 +13,5 @@ def global_vars(request):
         'LOGIN_REDIRECT_URL': settings.LOGIN_REDIRECT_URL,
         'REDIRECT_FIELD_NAME': auth.REDIRECT_FIELD_NAME,
         'redirect_to': request.REQUEST.get(auth.REDIRECT_FIELD_NAME),
+        'HOME_CHANNEL_ID': views.HOME_CHANNEL_ID,
```
mitar: Group constants first, then variables. Maybe we could also add an empty line inbetween.

piplmesh/frontend/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+CHECK_ONLINE_USERS_RECONNECT_TIMEOUT = 2 * settings.CHECK_ONLINE_USERS_INTERVAL
+
+@task.task
+def check_online_users():
+    for user in models.User.objects():
+        if models.User.objects(
+            id=user.id,
+            is_online=True,
+            connections=[],
+            connection_last_unsubscribe__lt=datetime.datetime.now() - datetime.timedelta(seconds=CHECK_ONLINE_USERS_RECONNECT_TIMEOUT), \
```
mitar: Forgot `\` at the end here.

agiz: Removed.

piplmesh/frontend/context_processors.py
agiz: Ordered and separated.

piplmesh/settings.py
```diff
@@ -152,8 +152,66 @@
...
+}
+
+CHECK_ONLINE_USERS_INTERVAL = 10
+
+CELERY_RESULT_BACKEND = 'mongodb'
+CELERY_MONGODB_BACKEND_SETTINGS = {
+    'host': '127.0.0.1',
+    'port': 27017,
+    'database': 'celery',
+    'taskmeta_collection': 'messages',
```
mitar: I see, you read only my first comment and not my additional comments or changes to them. :-) I corrected myself, e-mail notifications are not always correct here.

agiz: Fixed.

agiz:
Yay!

PR #124, opened 2012-04-15
Small CSS fixes (FB login button and forms)

aljana:
Border on facebook login button is now removed. I also restructured html code for registration/login and changed css form styling so that is reusable.


aljana:
Hmm, the FB button doesn't really look good in IE and in Firefox neither. Do we really need to use a button? Actually the whole site looks a bit weird in IE9 (unwanted borders).

aljana:
The button behavior in IE (it adds extra padding) cannot be changed ([http://stackoverflow.com/questions/5847915/how-can-i-prevent-internet-explorer-from-adding-padding-to-buttons-when-pressed]), why not use an anchor instead of a button?

piplmesh/account/templates/registration/login.html
```diff
@@ -10,22 +10,38 @@
...
-                		<form method="post" action="{% url facebook_login %}" id="fb_login_form2">
-                    		{% csrf_token %}
-	                  		<p><button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></button></p>
-	              		</form>
-	              </li>
+                    <form method="post" action="{% url facebook_login %}" id="fb_login_form2">
+                        {% csrf_token %}
+                         <p><button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></button></p>
+                    </form>
+                  </li>
```
mitar: Indentation.

piplmesh/account/templates/registration/login.html
```diff
@@ -10,22 +10,38 @@
...
             <ul>
                 <li>
-                		<form method="post" action="{% url facebook_login %}" id="fb_login_form2">
-                    		{% csrf_token %}
-	                  		<p><button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></button></p>
-	              		</form>
-	              </li>
+                    <form method="post" action="{% url facebook_login %}" id="fb_login_form2">
+                        {% csrf_token %}
+                         <p><button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></button></p>
```
mitar: I was never sure why we need both `button` and `span`?

mitar: Indentation is a bit off?

piplmesh/account/templates/registration/login.html
```diff
@@ -10,22 +10,38 @@
...
+                                    <span class="required">{{ field.label }}<sup title="{% trans "Required" %}">*</sup></span>
+                                {% else %}
+                                    {{ field.label }}
+                                {% endif %}
+                            </label>
+                            {{ field }} 
+                        </li> 
+                    {% endfor %} 
+                </ul>
+                <div class="button"><button type="submit" class="auth_button" value="value">{% trans "Login" %}</button></div>
```
mitar: In general I use `buttons` for such `div` because sometimes there are more buttons (like reset button).

piplmesh/account/templates/registration/login.html
```diff
@@ -10,22 +10,38 @@
         <div class="login_form">
             <form method="post" action="{% url login %}">
                 {% csrf_token %}
-                {{ form.as_p }}
-                <p><button type="submit" class="auth_button" value="value">{% trans "Login" %}</button></p>
+                <ul class="form_fields">
+                    {% for field in form %}
+                        <li class="field_error">
+                            {{ field.errors.as_text }}
```
mitar: `as_text`? Is this really the best?

mitar:
Have you started this pull request from the latest official repository? It is saying to me that I cannot automatically merge (so there are some conflicts)?

mitar:
We are using buttons because we want form because we want POST request. GET requests should not modify state in any way (like logging user). We spoke about that at the class.

mitar:
And I do not like JavaScript solutions to this. So if you find another way, OK.

Anyway, try your best, how do other sites do that? What is Facebook proposed way?

aljana:
Most sites use button that is generated by FB javascript SDK and the same way is described in FB documentation (http://developers.facebook.com/docs/reference/plugins/login/) - but we don't want that I guess.
There are only 2 ways of submiting form, using button or javascript:document.form.submit().

mitar:
OK. Then do it like this: make it consistent. Do not try to hide the button, let it be the button, everywhere. Just style it in a nice way. So that maybe borders are also blue or something.

(But I really do not understand this, why it does not work: check our language changing links - they are also buttons, but they look like links. No?)

BTW: You have to merge this your branch with the latest from the main repository. And fix my other comments.

piplmesh/account/templates/registration/login.html
aljana: It is useful for setting two backgrounds and we need two because of the sprite image that we are using. Well we could use just one image and wouldn't have problems with aligning them. But I just accidentally found out that if I put position:relative on span, the button in IE starts to behave like in all other browsers  - wow - I'm not sure if this is a bug or a feature but it works so I would keep this span.  :)

mitar: I think also other more drastic CSS features disable button behavior. I do not know for IE but have experience with Safari. There it uses native buttons (you have problems styling) until you really style them in all possible ways (overriding even defaults with same values). Then it switches to your version.

So I am not sure if `position: relative` is really needed, maybe also some other CSS styles would trigger that. Search for Google on how to disable "native form element rendering" and so on in various browsers.

BTW, why are we using sprite image? Why not create our own image? I thought that we are using sprite because we are using official code. But now it seems that we have hard-coded they button, so we can also hard-code the image?

aljana:
Language links don't look OK in IE (it adds press effect) - I can fix them too because I just found out how. Will do all the fixes later.

mitar:
I can say only: ah this IE. :-)

(BTW, for me is OK if we do not support IE.)

piplmesh/account/templates/registration/login.html
aljana: Trust me I've searched the whole google with different keywords and the only answer I've got is: it is impossible. In IE of course, other browsers are not a problem. Well maybe i'm not very good at searching so I experimented a bit and found out this fix with positioning works. And it works okay in all browsers - why not use it? :)

And I have no idea why we are using sprites. I can fix this.

mitar: So you are saying that buttons [here](https://nodes.wlan-si.net/account/login/) also border?

I think those general guidelines for IE could be helpful:
- Any element with this hack needs to have position, so unless it already has a position, attach position: relative.
- It can act funny on some elements that are natively inline, even if you attach display: block, although not all the time (fun!).
- It also has issues with elements that don’t ‘have layout’. Attach zoom: 1; to get around this.

Zoom is for IE6, `position: relative` seems a [known hack](http://compass-style.org/reference/blueprint/ie/), check the source of _blueprint-ie-defaults_ for _button_. :-)

This could also be solved with use of some `reset.css` file the professor was talking about (I think you were missing at that time). You could find some good one. :-)

aljana: They don't border - the problem is not in the border but in the padding that is added when you press the button - so yes, they behave different than in other browsers. Adding position:relative to a button is not enough (I don't know what is it good for) - it has to be added to child element like span. 

Anyway I can find a good reset.css if you think we should use it. (but this is not really related since it won't help with this problem)

mitar: Good. Just make it the way you believe it is the best. Also decide for yourself about reset CSS. I have never used it myself. But maybe it is really a good practice.

mitar:

aljana:
I will, don't worry :) But later... tomorrow I hope. But I think I will skip reset.css this time and maybe include it in some other pull request.

mitar:
OK.

mitar:
Ping.

mitar:
You have to update from main repository.

aljana:
Pong. After merging with main repository I get the error: 'module' object has no attribute 'HOME_CHANNEL_ID'.  If I open views there really is no home channel id defined.

mitar:
Ah, yes. My mistake. Forgot to commit and push. :-) Please try again. :-)

aljana:
Can I correct this manually or do I have to merge again?

mitar:
You just pull from main repository again and it will probably do a fast-forward. Merge is done only when some merging had to be done (the same file has been changed in both branches, your and mine). In this case you probably have no new changes on your side.

aljana:
I think this is it now.

aljana:
Ahh typo. I meant "renamed" not "removed". How can I correct the commit? :)

mitar:
After the push ... almost impossible.

piplmesh/account/templates/registration/login.html
aljana: Fixed.

piplmesh/account/templates/registration/login.html
aljana: Fixed.

piplmesh/account/templates/registration/login.html
aljana: Fixed.

mitar:
Why are you still using sprites? And not simply two background images?

mitar:
Please name files with full names, not `btn`. :-)

aljana:
I thought I should use sprite for rollover images. Well  I can make two images, it's even more simple. But what's wrong with sprites? I will rename the file (after we decide what to use).

mitar:
Mah, probably nothing wrong, but I have bad experience with them. For example, when user resizes font size it happens that some other sprite is revealed.

aljana:
Well I'm not a fan of sprites either but they are somehow recommended for doing such stuff..  OK I'll make 2 backgrounds.

mitar:
Yes, it is optimization. But I would optimize once we see it is a problem. :-) Premature optimization is the root of all evil.

aljana:
And before I do it, tell me how should I rename the rollover image because I'm sure I won't choose the right name. :)

mitar:
Just be verbose (no abbreviations) and consistent (how are we naming other images? camel case? underscores?).

aljana:
I meant the postfix, should it be "roll", "hover", "on", something else?

mitar:
Be consistent: match that of CSS used.

aljana:
Done. But now we have a problem with (not) preloading. Why did I even have to add that roll over.. :)

mitar:
Hm. So sprites exist for a reason?

(So that you would have more commits.)

aljana:
Haha I could see that coming.. Well this can be fixed with javascript. But I guess you don't like that.

mitar:
I don't. :-)

aljana:
Ok done.

piplmesh/account/forms.py
```diff
@@ -94,8 +70,8 @@ def save(self):
         new_user.save()
 
         return self.cleaned_data['username'], self.cleaned_data['password2']
-    
+
```
mitar: Hm, unnecessary change?

mitar:
Why is `fb_login_btn.png` still there? Also `fb_login_button_hover.png` is now not used anymore?

piplmesh/account/templates/registration/login.html
```diff
@@ -5,30 +5,47 @@
...
-                <ul>
-                    <li>
-                        <form method="post" action="{% url facebook_login %}" id="fb_login_form2">
-                            {% csrf_token %}
-                            <p><button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Login with Facebook" %}</span></button></p>
-                        </form>
-                    </li>
-                    <li>{% comment %}Place for OpenID button.{% endcomment %}</li>
-                    <li><a href="{% url registration %}">{% trans "Register" %}</a></li>
+<div id="main">
```
mitar: Why you unindented everything for one level?

mitar: If I check other templates, we are always indenting. So the `registration.html` should be indented instead.

mitar:
Otherwise good. We could extract this form rendering to an external template someday, because as you see it is duplicated again and again. But not now and here.

piplmesh/account/templates/registration/login.html
aljana: Oops that certainly wasn't on purpose... someone fix this please...

ghost: Corrected.

piplmesh/account/templates/registration/login.html
```diff
@@ -22,13 +39,13 @@
                     <li>
                         <form method="post" action="{% url facebook_login %}" id="fb_login_form2">
                             {% csrf_token %}
-                            <p><button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Login with Facebook" %}</span></button></p>
+                            <p><button type="submit" class="fb_login_button" value=""><span class="fb_button_text">{% trans "Log in with Facebook" %}</span></button></p>
```
mitar: Please be consistent. We are using "login" elsewhere. Not "log in".

piplmesh/account/templates/registration/login.html
```diff
@@ -22,13 +39,13 @@
...
                         </form>
-                    </li>
+                      </li>
```
mitar: What is this additional space here?

piplmesh/account/templates/registration/login.html
```diff
@@ -9,12 +9,29 @@
...
+                                        <span class="required">{{ field.label }}<sup title="{% trans "Required" %}">*</sup></span>
+                                    {% else %}
+                                        {{ field.label }}
+                                    {% endif %}
+                                </label>
+                                {{ field }} 
+                            </li> 
+                        {% endfor %} 
+                    </ul>
+                    <div class="button"><input type="submit"  value="{% trans "Login" %}"></div>
```
mitar: Double space before `value`.

piplmesh/account/templates/registration/registration.html
```diff
@@ -5,30 +5,31 @@
...
-            <ul>
-                <li><input type="submit" value="{% trans "Register" %}" /></li>
-                <li><input type="reset" value="{% trans "Reset" %}" /></li>
-            </ul>
-        </div>
-    </form>
-</div>
+            <div class="buttons">
+                <input type="submit" value="{% trans "Register" %}" />
+                <input type="reset" value="{% trans "Reset" %}" />
```
mitar: Just one question I asked myself now: have you ever ever in your life used reset button? I have not. Maybe we do not need it? Would this make the form nicer?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -7,22 +7,100 @@ body {
...
+}
 .gravatar {
     border: 0px;
 }
 
-/* Styling forms*/
-ul.errorlist {
-    list-style: none;
+/* -------------------- FORMS --------------------- */
+form input[type="submit"], form input[type="reset"] {
```
mitar, line 45: Don't we have `buttons` class on a div around those buttons exactly for the purpose of styling them?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -7,22 +7,100 @@ body {
...
+a, button {
+    outline: none;
+}
+
+a img {
+    border: none;
+}
+
+button {
+    position: relative;
```
mitar, line 29: This is also necessary for IE? Then comment it please.

piplmesh/account/templates/registration/login.html
mitar: `button` class -> `buttons` class.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -7,22 +7,100 @@ body {
...
-    list-style: none;
+/* -------------------- FORMS --------------------- */
+form input[type="submit"], form input[type="reset"] {
+    border: medium none;
+    font-weight: bold;
+    min-width: 105px;
+    padding: 5px;
+}
+
+form input[type="email"], form input[type="text"], form input[type="password"], form textarea, form select {
```
mitar: Is not this simply all input form elements inside `form_fields` class?

aljana:
I made a special template for the form so it doesn't repeat but it caused an error because the translations were not in the right place anymore. So I had  to correct .po translation file... I followed instructions in our documentation and tried to compile it but I got the error: 'msgfmt' is not recognized as an internal or external command, operable program or batch file. I guess I have to add something to sys path but I have no idea what since it's not in the documentation?

mitar:
Probably Windows?

mitar:
https://docs.djangoproject.com/en/dev/topics/i18n/translation/#gettext-on-windows

mitar:
How is it going?

mitar:
Don't forget to update from main repository.

piplmesh/account/templates/registration/login.html
aljana: Fixed.

piplmesh/account/templates/registration/registration.html
aljana: Removed.

piplmesh/account/templates/registration/login.html
aljana: Fixed.

piplmesh/frontend/static/piplmesh/css/style.css
aljana: Fixed.

aljana:
Ok  I've solved all the problems with several installation and merged with master - is it ok?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -7,22 +7,98 @@ body {
...
+}
+
+.form_fields .field label {
+    float: left;
+    width: 16em;
+    padding-right: 2em;
+    margin-left: -13em;
+}
+
+/* Styling nested lists such as radio buttons. */
```
mitar: No dot at the end.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -172,16 +248,18 @@ input.gsc-search-button {
...
     text-decoration: none;
 }
 
+.auth_button:hover span {
+    text-decoration: underline;
+}
+
 /* ---------------- FB LOGIN BUTTON ------------------- */
 .fb_login_button {
-    background: #29447E url('../images/fb_sprite.png');
```
mitar, line 183: Why is `fb_sprite.png` then still in commit?

piplmesh/frontend/templates/form/form.html
```diff
@@ -1,24 +1,19 @@
```
mitar, line 0: I would have whole form in here. For example, check [how it is done in nodewatcher](https://github.com/wlanslovenija/nodewatcher/blob/master/nodewatcher/frontend/templates/form.html) (the structure is a bit different, but you can get the idea).

Support for fieldsets we do not yet do, I recommend we add support for it to our forms. (There is a [mixin in nodewatcher](https://github.com/wlanslovenija/nodewatcher/blob/master/nodewatcher/frontend/account/metaforms.py#L141) code for that. (This could be go to django-missing, someday, probably.))

Then I think that for `next`, we have named context processor variable `redirect_to`, in above code it is `context_processor_next`, I would in fact rename it to `request_get_next`.

You can then do something [like this](https://github.com/wlanslovenija/nodewatcher/blob/master/nodewatcher/frontend/templates/registration/registration_form.html). (So, using `{% with %}` template tag, be careful, in nodewatcher's there is old syntax.)

mitar, line 0: Ah, yes. Also it is useful to add `form` class to this form, so that you can style only that. To not style `form` HTML element, but `form` class. Because form HTML element you often use also for some special forms/buttons/links around the web page.

piplmesh/frontend/static/piplmesh/css/style.css
aljana, line 183: I deleted it in the next commit.

mitar, line 183: Aha. Hm. OK, this is some new GitHub feature. It displays also deleted files. I now see that there is one minus icon before. OK.

piplmesh/frontend/templates/form/form.html
aljana, line 0: I fixed the form and refactored that variable. Mixin is for another pull request I guess?

aljana, line 0: Oh, and I haven't use "with" tag because it doesn't work with the URL (I wanted to do `{% include "form/form.html" with form_action={% url login %} %}`) and the old way works just fine?

mitar, line 0: `url` tag knows how to put results into the context itself. :-) [See](https://github.com/wlanslovenija/nodewatcher/blob/master/nodewatcher/frontend/templates/registration/registration_form.html#L16).

mitar, line 0: > Mixin is for another pull request I guess?

As you want.

aljana, line 0: I did it that way but I couldn't do it using `with`, I tried different syntaxes but no one worked.

mitar, line 0: I don't understand. What we are speaking now about? `url` tags knows how to populate template context by itself. You use `with` tag only to rename or set other values to the template context variable names `form.html` expects.

aljana, line 0: Aha I get it, so the way it is now is OK?
There is also one problem: now some other templates (change password, account) are without a label on a button - most probably - I cannot check because I get the error when I try to open my profile. Should I fix this anyway?

mitar, line 0: You have to configure a context variable with a label. You use `with` for that. :-)

Yes, you should fix it. But first fix that your local version works. :-) You can always logout. And register again. :-)

aljana, line 0: But why would I use `with` if it works the way I did it now?

So the user profile should work? Because I get: `Generic detail view UserView must be called with either an object pk or a slug.` And I have no idea why.

mitar, line 0: > But why would I use `with` if it works the way I did it now?

Because `with` removes the variable at the end of the scope. Without `with` you put the variable into the scope until the end of template, with `with` you limit this scope. It means that there is less chance that you overwrite something else and so on.

> So the user profile should work?

Yes. You do not have right version of mongogeneric installed, I bet.

aljana, line 0: Ok tnx... And you're probably right, the installation of mongogeneric was quite problematic, will check this later.

piplmesh/account/templates/user/login.html
```diff
@@ -8,21 +8,16 @@
...
-                <form method="post" action="{% url login %}">
-                    {% csrf_token %}
-                    {{ form.as_p }}
-                    {% with next|default:redirect_to|default:LOGIN_REDIRECT_URL as next_url %}
-                        <input type="hidden" name="{{ REDIRECT_FIELD_NAME }}" value="{{ next_url }}" />
-                    {% endwith %}
-                    <p><button type="submit" class="auth_button" value="value">{% trans "Login" %}</button></p>
-                </form>
+                {% url login as form_action %}
+                {% trans "Login" as form_submit %}
```
mitar: Use please with here. To not ditty context more then necessary:

```
{% with form_submit=_("Login") %}
    {% url login as form_action %}
    {% include "form/form.html" %}
{% endwith %}
```

piplmesh/frontend/templates/form/form.html
mitar, line 0: Windows. Yes. :-) I heard is a bit "challenging" to install things there with new requirements. :-) Somebody could put some guidelines to our Etherpad. :-)

piplmesh/account/templates/user/login.html
```diff
@@ -8,21 +8,16 @@
...
+                {% include "form/form.html" %}
             </div>
             <div id="other_login_options">
                 <ul>
                     <li>
                         <form method="post" action="{% url facebook_login %}" id="fb_login_form2">
                             {% csrf_token %}
-                            <p><button type="submit" class="fb_login_button" value="value"><span class="fb_button_text">{% trans "Login with Facebook" %}</span></button></p>
+                            <p><button type="submit" class="fb_login_button" value=""><span class="fb_button_text">{% trans "Login with Facebook" %}</span></button></p>
```
mitar: Is `value="` really needed?

piplmesh/frontend/templates/form/form.html
```diff
@@ -1,24 +1,31 @@
...
-                <label for="{{ field.id_for_label }}">
-                    {% if field.field.required %}
-                        <span class="required">{{ field.label }}<sup title="{% trans "Required" %}">*</sup></span>
-                    {% else %}
-                        {{ field.label }}
-                    {% endif %}
-                </label>
-            </li>
-            <li class="form_textbox">{{ field }}</li>
+
```
mitar, line 10: No empty line.

piplmesh/frontend/templates/form/form.html
```diff
@@ -1,24 +1,31 @@
...
+{% if form.is_multipart %}
+    <form action="{{ form_action }}" method="post" enctype="multipart/form-data">
+{% else %}
+    <form action="{{ form_action }}" method="post">
+{% endif %}
+        {{ form.non_field_errors }}
+        {% csrf_token %}
+        {% with next|default:request_get_next|default:LOGIN_REDIRECT_URL as next_url %}
```
mitar: No, `default:LOGIN_REDIRECT_URL` is not good in generic form template. You set this with `with` before including the form in templates you want it.

piplmesh/frontend/templates/form/form.html
mitar, line 0: A comment describing this form template would be useful.

Also `spaceless` tag is useful to remove all this newlines and so on. Because newlines in HTML are understood as spaces in inline context, you could get some unwanted space around labels, buttons and so on. Maybe in our case this is nothing wrong. But check it, maybe things get better. Or not.

mitar:
Please update the pull request with latest from main repository.

piplmesh/frontend/templates/form/form.html
aljana: I have no idea how, I don't understand what this code does. Why do we need two default values?

mitar: No, we don't. That's why I am saying to remove it. ;-)

Originally, that `with` statement was used only for login. So it was OK to try, one after another:
- `next` from current template context
- `next` from request's GET
- `LOGIN_REDIRECT_URL` from `settings.py`

Of course the last one should not be in a general form template we are having here. But it could be used (just as an example) in `login.html` template or somewhere, before you include the form template, like this:

```
{% with next|default:request_get_next|default:LOGIN_REDIRECT_URL as next %}
    {% include "form/form.html %}
{% endwith %}
```

Because this sets `next` template context variable, in the included form template, `with` gets the value in `next` and uses that for `next_url`.

aljana: Ok, thanks.

mitar: Try to explain this in the comment in `form.html`. ;-)

piplmesh/account/templates/user/login.html
mitar: TODO.

piplmesh/account/templates/user/login.html
```diff
@@ -8,27 +8,23 @@
...
                         </form>
                     </li>
                     <li>
                         <form method="post" action="{% url twitter_login %}" id="twitter_login_form3">
                             {% csrf_token %}
-                            <p><button type="submit" class="twitter_login_button" value="value"><span class="twitter_button_text">{% trans "Login with Twitter" %}</span></button></p>
+                            <p><button type="submit" class="twitter_login_button" value=""><span class="twitter_button_text">{% trans "Login with Twitter" %}</span></button></p>
```
mitar: Same.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -7,22 +7,98 @@ body {
...
+    margin: 0.1em 0 0.1em 0;
+    padding: 0.2em 0.3em;
+    clear: left;
+    margin-left: 13em;
+    min-height: 2em;
+}
+
+.form_fields .field input {
+    width: 15.3em;
+    border: 1px solid #AFAFAF;
```
mitar, line 66: Lower case please.

mitar, line 66: Hm, I thought we are using lower case. Just make it consistent. ;-)

piplmesh/frontend/templates/form/form.html
mitar, line 0: TODO comment.

aljana, line 0: I don't know what should I comment... I added one line anyway.

mitar, line 0: You should comment in a similar way [how it is done in nodewatcher](https://github.com/wlanslovenija/nodewatcher/blob/master/nodewatcher/frontend/templates/form.html). You should tell users of this form template, what and how can they do anything with it.

piplmesh/frontend/static/piplmesh/css/style.css
aljana, line 66: There's nothing to fix. But note taken, I have to fix this in another issue.

piplmesh/account/templates/user/login.html
aljana: Fixed

piplmesh/account/templates/user/login.html
aljana: Fixed

piplmesh/frontend/templates/form/form.html
aljana, line 0: Hmm I hope "C/P" is ok. :)

piplmesh/frontend/templates/form/form.html
```diff
@@ -1,24 +1,40 @@
+{% comment %}
+This is a standard template for rendering forms.
+It uses `form` (form object), `form_submit` (submit button label) and `form_action` (form action URL) from
+the context.
+
+We use `next` from the context or use `request_get_next`, which is set to `next_url` from request arguments, to set
```
mitar:
```
which is set to request's GET `next` argument
```

piplmesh/frontend/templates/form/form.html
mitar, line 0: It would almost be if you would not change it to something wrong. ;-)

aljana, line 0: heh oops.. fixed.

mitar, line 0: Does it look understandable now? For me not. :-) I have been updating that my commit comment a bit. Maybe you have missed the last version?

mitar, line 0:
```
which is set to request's GET `next` argument
```

aljana, line 0: Ok. Is everything else checked and ok? Before I make thousand of "useless" commits... :)

mitar, line 0: I hope so. I will not ever check. ;-)

aljana, line 0: Fine with me. :) It's fixed anyway...

PR #130, opened 2012-04-16
Add string for query search

domenc:
Add string "Query result for" on search page - issue #86

### DIFF
--- a/piplmesh/frontend/templates/search.html
+++ b/piplmesh/frontend/templates/search.html
@@ -6,7 +6,7 @@
 
 {% block content %}
     {% if "q" in request.GET %}
-    <div id="google_query">{% trans "Query results for:" %}&nbsp;<b>{{ request.GET.q }}</b></div>
+    <div id="google_query">{% trans "Query results for:" %} <span class="google_query_text">{{ request.GET.q }}</span></div>
     {% endif %}
     <div id="cse" class="result"></div>
     <script src="https://www.google.com/jsapi" type="text/javascript"></script>
--- a/piplmesh/frontend/templates/search.html
+++ b/piplmesh/frontend/templates/search.html
@@ -5,6 +5,9 @@
 {% block title %}{% trans "Search" %}{% endblock %}
 
 {% block content %}
+    {% if "q" in request.GET %}
+    <div id="google_query">{% trans "Query results for:" %}&nbsp;<b>{{ request.GET.q }}</b></div>
+    {% endif %}
     <div id="cse" class="result"></div>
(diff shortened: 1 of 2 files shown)


piplmesh/frontend/templates/search.html
```diff
@@ -5,6 +5,9 @@
 {% block title %}{% trans "Search" %}{% endblock %}
 
 {% block content %}
+    {% if "q" in request.GET %}
+    <div id="google_query">{% trans "Query results for:" %}&nbsp;<b>{{ request.GET.q }}</b></div>
```
mitar: Indent. You do not have to use non-braking space. Just put a normal space there and this is it. Don't use `<b>`, use span with class and style that with CSS.

mitar:
BTW, with "Past search query string should be visible on result page." I though that in the search input box, past query string should be visible. So when you search and come to results, search input box should be empty.

domenc:
ok :)

piplmesh/frontend/templates/search.html
```diff
@@ -5,6 +5,9 @@
 {% block title %}{% trans "Search" %}{% endblock %}
 
 {% block content %}
+    {% if "q" in request.GET %}
+    <div id="google_query">{% trans "Query results for:" %} <span class="google_query_text">{{ request.GET.q }}</span></div>
```
mitar: Indentation! Inside of `if` you should indent once.

Use only _Results for:_ as a string. :-)

domenc:
'BTW, with "Past search query string should be visible on result page." I though that in the search input box, past query string should be visible. So when you search and come to results, search input box should be empty.'

Javascript not working -> $('[name='search']').val(query)

And image left of the input box. Why? Isn't go taht image is inside input box?

mitar:
Why JavaScript for that? I think you can do it without? In template rendering?

Image on the left? Just check how Firefox Google search looks like. Do same.

domenc:
Because, JS create input box (from google) and i haven't any other CSS in template for search bar (css is from google servers)

mitar:
Strange that Google does not support that. Anyway, you have to set value of input box after Google code added it. So you have to do that in callback from it. Are you doing that?

domenc:
I insert code in JS file (search_bar.js) after $('#google_results_link').prop('href', open_in_google_link);
and in HTML file after (all JS scripts) and doesnt work

mitar:
Again. This is not the correct place. You have to call this code after Google code has done its thing. In callback from Google search.

mitar:

domenc:
All in issue #144

PR #132, opened 2012-04-18
Fixed TODO: method validate_unique (Team 5)

ghost:


mitar:
What exactly do you think `validate_unique` should do?

ghost:
It checks forms that are set to be unique if they are unique right?

mitar:
Yes. And your code does what? So do you know what your code does?

mitar:
So we talked about this today and the point is that `validate_unique` should simply do nothing (pass) in our case. If you check Django implementation, it calls `validate_unique` on model instance, and our MongoEngine objects do not have this. So, just do `pass`, but comment/explain it accordingly.

ghost:
Fixed now, next time I will actually read the code. :)

PR #135, opened 2012-04-18
Facebook Fix (team 1)

MartinJakomin:
Fixed not able to login without facebook username and also fixed facebook avatar for people without facebook username.


piplmesh/frontend/templates/header.html
```diff
@@ -27,7 +27,11 @@
                     If user is already logged in, logout link is shown instead.
                 {% endcomment %}
                 {% if user.is_authenticated %}
-                    <li class="image"><img src="{% facebook_picture user.username "square" %}" alt="{% trans "User picture" %}"/></li>
+                    {% if user.facebook_id == None %}
```
mitar: I am not really sure this works. `None` is defined in Django templates?

piplmesh/frontend/templates/header.html
```diff
@@ -27,7 +27,11 @@
...
+                        <li class="image"><img src="{% facebook_picture user.username "square" %}" alt="{% trans "User picture" %}"/></li>
```
mitar: Duplication, duplication, duplication. No!

Something like this maybe?

```
<li class="image"><img src="{% facebook_picture user.facebook_id|default:user.username "square" %}" alt="{% trans "User picture" %}"/></li>
```

piplmesh/frontend/templates/header.html
mitar: Anyway, how can this be undefined? If it is not Facebook-linked account? Then probably there is no reason to query Facebook anyway. We use Gravatar instead. (And there is already a ticket, #63, and pull request, #116, for that.

MartinJakomin: It works for me, so I suppose it is defined.
Btw: My intention here was strictly to fix the bug, where you would get no facebook image if you havent got facebook username. For Gravatar as you said, there is another ticket.

mitar: But then, why not always just use Facebook ID?

piplmesh/frontend/templates/header.html
MartinJakomin: Did not suspect that this would work :)

Fixed.

piplmesh/frontend/templates/header.html
mitar: So who does not have Facebook ID?

MartinJakomin: If you login "the normal way" with username and password, you do not have facebook_id (it is set as "None"). 
Then this function tries to get facebook image with some strange id, and you end up with default facebook image and not with avatar you had.

mitar: I don't understand. Currently, we have or Facebook ID (because you have logged in through Facebook) or not (because you logged in with username and password). Why would you in the second case go and ask Facebook for user image? If Facebook ID is not set, you simply do not display an image (or use Gravatar, what we will do later).

MartinJakomin: Well, that was the way we used it before, I didnt want to change it.
Will change it now to no image (and later we will use Gravatar).

mitar: Good. Because you see, if you just leave things in, then somebody else will think that there is a reason for it. :-)

MartinJakomin: Changed.

piplmesh/frontend/templates/header.html
```diff
@@ -27,7 +27,14 @@
                     If user is already logged in, logout link is shown instead.
                 {% endcomment %}
                 {% if user.is_authenticated %}
-                    <li class="image"><img src="{% facebook_picture user.username "square" %}" alt="{% trans "User picture" %}"/></li>
+                    {% if user.facebook_id != None %}
```
mitar:
```
{% if user.facebook_id %}
```

piplmesh/frontend/templates/header.html
```diff
@@ -27,7 +27,14 @@
...
+                        <li class="image"><img src="{% facebook_picture user.facebook_id "square" %}" alt="{% trans "User picture" %}"/></li>
+                    {% else %}
+                        {% comment %}
+                        Here, we will use Gravatar.
+                        {% endcomment %}
+                        <li class="image"><img src=""/></li>
```
mitar: Just leave `<li>` empty. Or link to our default avatar image (we have already for Gravatar).

piplmesh/frontend/templates/header.html
MartinJakomin: Cant believe I actually did that :)

piplmesh/frontend/templates/header.html
MartinJakomin: I used our default Gravatar

piplmesh/frontend/templates/header.html
```diff
@@ -27,7 +27,11 @@
                     If user is already logged in, logout link is shown instead.
                 {% endcomment %}
                 {% if user.is_authenticated %}
-                    <li class="image"><img src="{% facebook_picture user.username "square" %}" alt="{% trans "User picture" %}"/></li>
+                    {% if user.facebook_id %}
+                        <li class="image"><img src="{% facebook_picture user.facebook_id "square" %}" alt="{% trans "User picture" %}"/></li>
+                    {% else %}
+                        <li class="image"><img src="static/piplmesh/images/unknown.png" height="50" width="50"/></li>
```
mitar: Check how you should reference static files (there are examples already, like our logo). Add `alt` (same as above). Add space before `/`.

MartinJakomin: I have just noticed that and have already change it.

piplmesh/frontend/templates/header.html
```diff
@@ -27,7 +27,11 @@
                     If user is already logged in, logout link is shown instead.
                 {% endcomment %}
                 {% if user.is_authenticated %}
-                    <li class="image"><img src="{% facebook_picture user.username "square" %}" alt="{% trans "User picture" %}"/></li>
+                    {% if user.facebook_id %}
+                        <li class="image"><img src="{% facebook_picture user.facebook_id "square" %}" alt="{% trans "User picture" %}"/></li>
+                    {% else %}
+                        <li class="image"><img src="{{ STATIC_URL }}piplmesh/images/unknown.png" height="50" width="50" alt="User Picture" /></li>
```
mitar, line 33: You are really careless. `alt` should be translatable string.

piplmesh/frontend/templates/header.html
```diff
@@ -27,7 +27,11 @@
                     If user is already logged in, logout link is shown instead.
                 {% endcomment %}
                 {% if user.is_authenticated %}
-                    <li class="image"><img src="{% facebook_picture user.username "square" %}" alt="{% trans "User picture" %}"/></li>
+                    {% if user.facebook_id %}
+                        <li class="image"><img src="{% facebook_picture user.facebook_id "square" %}" alt="{% trans "User picture" %}"/></li>
```
mitar, line 31: Space before `/`.

mitar:
23 commits for 6 lines of code ... Hmmm.

piplmesh/frontend/templates/header.html
MartinJakomin, line 31: That is not my code, I wrongly assumed that everything is already in order :S
And sorry for so many mistakes, I haven't slept for ages :S

mitar, line 31: I don't mind. You will just have to do new pull request with this cleaned code. :-)

mitar:
Please create a new pull request with only those changes. Otherwise (with those minor changes) things are good.

PR #138, opened 2012-04-19
Facebook Fix (team 1)

MartinJakomin:
Fixed not able to login without facebook username and also fixed facebook avatar for people without facebook username.

Now everything should be in order.


mitar:
There are still 20 commits in this pull request. Content is otherwise OK, but pull request is ugly. Please make a new pull request from another branch with only relevant commits (preferably one).

PR #142, opened 2012-04-19
Facebook Fix (team 1)

MartinJakomin:
After painfully long day, I have finally done it :S

PS: It seems like branching inherits commits. Why is that usefull ? It just overcomplicate the whole thing.

### DIFF
--- a/piplmesh/account/backends.py
+++ b/piplmesh/account/backends.py
@@ -64,7 +64,7 @@ class FacebookBackend(MongoEngineBackend):
         user, created = self.user_class.objects.get_or_create(
             facebook_id=fb.get('id'),
             defaults={
-                'username': fb.get('username'),
+                'username': fb.get('username', fb.get('first_name') + fb.get('last_name')),
                 'first_name': fb.get('first_name'),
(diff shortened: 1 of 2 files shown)


mitar:
Jupiii!

mitar:
Simply? Branches are branches in the commit tree, all commits before it are also included if those commits are not in repository you are pushing to.

MartinJakomin:
The thing is, I can not get rid of commits in our master branch, and every branch made from master will have that commits inherited (about 14 of them).
Also I dont think that u are correct here, because our master branch is exactly the same as your master (no changed files whatsoever) but commits are stil there when I try to pull !!
Is there anyway to delete them, or any way reset master branch ?

mitar:
Yes, you should never have in your master branch anything else except commits in from official repository. Otherwise you will/are having big problems.

I am not sure what is the best procedure here. Maybe check Google? When I have my own personal forks, I just delete the whole fork and create a new one. :-) But if there are many people working ... But probably you should do the same. Just take care you have your code stored somewhere and then put it again there.

Issue #143, opened 2012-04-19
Registration form fails when birthday is not set


mitar:
Registration form fails when birthday is not set.

mitar:
Done with #145.

PR #145, opened 2012-04-19
Pull request for issue #143 #team2

ghost:
#143: Corrected registration bug.

### DIFF
--- a/piplmesh/account/fields.py
+++ b/piplmesh/account/fields.py
@@ -11,6 +11,9 @@ GENDER_CHOICES = (
 )
 
 def limit_date(value, lower_limit, upper_limit, error):
+    if value is None:
+        return value
+    
     if upper_limit:
         tmp_value = value
         tmp_upper_limit = upper_limit
--- a/piplmesh/account/fields.py
+++ b/piplmesh/account/fields.py
@@ -11,8 +11,8 @@ GENDER_CHOICES = (
 )
 
 def limit_date(value, lower_limit, upper_limit, error):
-    if value is None:
-        return value
+    if not value:
+        return
     
     if upper_limit:
         tmp_value = value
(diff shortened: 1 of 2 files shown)


piplmesh/account/form_fields.py
```diff
@@ -27,7 +27,8 @@ def clean(self, value):
 
         def error():
             raise forms.ValidationError(self.error_messages['bounds'])
-
-        fields.limit_date(value, self.lower_limit, self.upper_limit, error)
+        
+        if value != None:
```
mitar: Checking for `None` value should be always done with `is`:

```
if value is not None:
```

Anyway, wouldn't it be better to test this in `limit_date` function? If `value` is `None` it returns immediately?

ghost:
It's better?

piplmesh/account/fields.py
```diff
@@ -11,6 +11,9 @@
 )
 
 def limit_date(value, lower_limit, upper_limit, error):
+    if value is None:
```
mitar: In this case it is probably enough:

```
if not value:
    return
```

piplmesh/account/form_fields.py
```diff
@@ -27,7 +27,7 @@ def clean(self, value):
 
         def error():
             raise forms.ValidationError(self.error_messages['bounds'])
-
+
```
mitar: Why change here?

piplmesh/account/fields.py
ghost: Corrected condition.

piplmesh/account/form_fields.py
ghost: Sorry, my mistake. :)

mitar: No problem, fix it. :-)

ghost: Done.

PR #146, opened 2012-04-21
Display gravatar image if user has our login

domenc:
Fix user picture if has our login (not FB login) - display gravatar image

### DIFF
--- a/piplmesh/frontend/templates/gravatar.html
+++ b/piplmesh/frontend/templates/gravatar.html
@@ -1,1 +1,2 @@
-<img src="{{ gravatar.url }}" width="{{ gravatar.size }}" height="{{ gravatar.size }}" class="gravatar" />
+{% load i18n %}
+<li class="image"><img src="{{ gravatar.url }}" height="{{ gravatar.size }}" width="{{ gravatar.size }}" alt="{% trans "User picture" %}" /></li>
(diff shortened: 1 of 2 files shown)


piplmesh/frontend/templates/gravatar.html
```diff
@@ -1 +1,2 @@
-<img src="{{ gravatar.url }}" width="{{ gravatar.size }}" height="{{ gravatar.size }}" class="gravatar" />
\ No newline at end of file
+{% load i18n %}
+<li class="image"><img src="{{ gravatar.url }}" height="{{ gravatar.size }}" width="{{ gravatar.size }}" alt="{% trans "User picture" %}" /></li>
```
mitar: No, `li` should not be in this template. We can use the same tag in multiple places. Only `img` should be there.

piplmesh/frontend/templates/header.html
```diff
@@ -30,7 +30,8 @@
                     {% if user.facebook_id %}
                         <li class="image"><img src="{% facebook_picture user.facebook_id "square" %}" alt="{% trans "User picture" %}" /></li>
                     {% else %}
-                        <li class="image"><img src="{{ STATIC_URL }}piplmesh/images/unknown.png" height="50" width="50" alt="{% trans "User picture" %}" /></li>
+                        {% load gravatar %}
```
mitar: Load tag libraries always all together at the top of the template.

piplmesh/frontend/templates/gravatar.html
domenc: OK

piplmesh/frontend/templates/header.html
domenc: OK

Issue #68, opened 2012-03-21
REST interface for posts and comments


mitar:
Define [REST](https://en.wikipedia.org/wiki/Representational_state_transfer) interface for posts, comments, and attachments. We will probably use [Tastypie](http://tastypieapi.org/) and its MongoEngine extension [django-tastypie-mongoengine](https://github.com/mitar/django-tastypie-mongoengine), but you can also research other libraries. Tastypie takes care of defining the API, like query and format interface, but we still have to define URL structure. Let it be like:

```
/api/post/
/api/post/id/
/api/post/id/comment/
/api/post/id/comment/id/
/api/post/id/attachment/
/api/post/id/attachment/id/
```

IDs in MongoDB are not numerical/sequential values. You will have to play a bit with [nested resources](http://django-tastypie.readthedocs.org/en/latest/cookbook.html#nested-resources).

For now, for API authentication, use Django authentication (so authentication of the Django request). Or even more: simply allow everything.

matevzmihalic:
If we have different documents for `Post` we can't have a structure you described above.
We would have to have `/api/textpost`, `/api/imagepost`,...
If we want to have only `/api/post`, we must change our document for Post.

mitar:
Can we override saving method so that we have some "type" field value which specifies "text", "image" or something and according to that correct Post child class is used when saving?

maticko:
This pull is not fully working yet, but some comments would be great, to see if we are on right way to solve this problem.

mitar:
Done with 4b8639ac766c1652698e78e5f3bb5a5f47f5b725.

PR #148, opened 2012-04-24
Issue #68 #team2

maticko:
Pull request for issue #68.


piplmesh/urls.py
```diff
@@ -1,8 +1,18 @@
 from django.conf.urls.defaults import patterns, include, url
 
 from piplmesh.account import views as account_views
+from piplmesh.api import api as api_resources
 from piplmesh.frontend import views as frontend_views
 
+from tastypie import api
```
mitar: Move this above, before piplmesh (piplmesh is the last group).

piplmesh/urls.py
```diff
@@ -17,4 +27,8 @@
     # Facebook
     url(r'^facebook/login/$', account_views.FacebookLoginView.as_view(), name='facebook_login'),
     url(r'^facebook/callback/$', account_views.FacebookCallbackView.as_view(), name='facebook_callback'),
+
+    # REST-style api
+    url(r'^post/',include('piplmesh.urls')),
```
mitar: What is this? This is some recursive stuff?

piplmesh/urls.py
```diff
@@ -1,8 +1,18 @@
...
+from piplmesh.api import api as api_resources
 from piplmesh.frontend import views as frontend_views
 
+from tastypie import api
+
+comment_api = api.Api(api_name='comment')
+comment_api.register(api_resources.CommentResource())
+post_api = api.Api(api_name="post")
+post_api.register(api_resources.PostResource())
+post_api.register(api_resources.UserResource())
```
mitar: Hm, we will only have Post resource in this case. At least in this ticket/pull request.

piplmesh/urls.py
```diff
@@ -1,8 +1,18 @@
...
 
 from piplmesh.account import views as account_views
+from piplmesh.api import api as api_resources
 from piplmesh.frontend import views as frontend_views
 
+from tastypie import api
+
+comment_api = api.Api(api_name='comment')
+comment_api.register(api_resources.CommentResource())
+post_api = api.Api(api_name="post")
```
mitar: Name it `v1`. This is for different versions.

piplmesh/urls.py
```diff
@@ -1,8 +1,18 @@
 from django.conf.urls.defaults import patterns, include, url
 
 from piplmesh.account import views as account_views
+from piplmesh.api import api as api_resources
 from piplmesh.frontend import views as frontend_views
 
+from tastypie import api
+
+comment_api = api.Api(api_name='comment')
+comment_api.register(api_resources.CommentResource())
```
mitar: Hm, aren't comment resources nested to post? They do not need to be independent.

piplmesh/urls.py
```diff
@@ -1,8 +1,18 @@
...
+post_api = api.Api(api_name="post")
+post_api.register(api_resources.PostResource())
+post_api.register(api_resources.UserResource())
+post_api.register(comment_api)
```
mitar: Nested APIs? This work?

piplmesh/urls.py
```diff
@@ -17,4 +27,8 @@
     # Facebook
     url(r'^facebook/login/$', account_views.FacebookLoginView.as_view(), name='facebook_login'),
     url(r'^facebook/callback/$', account_views.FacebookCallbackView.as_view(), name='facebook_callback'),
+
+    # REST-style api
+    url(r'^post/',include('piplmesh.urls')),
+    url(r'^api/',include(post_api.urls)),
```
mitar: So here should be something like:

```
url(r'^api/', include(v1_api.urls)),
```

piplmesh/urls.py
```diff
@@ -1,8 +1,18 @@
...
 from piplmesh.account import views as account_views
+from piplmesh.api import api as api_resources
 from piplmesh.frontend import views as frontend_views
 
+from tastypie import api
+
+comment_api = api.Api(api_name='comment')
+comment_api.register(api_resources.CommentResource())
+post_api = api.Api(api_name="post")
+post_api.register(api_resources.PostResource())
```
mitar: Good.

piplmesh/urls.py
mitar:
```
v1_api = api.Api(api_name='v1')
```

piplmesh/api/api.py
```diff
@@ -0,0 +1,37 @@
```
mitar: Maybe just rename this file to `resources.py`? Then you will also not have importing problems bellow.

piplmesh/api/api.py
```diff
@@ -0,0 +1,37 @@
+from tastypie import authorization
+from tastypie import fields as tastypie_fields
```
mitar: You can combine those two imports into one line.

piplmesh/api/api.py
```diff
@@ -0,0 +1,37 @@
...
+
+from piplmesh import account
+from piplmesh import models
+
+class  UserResource(resources.MongoEngineResource):
+
+    class Meta:
+        queryset = account.User.objects.all()
+        resource_name ='user'
+        fields = ['username','is_online']
```
mitar: Use tuples.

piplmesh/api/api.py
```diff
@@ -0,0 +1,37 @@
...
+    user = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='user',full=True)
+
+    class Meta:
+        object_class = models.Comment
+        resource_name = 'comment'
+
+
+class PostResource(resources.MongoEngineResource):
+    user = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='user', full=True)
+    comment = fields.EmbeddedDocumentField(embedded='piplmesh.api.resources.CommentResource',attribute='comment')
```
mitar: No, use `EmbeddedListField`. Comments are embedded in list!

piplmesh/api/api.py
mitar: I would suggest that you peek at [tests](https://github.com/mitar/django-tastypie-mongoengine/blob/master/tests/test_app/documents.py). :-)

piplmesh/api/api.py
```diff
@@ -0,0 +1,37 @@
...
+class  UserResource(resources.MongoEngineResource):
+
+    class Meta:
+        queryset = account.User.objects.all()
+        resource_name ='user'
+        fields = ['username','is_online']
+        allowed_methods = ['get']
+
+class CommentResource(resources.MongoEngineResource):
+    user = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='user',full=True)
```
mitar: `ForeignKey` is to link to relational-based objects.

mitar: Use `ReferenceField`.

mitar:
Good direction. Continue. :-)

piplmesh/urls.py
maticko: ok

piplmesh/urls.py
maticko: Fixed.

piplmesh/urls.py
maticko: Removed.

piplmesh/urls.py
maticko: Renamed.

piplmesh/urls.py
maticko: Moved.

piplmesh/api/api.py
maticko: Changed.

maticko: But this produces an error:
AttributeError 
Exception Value:    'module' object has no attribute 'ReferenceField'

In test_app this is used:
class CustomerResource(resources.MongoEngineResource):
    person = tastypie_fields.ForeignKey(to='test_app.api.resources.PersonResource', attribute='person', full=True)

mitar: Sorry, my mistake.

piplmesh/api/resources.py
```diff
@@ -0,0 +1,31 @@
+from tastypie import authorization, fields as tastypie_fields
+
+from tastypie_mongoengine import resources, fields
+
+from piplmesh import account
```
mitar: `from piplmesh.account import models as account_models`

piplmesh/api/resources.py
```diff
@@ -0,0 +1,31 @@
...
+from piplmesh.api import models
```
mitar: `from piplmesh.api import models as api_models`

piplmesh/api/resources.py
```diff
@@ -0,0 +1,31 @@
+from tastypie import authorization, fields as tastypie_fields
+
+from tastypie_mongoengine import resources, fields
```
mitar: `f` is before `r`.

piplmesh/urls.py
```diff
@@ -1,8 +1,15 @@
 from django.conf.urls.defaults import patterns, include, url
 
+from tastypie import api
+
 from piplmesh.account import views as account_views
+from piplmesh.api import resources as api_resources
```
mitar: `from piplmesh.api import resources` would be enough?

piplmesh/urls.py
```diff
@@ -1,8 +1,15 @@
...
 from piplmesh.frontend import views as frontend_views
 
+v1_api = api.Api(api_name="v1")
+v1_api.register(api_resources.PostResource())
+v1_api.register(api_resources.UserResource())
```
mitar: No need for user resource.

piplmesh/urls.py
```diff
@@ -1,8 +1,15 @@
 from django.conf.urls.defaults import patterns, include, url
 
+from tastypie import api
+
 from piplmesh.account import views as account_views
+from piplmesh.api import resources as api_resources
 from piplmesh.frontend import views as frontend_views
 
+v1_api = api.Api(api_name="v1")
```
mitar: Constant string -> `'`.

piplmesh/urls.py
```diff
@@ -17,4 +24,7 @@
     # Facebook
     url(r'^facebook/login/$', account_views.FacebookLoginView.as_view(), name='facebook_login'),
     url(r'^facebook/callback/$', account_views.FacebookCallbackView.as_view(), name='facebook_callback'),
+
+    # REST-style api
```
mitar: _REST API_

piplmesh/urls.py
```diff
@@ -17,4 +24,7 @@
...
+    url(r'^api/',include(v1_api.urls)),
```
mitar: Comma after `,`.

piplmesh/api/resources.py
```diff
@@ -0,0 +1,31 @@
...
+from piplmesh import account
+from piplmesh.api import models
+
+class UserResource(resources.MongoEngineResource):
+
+    class Meta:
+        queryset = account.models.User.objects.all()
+        resource_name = 'user'
+        fields = ('username','is_online')
+        allowed_methods = ['get']
```
mitar: Tuple!

piplmesh/api/resources.py
```diff
@@ -0,0 +1,31 @@
...
+
+from piplmesh import account
+from piplmesh.api import models
+
+class UserResource(resources.MongoEngineResource):
+
+    class Meta:
+        queryset = account.models.User.objects.all()
+        resource_name = 'user'
+        fields = ('username','is_online')
```
mitar: Space after comma.

piplmesh/api/resources.py
```diff
@@ -0,0 +1,31 @@
+from tastypie import authorization, fields as tastypie_fields
+
+from tastypie_mongoengine import resources, fields
+
+from piplmesh import account
+from piplmesh.api import models
+
+class UserResource(resources.MongoEngineResource):
+
```
mitar: Unnecessary empty line.

piplmesh/api/resources.py
```diff
@@ -0,0 +1,31 @@
...
+    class Meta:
+        queryset = account.models.User.objects.all()
+        resource_name = 'user'
+        fields = ('username','is_online')
+        allowed_methods = ['get']
+
+class CommentResource(resources.MongoEngineListResource):
+    user = tastypie_fields.ReferenceField(to='piplmesh.api.resources.UserResource', attribute='user',full=True)
```
mitar: Space after comma.

piplmesh/api/resources.py
```diff
@@ -0,0 +1,31 @@
...
+
+    class Meta:
+        object_class = models.Comment
+        resource_name = 'comment'
+
+class PostResource(resources.MongoEngineResource):
+    user = tastypie_fields.ReferenceField(to='piplmesh.api.resources.UserResource', attribute='user', full=True)
+    comment = fields.EmbeddedListField(embedded='piplmesh.api.resources.CommentResource',attribute='comment')
```
mitar: Space after comma.

piplmesh/api/resources.py
```diff
@@ -0,0 +1,31 @@
...
+
+    class Meta:
+        queryset = models.Post.objects.all()
+        resource_name = 'post'
+        #TODO: make proper authorisation. Current implementation is for development only.
```
mitar: Style TODO same as others (space before TODO, start with an upper case letter).

piplmesh/api/api.py
maticko: Renamed back to tastypie_fields.ForeignKey.

piplmesh/api/resources.py
maticko: Fixed.

piplmesh/api/resources.py
maticko: Fixed.

piplmesh/api/resources.py
maticko: Chenged.

piplmesh/urls.py
maticko: Ok.

piplmesh/urls.py
maticko: Removed.

piplmesh/urls.py
maticko: Changed.

piplmesh/urls.py
maticko: Changed.

piplmesh/urls.py
maticko: You mean space after comma? I fixed that.

piplmesh/api/resources.py
maticko: Changed.

piplmesh/api/resources.py
maticko: Done.

piplmesh/api/resources.py
maticko: Removed.

piplmesh/api/resources.py
maticko: Changed.

piplmesh/api/resources.py
```diff
@@ -0,0 +1,30 @@
...
+
+from piplmesh.account import models as account_models
+from piplmesh.api import models as api_models
+
+class UserResource(resources.MongoEngineResource):
+    class Meta:
+        queryset = account_models.User.objects.all()
+        resource_name = 'user'
+        fields = ('username', 'is_online')
+        allowed_methods = ('get')
```
mitar: `('get')` -> `('get',)`

This is one-element tuple, no?

piplmesh/api/resources.py
```diff
@@ -0,0 +1,30 @@
...
+        resource_name = 'comment'
+
+class PostResource(resources.MongoEngineResource):
+    user = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='user', full=True)
+    comment = fields.EmbeddedDocumentField(embedded='piplmesh.api.resources.CommentResource', attribute='comment')
+
+    class Meta:
+        queryset = api_models.Post.objects.all()
+        resource_name = 'post'
+        # TODO: Make proper authorisation. Current implementation is for development use only.
```
mitar: _authorisation_ -> _authorization_

piplmesh/api/resources.py
```diff
@@ -0,0 +1,30 @@
...
+class CommentResource(resources.MongoEngineListResource):
+    user = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='user', full=True)
+
+    class Meta:
+        object_class = api_models.Comment
+        resource_name = 'comment'
+
+class PostResource(resources.MongoEngineResource):
+    user = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='user', full=True)
+    comment = fields.EmbeddedDocumentField(embedded='piplmesh.api.resources.CommentResource', attribute='comment')
```
mitar: No _EmbeddedListField_?

mitar: _comment_ -> _comments_?

mitar:
And, does it work?

piplmesh/api/resources.py
maticko: 'EmbeddedListField' does not exist in fileds...  This is error I get if i try to run it.

maticko:
this http://127.0.0.1:8000/api/v1/?format=json :

returns : {"post": {"list_endpoint": "/api/v1/post/", "schema": "/api/v1/post/schema/"}}

this: http://127.0.0.1:8000/api/v1/post/?format=json

returns : 

{"meta": {"limit": 20, "next": null, "offset": 0, "previous": null, "total_count": 0}, "objects": []}

Is this supposed behavior?

piplmesh/api/resources.py
mitar: I can see `EmbeddedListField` in `tastypie_mongoengine.fields`?

mitar:
What about POST request? Can you create a post?

piplmesh/api/resources.py
mitar: Ping. :-)

piplmesh/api/resources.py
maticko: Ok. I figured out what was wrong. Now we use `EmbeddedListField`

piplmesh/api/resources.py
```diff
@@ -0,0 +1,30 @@
...
+
+class UserResource(resources.MongoEngineResource):
+    class Meta:
+        queryset = account_models.User.objects.all()
+        resource_name = 'user'
+        fields = ('username', 'is_online')
+        allowed_methods = ('get',)
+
+class CommentResource(resources.MongoEngineListResource):
+    user = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='user', full=True)
```
mitar: This should also be `author`, no?

piplmesh/api/resources.py
```diff
@@ -0,0 +1,30 @@
...
+
+from piplmesh.account import models as account_models
+from piplmesh.api import models as api_models
+
+class UserResource(resources.MongoEngineResource):
+    class Meta:
+        queryset = account_models.User.objects.all()
+        resource_name = 'user'
+        fields = ('username', 'is_online')
+        allowed_methods = ('get',)
```
mitar: Set `allowed_methods` to `()`. The idea is that we try to not expose user objects, but still allow referencing them.

piplmesh/urls.py
```diff
@@ -1,8 +1,14 @@
 from django.conf.urls.defaults import patterns, include, url
 
+from tastypie import api
+
 from piplmesh.account import views as account_views
+from piplmesh.api import resources
 from piplmesh.frontend import views as frontend_views
 
+v1_api = api.Api(api_name='v1')
+v1_api.register(resources.PostResource())
```
mitar, line 10: Add here also users.

piplmesh/api/resources.py
```diff
@@ -0,0 +1,30 @@
...
+class CommentResource(resources.MongoEngineListResource):
+    user = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='user', full=True)
+
+    class Meta:
+        object_class = api_models.Comment
+        resource_name = 'comment'
+
+class PostResource(resources.MongoEngineResource):
+    author = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='author', full=True)
+    comments = fields.EmbeddedListField(of='piplmesh.api.resources.CommentResource', attribute='comments')
```
mitar: Set `null=True` to allow empty list for comments.

piplmesh/api/resources.py
maticko: Fixed.

piplmesh/api/resources.py
maticko: Fixed.

piplmesh/urls.py
maticko, line 10: Fixed.

piplmesh/api/resources.py
maticko: Fixed.

maticko:
We still have to implement things with Textpost/imagepost stuff, right?

piplmesh/api/resources.py
```diff
@@ -0,0 +1,32 @@
+from tastypie import authorization, fields as tastypie_fields
+
+from tastypie_mongoengine import fields, resources
+
+from piplmesh.account import models as account_models
+from piplmesh.api import models as api_models
+
+#override dispatch_details ali pa post_detail / obj_create
```
mitar: Write as a TODO comment, when you do such comments. :-)

mitar:
Yes. So as I explained, I would add an additional field to Post document named `type` (which would still be stored) (or maybe expose internal `_type` field which is added by MongoEngine), which would in `obj_create` be used to override the class of the object created.

piplmesh/api/resources.py
maticko: I totaly forgot about this comment. I will delete this.

piplmesh/api/models.py
```diff
@@ -25,6 +26,7 @@ class Post(mongoengine.Document):
     created_time = mongoengine.DateTimeField(default=lambda: datetime.datetime.now())
     updated_time = mongoengine.DateTimeField()
     comments = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Comment))
+    type = mongoengine.StringField(max_length=TYPE_MAX_LENGTH)
```
mitar, line 29: Wouldn't it be better to use `choices` for this? So to provide only fixed set of possible types.

aljana, line 29: Right, will fix this.

aljana:
I was playing with this object creation but can't get it to work.. Any idea what's wrong? The post is created but it doesn't show the right fields because meta object class is still empty/NoneType.

piplmesh/api/resources.py
```diff
@@ -0,0 +1,38 @@
...
+        resource_name = 'comment'
+
+class PostResource(resources.MongoEngineResource):
+    author = tastypie_fields.ForeignKey(to='piplmesh.api.resources.UserResource', attribute='author', full=True)
+    comments = fields.EmbeddedListField(of='piplmesh.api.resources.CommentResource', attribute='comments', null=True)
+    
+    # TODO
+    def obj_create(self, bundle, request=None, **kwargs):
+        old = self._meta.object_class
+        self._meta_object_class =  getattr(api_models, bundle.data['type'])
```
mitar, line 29: `self._meta.object_class =` could be better than `self._meta_object_class =`, no? ;-)

mitar, line 29: There is also one space too much there.

And you should create some mapping between allowed values (choices) and models. This is a bit tricky to just `getattr`, but for testing is good enough.

mitar:
But it seems it is not enough. Fields are not processed because they are not in Post model.

mitar:
So also something else should be done.

One way could be to define Resource models for all Post types and then in-live change according to type field.

piplmesh/api/resources.py
aljana, line 29: I was testing the right code but I stashed it because I made too many changes everywhere. Then I wrote this again (and made a typo).. So it should work... EDIT: ah, this is not the only problem, ok..

piplmesh/api/models.py
aljana, line 29: But Mongoengine doesn't have a choice field? So I create custom TypeField?

mitar, line 29: No, choices is not a type of field, but an argument of allowed values for a given field. For integer field, this could be a list of integers, for string, list of strings.

That is a good idea to create a custom field, like we have done for gender. If you want. I think it will be some work because I will require a general type chooser field which will you then have to configure for our set of classes. (The reason is that you will get otherwise an import cycle, fields importing from models to get a list of models, and models importing from fields to get the field.)

piplmesh/api/resources.py
mitar, line 29: OK, so you are also getting posts of correct type in the database (if you inspect them with `./manage.py shell`)?

mitar, line 29:
```
from piplmesh.api import models as api
api.Post.objects()
```

aljana, line 29: Yes this works. So I only have to define other resource models and do the "in-live change" - I am clueless what and where this should be.

mitar, line 29: Yes, we have to find a way to do this. One would be to not use field type, but to use different [content-types](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7) for different post types. So route by that to those different resources. So instead of using URL, using HTTP field to chose which resource to invoke. Check how this is dispatched.

mitar, line 29: OK. I moved this to django-tastypie-mongoengine. It is in initial stage but probably can be used. Check tests there to see how to define such polymorphic documents.

mitar:
On last meeting we decide to modify schema. We will have one post type with attachments of different kind.

PR #149, opened 2012-04-25
Fixed TODO: query is limited so there's no iterating over all users

lazarm:


piplmesh/frontend/tasks.py
```diff
@@ -14,26 +14,23 @@
 @task.task
 def check_online_users():
     # TODO: Iterating over all users in Python could become really slow once there are millions of users, much better is to limit the query only to potentially interesting users (conditions are already bellow)
-    for user in models.User.objects():
-        if models.User.objects(
-            id=user.id,
+    for user in [users for users in models.User.objects(
```
mitar: This is really strange iteration?

lazarm:
Why, because of two for loops? I put all user objects that match the query and meet the condition into a list, and then do update for all users in this list. Wrong way? :)

mitar:
What is wrong with simply:

```
for user in models.User.objects(
```

piplmesh/frontend/tasks.py
```diff
@@ -14,26 +14,23 @@
...
             is_online=True,
             connections=[],
             connection_last_unsubscribe__lt=datetime.datetime.now() - datetime.timedelta(seconds=CHECK_ONLINE_USERS_RECONNECT_TIMEOUT),
-        ).update(set__is_online=False):
-            updates.send_update(
+            ) if users.update(set__is_online=False)]:
```
mitar: This is not good. There is a race condition. What if some other task (if tasks will be very long or multiple servers or whatever) changes `is_online` to `False` already: then you will send an update again. You have to set `is_online` to `False` in an atomic operation only if all conditions still hold true.

lazarm: Better?

piplmesh/frontend/tasks.py
```diff
@@ -14,13 +14,12 @@
...
-    for user in models.User.objects():
-        if models.User.objects(
-            id=user.id,
+    for user in models.User.objects(
             is_online=True,
             connections=[],
             connection_last_unsubscribe__lt=datetime.datetime.now() - datetime.timedelta(seconds=CHECK_ONLINE_USERS_RECONNECT_TIMEOUT),
-        ).update(set__is_online=False):
+            ):
+        if user.update(set__is_online=False):
```
mitar: Still race condition.

piplmesh/frontend/tasks.py
```diff
@@ -14,13 +14,12 @@
...
     # TODO: Iterating over all users in Python could become really slow once there are millions of users, much better is to limit the query only to potentially interesting users (conditions are already bellow)
-    for user in models.User.objects():
-        if models.User.objects(
-            id=user.id,
+    for user in models.User.objects(
             is_online=True,
             connections=[],
             connection_last_unsubscribe__lt=datetime.datetime.now() - datetime.timedelta(seconds=CHECK_ONLINE_USERS_RECONNECT_TIMEOUT),
-        ).update(set__is_online=False):
+            ):
```
mitar: Also indentation here is strange.

piplmesh/frontend/tasks.py
mitar: You should leave the update query as it was before, complicated and atomic. Just limit the user objects over which we go over.

lazarm: So should I limit user objects with the same query as update query, without id=user.id condition?

mitar: Maybe you should read documentation about queries we currently have and reasons for why they are like that. Also comments on the original pull request for this feature could help you. Then you will understand how the query works and what is that condition.

mitar:
Ping.

NejcZupec:
Request timeout for icmp_seq 0
Request timeout for icmp_seq 1
Request timeout for icmp_seq 2
Request timeout for icmp_seq 3
Request timeout for icmp_seq 4

Host lazarm is unreachable. Sorry :(

mitar:
???

mitar:
So somebody else from the team could pick this up?

NejcZupec:
That was joke :)

mitar:
Yes, a good one. :-) But still somebody else could pick it up ...

lazarm:
Sorry about that, I'm blaming george martin (and my lazy ass). I would still like to finish this, and if I won't be able to do it in  let's say three days, someone else can pick it up.
So if understand it correctly from here http://ask.github.com/celery/userguide/tasks.html#state , condition id=user.id has to be in atomic query to avoid race condition. But i still don't understand in what way we should limit the query. Wouldn't potentially interesting users be those who are online?

mitar:
Again. There are two things the code does:

1.) get a list of user candidates to flip the flag around (is_online: True -> False, False -> True)
2.) flips the flag around

The problem is, that between 1.) and 2.) there is some time and somebody else could flip the flag. So once we get candidates, we should make sure that in 2.) we flip the flag only if conditions are still valid. Not blindly do `user.update(set__is_online=...)`. You already have the code which makes sure that we flip the flag only if conditions are valid. The only problem is that we are checking this for too many users (all). We want only to limit the 1.). Not introduce the race condition between 1.) and 2.) (where there was no race condition before).

lazarm:
I already know that there is time between 1 and 2, and to avoid race condition, we need to refetch user parameters. i also  know that i shouldnt touch this part of the code, but at least i understand it now :) i was just asking if interesting users equals online users and should i check if user is online, before we do update query and send an update (it says that conditions are below)?

mitar:
> we need to refetch user parameters

Not refetch. Just limit the update query with assumed values. If values do not match, query is not run.

No, interesting users are like you have done it. So all those who match the assumed values, too. :-) So not just online users, but those who seems to not be online anymore.

lazarm:
update.

piplmesh/frontend/tasks.py
```diff
@@ -13,32 +13,39 @@
...
-        ).update(set__is_online=True):
-            updates.send_update(
-                views.HOME_CHANNEL_ID,
-                {
-                    'type': 'userlist',
-                    'action': 'JOIN',
-                    'username': user.username,
-                }
-            )
+	for user in models.User.objects(
```
mitar: Indendation!

piplmesh/frontend/tasks.py
```diff
@@ -13,32 +13,39 @@
...
+		).update(set__is_online=False):
+			updates.send_update(
+				views.HOME_CHANNEL_ID,
+				{
+					'type': 'userlist',
+					'action': 'PART',
+					'username': user.username,
+				}
+			)
+	for user in models.User.objects(
```
mitar: Because we do not have `elif` anymore, there could be a "flicker" for some users which just happen to get online between first for loop and second. In first we would notify all about them going offline and in the second we would notify all the they came online. Let's turn those two for loops around, first for switching to `True`, then to `False`.

piplmesh/frontend/tasks.py
mitar: Use spaces, not tabs!

mitar:
So otherwise good!

One more thing, if you are changing this, could you change `id` to `pk` in queries? `pk` is more robust to possible changing what `id` field is in the document.

piplmesh/frontend/tasks.py
lazarm: Okay, I just had some problems with python complaining about tabs and spaces which were mixed up together, and I decided to change all spaces to tabs. Wrong way obviously :-)

piplmesh/frontend/tasks.py
```diff
@@ -13,32 +13,40 @@
...
 @task.task
 def check_online_users():
-    # TODO: Iterating over all users in Python could become really slow once there are millions of users, much better is to limit the query only to potentially interesting users (conditions are already bellow)
-    for user in models.User.objects():
+    for user in models.User.objects(
+        is_online=False,
+        connections__ne=[],
+    ):
         if models.User.objects(
             id=user.id,
```
mitar: `id` to `pk` please.

piplmesh/frontend/tasks.py
```diff
@@ -13,32 +13,40 @@
...
-            connections=[],
-            connection_last_unsubscribe__lt=datetime.datetime.now() - datetime.timedelta(seconds=CHECK_ONLINE_USERS_RECONNECT_TIMEOUT),
-        ).update(set__is_online=False):
+            is_online=False,
+            connections__ne=[],
+        ).update(set__is_online=True):
             updates.send_update(
                 views.HOME_CHANNEL_ID,
-                {
+                    {
```
mitar: Too much indentation.

piplmesh/frontend/tasks.py
```diff
@@ -13,32 +13,40 @@
...
                     'type': 'userlist',
-                    'action': 'PART',
+                    'action': 'JOIN',
                     'username': user.username,
-                }
+                    }
```
mitar: Too much indentation.

piplmesh/frontend/tasks.py
lazarm: Sorry, I didn't notice your last comment :-)

piplmesh/frontend/tasks.py
```diff
@@ -13,32 +13,40 @@
...
+        is_online=False,
+        connections__ne=[],
+    ):
         if models.User.objects(
-            id=user.id,
-            is_online=True,
-            connections=[],
-            connection_last_unsubscribe__lt=datetime.datetime.now() - datetime.timedelta(seconds=CHECK_ONLINE_USERS_RECONNECT_TIMEOUT),
-        ).update(set__is_online=False):
+            pk=user.id,
```
mitar: And `user.pk`? :-)

lazarm: Damn it :D

Issue #141, opened 2012-04-19
Design list of online users


mitar:
Current design list of online users could be improved. For example, small (20x20) user images should be displayed next to the usernames, each entry in the list could be of a fixed width and length, with some scrollbar on the right. Whole list should be sorted, or even sortable. Maybe even searchable/filterable (on top a input box when you enter something that only those entries are displayed). Also some additional information could be displayed bellow the username (second text line, both lines are one user entry), like link to the profile page of this user.

Probably correct way to do this is to keep the list of online uses as a list of user objects in JavaScript, keep it current with updates, and simply just render this list on updates into a placeholder.

Blocked by: #63

premik91:
I know this is to early, but would something like this be okay(this is just design)?
http://shrani.si/f/3x/Gw/3KRkNgmi/userlist.png

mitar:
Why too early?

I would put it in some block somewhere. But yes, something like this.

premik91:
A simple design for list of online users. I also added the toggle animation for hidding it and the title for this 'box' (Active Users or Logged in Users doesn't realy sound good, but that's the least of my worries :)). If you think it's okay, I can make a pull request. A list item should be like:
< img src="#" alt="image" >
user2
< div class="userInfo">information about user< /div >

https://github.com/premik91/PiplMesh/compare/master...Issue-%23141

http://shrani.si/f/11/QQ/1BM3J6M1/loggedinusers.png

mitar:
Where is pull request for this?

mitar:
Please keep objects in JavaScript, as described in ticket description.

PR #157, opened 2012-05-09
Issue #141 (team 5)

premik91:
A simple design for list of online users. I also added the toggle animation for hidding it and the title for this 'box' (Active Users or Logged in Users doesn't realy sound good, but that's the least of my worries :)). If you think it's okay, I can make a pull request.

Also added the ability to display only users that contain the searched string and to arrange list of logged in users by alphabet(design needs to be improved). Also the true image of the user should be passed in 'data' to javascript. We (I and igypigi->did most of the javascript) were also unable to test this completly, because the updating logged in users doesn't always work(on our computers) and we weren't able to figure out why.

PR #159, opened 2012-05-13
Issue #141 - team 5

premik91:
A simple design for list of online users. I also added the toggle animation for hidding it and the title for this 'box' (Active Users or Logged in Users doesn't really sound good, but that's the least of my worries :)).

Also added the ability to display only users that contain the searched string and to arrange list of logged in users by alphabet(design needs to be improved). Also the true image of the user should be passed in 'data' to javascript. We (I and igypigi->did most of the javascript) were also unable to test this completely, because the updating logged in users doesn't always work(on our computers) and we weren't able to figure out why.


premik91:
Old thread: https://github.com/wlanslovenija/PiplMesh/pull/157

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,67 @@
```
mitar, line 0: For JavaScript, I recommend you to run some JavaScript lint tool over it. Will tell you even more JavaScript recommendations than me. :-)

piplmesh/frontend/tasks.py
```diff
@@ -1,11 +1,7 @@
 import datetime
-
```
mitar, line 2: Why? removing those empty lines? We have them to separate import sections (system modules, django modules, library modules, our modules).

piplmesh/frontend/tasks.py
```diff
@@ -23,11 +19,14 @@ def check_online_users():
         ).update(set__is_online=True):
             updates.send_update(
                 views.HOME_CHANNEL_ID,
-                {
+                    {
```
mitar: Why changing indentation?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,67 @@
...
+var users = new Array();
+var searchUsers;
+var alphabeticalOrder = false;
+var userList;
+
+function updateUserList(data) {
+    if (data.action == 'JOIN') {
+        users[users.length] = data;
+        if (alphabeticalOrder) {
+            rebuildUserList();
```
mitar: You didn't yet change JavaScript in the way I commented. Will wait for that.

piplmesh/frontend/tasks.py
premik91, line 2: Sorry I didn't know we were doing it that way, thought it was a mistake.

piplmesh/frontend/tasks.py
premik91: I didn't, probably Pycharm did, I will fix it.

piplmesh/frontend/static/piplmesh/js/home.js
premik91: Yeah, I told igypigi to fix it(since he wrote most of it, he should fix it).

premik91:
I think the rest is up to igypigi :)

piplmesh/frontend/tasks.py
mitar, line 2: Then you could check in other files. :-)

premik91, line 2: That's the funny thing. I did but I probably opened those that aren't seperated :)

mitar, line 2: We have those? Where?

premik91, line 2: No I am not saying they are wrong. I just randomly opened some .py files to see if all imports have one empty line between them. And those that I opened were probably all from the same section. I remeber opening signals.py and some others.

mitar:
Please update from main repository. Now list should work as expected.

mitar:
So you should be able to test everything.

mitar:
So, does it work now for you?

premik91:
Yeah it does. Excelent! :) 
I will wait for igypigi one more day, than I am gonna do it.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -346,4 +346,54 @@ input.gsc-search-button {
...
+#userlist .userInfo {
+    font-size: 8px;
+    color: #333333;
+}
+
+#userlist a:active, #userlist a:hover {
+    color: #006600;
+}
+
+/* BOXES */
```
mitar: Style this comment like others. Renamed boxes to panels.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -346,4 +346,54 @@ input.gsc-search-button {
...
+.box {
+    width: 260px;
+}
+
+.boxHeader {
```
mitar: `.boxHeader` -> `.panel .header`?

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,17 @@
...
+    <div class="boxHeader" title="{% trans "Show/Hide Users" %}">{% trans "Active Users" %}</div>
     <ul id="userlist">
         {% for user in online_users %}
             <li>{{ user }}</li>
         {% endfor %}
     </ul>
+    <form action="">
+        <img src="{{ STATIC_URL }}piplmesh/images/magnifying_glass.png" alt="magnifying_glass" />
+        <input type="text" id="search_users" />
+        <label>
```
mitar: Style this label and input same as we have style elsewhere. (for attribute for example).

mitar:
It would be great if it would be done until tomorrow so that others can see it. Can you call him?

premik91:
He had to take his dog out, and than he is going to sleep.
He says he is gonna do it between the lecture. :)

mitar:
I will be beaten by the professor if you will be doing things during the lecture.

premik91:
We will be discreet :)

piplmesh/frontend/static/piplmesh/css/style.css
premik91: Fixed

piplmesh/frontend/static/piplmesh/css/style.css
premik91: Fixed

piplmesh/frontend/templates/userlist.html
premik91: Fixed

piplmesh/frontend/tasks.py
igypigi, line 2: {% for user in online_users %}
            <li>{{ user }}</li>
        {% endfor %}
why do we need this? how do I update userlist?

mitar, line 2: This was for initial display. You should move this to JavaScript now.

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,17 @@
...
         {% for user in online_users %}
             <li>{{ user }}</li>
         {% endfor %}
     </ul>
+    <form action="">
+        <img src="{{ STATIC_URL }}piplmesh/images/magnifying_glass.png" alt="magnifying_glass" />
+        <input type="text" id="search_users" />
+        <label for="alphabet_order">
+            <input type="checkbox" id="alphabet_order" />A-Z
+        </label>
```
mitar: We have like:

```
<input type="checkbox" id="alphabet_order" /><label for="alphabet_order">A-Z</label>
```

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,17 @@
-<div>
+{% load i18n %}
+
+<div class="panel">
+    <div class="header" title="{% trans "Show/Hide Users" %}">{% trans "Active Users" %}</div>
     <ul id="userlist">
```
mitar: You should move this to JavaScript, some global variable. And initialize from there.

piplmesh/frontend/tasks.py
```diff
@@ -27,6 +27,9 @@ def check_online_users():
                     'type': 'userlist',
                     'action': 'JOIN',
                     'username': user.username,
+                    # TODO get user image
+                    'image': '',
+                    'id': user.id,
```
mitar: Why we need user ID exposed?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
-		$('<li/>').text(data.username).appendTo('#userlist');
-	}
-	else if (data.action == 'PART') {
-		// TODO: Improve escape and use more suitable :contains method
-		$('#userlist li:contains(' + escape(data.username) + ')').remove();
-	}
+// TODO add users already online to this array
+var users = new Array();
+var searchUsers;
+var alphabeticalOrder = false;
```
mitar: Why would we have non-alphabetical order? I would remove this, and always sort.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+var userList;
+function updateUserList(data) {
+    if (data.action == 'JOIN') {
+        users[users.length] = data;
+        redrawUserList();
+    } else if (data.action == 'PART') {
+        var index = users.indexOf(data);
```
mitar: Use jQuery.grep to remove the user from the array.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
-		$('#userlist li:contains(' + escape(data.username) + ')').remove();
-	}
+// TODO add users already online to this array
+var users = new Array();
+var searchUsers;
+var alphabeticalOrder = false;
+var userList;
+function updateUserList(data) {
+    if (data.action == 'JOIN') {
+        users[users.length] = data;
```
mitar: I like `users.push(data)` more.

BTW, you have to check if user is already in there.

mitar: Create User class:

```
function User(data) {
    var self = this;

    $.extend(self, data);

    self.someMethod = function () {
        ...
    };
}
```

And then you do:

```
var user = new User(data.user);
onlineUsers[user.username] = user;
```

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
-function updateUserlist(data) {
-	if (data.action == 'JOIN') {
-		$('<li/>').text(data.username).appendTo('#userlist');
-	}
-	else if (data.action == 'PART') {
-		// TODO: Improve escape and use more suitable :contains method
-		$('#userlist li:contains(' + escape(data.username) + ')').remove();
-	}
+// TODO add users already online to this array
+var users = new Array();
```
mitar: I would have a dictionary here:

```
var onlineUsers = {};
```

piplmesh/frontend/tasks.py
```diff
@@ -27,6 +27,9 @@ def check_online_users():
                     'type': 'userlist',
                     'action': 'JOIN',
```
mitar, line 29: Change this structure to:

```
{
    'type': 'userlist',
    'action': 'JOIN',
    'user': {
        'username': user.username,
        'profile_url': urlresolvers.reverse('user', kwargs={'username': user.username}),
        'image_url': '',
    },
}
```

piplmesh/frontend/static/piplmesh/js/home.js
igypigi: premik91 said users can be ordered by both time of login or alphabet.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+        if (index != -1) {
+            users.splice(index, 1);
+            redrawUserList();
+        }
+    }
+}
+
+// redraws userList
+function redrawUserList() {
+    var tmpUsers = users.slice();
```
mitar:
```
var users = [];
for (var user in onlineUsers) {
    if (onlineUsers.hasOwnProperty(user)) {
        users.push(user);
    }
}
users.sort(function (user1, user) {
    if (user1.username.toLowerCase() < user2.username.toLowerCase()) return -1;
    if (user1.username.toUpperCase() > user2.username.toUpperCase()) return 1;
    return 0;
}
```

piplmesh/frontend/static/piplmesh/js/home.js
mitar: Or, if we will have dictionary, you can easier remove it.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: Simplify. I would use only by alphabet.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+    if (alphabeticalOrder) {
+        tmpUsers.sort(function (user1, user2) {
+            if (user1.username.toUpperCase() < user2.username.toUpperCase()) return -1;
+            if (user1.username.toUpperCase() > user2.username.toUpperCase()) return 1;
+            return 0;
+        });
+    }
+    for (user in tmpUsers) {
+        if (user.username.indexOf(searchUsers) != -1) {
+            userlist.append("<li>" +
```
mitar: `userList` != `userlist`

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+                "<div class=\"userInfo\">" + user.info + "</div>" +
+                "</li>");
+        }
+    }
 }
 
 $(document).ready(function () {
-	$.updates.registerProcessor('home_channel', 'userlist', updateUserlist);
-});
+    userList = $("#userlist");
```
mitar: I am not sure if we need to store this. Because query by ID is very fast.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+            users.splice(index, 1);
+            redrawUserList();
+        }
+    }
+}
+
+// redraws userList
+function redrawUserList() {
+    var tmpUsers = users.slice();
+    userList.empty();
```
mitar: Move this lower.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: Don't concatenate strings, but use jQuery to build it (use `'`, not `"`):

```
var image = $('<img/>').attrs({'src': user.image_url, 'alt': _("User image")});
$('<li/>').append(image);
```

and so on ...

mitar:
OK. Enough for now. Include this your template somewhere in `home.html`. Pull from main repository new changes.

piplmesh/frontend/tasks.py
premik91: We don't :)

piplmesh/frontend/static/piplmesh/js/home.js
premik91: So we don't need the checkbox?

premik91:
So I should remove userlist.html and move the code to home.html?

premik91:
Should I make panel reusable like this? 

.userPanel {
    width: 260px;
}

.panelHeader {
    color: #FCFCFC;
    height: 20px;
    text-align: center;
    background-color: #68AE2C;
    cursor: pointer;
}

---

< div class="userPanel" >
    < div class="panelHeader" title="{% trans "Show/Hide Users" %}">{% trans "Active Users" %}< / div>

piplmesh/frontend/static/piplmesh/js/home.js
mitar: We do not.

mitar:
No, it is OK that it is in `userlist.html`. Just include `userlist.html` in `home.html`.

mitar:
For now leave panels as they are. We will generalize them later. (There is ticket for that.)

premik91:
It already is included in home.html

mitar:
Aha. OK.

mitar:
And the rest? JavaScript?

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,14 @@
...
         {% for user in online_users %}
             <li>{{ user }}</li>
         {% endfor %}
     </ul>
+    <form action="">
+        <img src="{{ STATIC_URL }}piplmesh/images/magnifying_glass.png" alt="magnifying_glass" />
```
mitar: Alt should be some translatable string ("Search users" maybe?).

piplmesh/frontend/tasks.py
```diff
@@ -27,7 +27,12 @@ def check_online_users():
                 {
                     'type': 'userlist',
                     'action': 'JOIN',
-                    'username': user.username,
+                    'user': {
+                        'username': user.username,
+                        'profile_url': urlresolvers.reverse('user', kwargs={'username': user.username}),
+                        # TODO get user image
```
mitar: Format this TODO into format of other:

```
# TODO: Get real user image
```

piplmesh/frontend/tasks.py
```diff
@@ -27,7 +27,12 @@ def check_online_users():
...
+                        'image_url': '',
```
mitar: For now you can use our default user image we already have. So that there will be something. Just hard-code it (but still build image path dynamically).

igypigi:
i'm working on JavaScript right now.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -400,3 +399,52 @@ input.gsc-search-button {
...
+    margin-bottom: 10px;
+}
+
+#userlist img {
+    height: 25px;
+    width: 25px;
+    float: left;
+}
+
+#userlist .userInfo {
```
mitar: _userInfo_ -> _info_?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -400,3 +399,52 @@ input.gsc-search-button {
...
+}
+
+#userlist li {
+    display: block;
+    margin-bottom: 10px;
+}
+
+#userlist img {
+    height: 25px;
+    width: 25px;
```
mitar, line 507: I think ticket is saying that user images should be 20x20 px? Is that too small? I don't think so.

mitar:
Good. You understand what I want from you in JavaScript, yes?

igypigi:
i think so. but i dont know how to move django/python code for initial display of users already online to javascript?

mitar:
The point is that you are currently redrawing the list of users on every update. And that you begin with an empty list (or dict) of users in JavaScript. But why wouldn't you start with a filled one, filled with currently known online users. And then just redraw on document ready. And then keep it updated on pushed updates.

So how can you get list of currently known online users into a JavaScript variable:
- you create a custom Django view, which dynamically create JavaScript code which defines this variable and load it through HTML script tag
- you output such code in home template itself, similar to how we are doing [it here](https://github.com/wlanslovenija/nodewatcher/blob/master/nodewatcher/frontend/templates/nodes/map.html#L90)

Heh, we are even doing this [already in PiplMesh](https://github.com/wlanslovenija/PiplMesh/blob/master/piplmesh/frontend/templates/base.html#L14).

mitar:
Once you have data in JavaScript variable - you know how to draw this into HTML, no? :-)

igypigi:
i will try that. thanks :)

piplmesh/frontend/static/piplmesh/css/style.css
premik91, line 507: I thought they were, that's why I made them 25x25. I thought they were much better and clearer this size. I thought they looked good and just the correct size? http://shrani.najdi.si/?3x/Gw/3KRkNgmi/userlist.png

mitar, line 507: Hm. In fact Facebook has 28x28 in chat. And the one in header menu is 23x23. I do not know where I found 20x20. But they have only 50x50 resized. The rest they resize with CSS. So what should we have?

piplmesh/frontend/tasks.py
premik91: Fixed

piplmesh/frontend/tasks.py
premik91: You mean like I had it before? Like 

'image_url': '../images/logo.png'

mitar: No, you use `django.contrib.staticfiles.storage.staticfiles_storage.url(path)`.

piplmesh/frontend/static/piplmesh/css/style.css
premik91, line 507: You tell me :). At the begining I had them 20x20, but they seemed to small and the addtitional user info didn't look so good next to the image. That was also one of the reasons I made them bigger.

piplmesh/frontend/templates/userlist.html
premik91: Fixed

piplmesh/frontend/tasks.py
premik91: I was just searching for something like that, thx

premik91:
As I have read javascript isn't supported yet by Django(but that could be old documentation)? So the way igypigi did is wrong?

premik91:
Nope doesn't work. I will inform him.

premik91:
I've also noticed that when you log in, the text "HOME PAGE CONTENT HERE " changes style. But this should be in another Issue.

mitar:
> As I have read javascript isn't supported yet by Django

What you mean by that? Supported for what?

premik91:
I meant using django code in javascript?

mitar:
What means "django code in JavaScript"?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,68 @@
-function updateUserlist(data) {
-	if (data.action == 'JOIN') {
-		$('<li/>').text(data.username).appendTo('#userlist');
-	}
-	else if (data.action == 'PART') {
-		// TODO: Improve escape and use more suitable :contains method
-		$('#userlist li:contains(' + escape(data.username) + ')').remove();
-	}
+var onlineUsers = {
```
mitar: Yes, this is wrong. This should go to Django templates. JavaScript files are static files and are not processed by Django. They could be. But we are not doing that with this file.

You can:
- move this to `home.html` template
- create a view returning this JavaScript code
- you can do the later through templates, if you want

premik91:
I was talking about this.
var onlineUsers = {
    {% for user in online_users %}
        {{ user.username }}: {{ user }} {% if not forloop.last %}, {% endif %}
    {% endfor %}
};
It is as simple as that, does this work or not. This is what I said was "Django code".

mitar:
Yes, I commented on this in the code.

premik91:
Yes I saw, we were probably commenting at the same time :)

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,68 @@
...
+};
+var searchUsers = "";
+
+function User(data) {
+    var self = this;
+    $.extend(self, data.user);
+}
+
+function redrawUserList() {
+    var keys = [];
```
mitar, line 8: `usernames` maybe?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,68 @@
...
+        if (key1.toUpperCase() < key2.toUpperCase()) return -1;
+        if (key1.toUpperCase() > key2.toUpperCase()) return 1;
+        return 0;
+    });
+    $("#userlist").empty();
+    for (key in keys) {
+        if (searchUsers === "" || key.indexOf(searchUsers) != -1) {
+            user = onlineUsers[key];
+            var li = $('<li/>');
+            var image = $('<img/>').attrs({'src':user.image_url, 'alt':'User image'});
```
mitar: Translatable string. [Check this](https://docs.djangoproject.com/en/dev/topics/i18n/translation/#internationalization-in-javascript-code).

mitar: Space after `:`.

mitar: You can also wrap dict in a similar way we do it in Python.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,68 @@
...
+            li.append(image);
+            li.append(user.username);
+            var div = $('<div/>').attrs({'class':'info'});
```
mitar: Space.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,68 @@
...
+            li.append(div);
+            $("#userlist").append(li);
+        }
+    }
+}
+
+function updateUserList(data) {
+    var user = new User(data.user);
+    if (data.action === 'JOIN') {
+        if (!onlineUsers.hasOwnProperty(user.username)) {
```
mitar: No, don't check. Maybe the new update has some new information (new user image, for example)?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,68 @@
...
+            onlineUsers[user.username] = user;
+            redrawUserList();
+        }
+    } else if (data.action === 'PART') {
+        var index = onlineUsers.indexOf(data);
+        if (onlineUsers.hasOwnProperty(user.username)) {
```
mitar: `onlineUsers[user.username]` is good enough.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,68 @@
...
+        }
+    }
 }
 
 $(document).ready(function () {
-	$.updates.registerProcessor('home_channel', 'userlist', updateUserlist);
+    $.updates.registerProcessor('home_channel', 'userlist', updateUserList);
+
+    $(".panel .header").click(function () {
+        $(this).next("ul").slideToggle("slow");
```
mitar: Do it fast.

Please use `'` everywhere for constant strings.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,68 @@
...
+    });
+
+    $("#search_users").keyup(function () {
```
mitar: `keyup`? Not `change`? What if I paste something in? Doe sit still work?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,68 @@
...
+        searchUsers = $(this).val();
```
mitar: I don't see the reason to use one more global variable. Why we cannot simply `$('#search_users').val()` in the code above when we need its content?

piplmesh/frontend/tasks.py
```diff
@@ -27,7 +27,12 @@ def check_online_users():
                 {
                     'type': 'userlist',
                     'action': 'JOIN',
-                    'username': user.username,
+                    'user': {
+                        'username': user.username,
+                        'profile_url': urlresolvers.reverse('user', kwargs={'username': user.username}),
+                        # TODO: Get real user image
+                        'image_url': django.contrib.staticfiles.storage.staticfiles_storage.url('../images/logo.png'),
+                    },
```
mitar, line 34: Of course you should import this in a way to have a shorted function call!

mitar, line 34: And no, it is not correct to use `../images/logo.png`. Do we use such path in Django templates for static files?

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,10 @@
-<div>
-    <ul id="userlist">
-        {% for user in online_users %}
-            <li>{{ user }}</li>
-        {% endfor %}
-    </ul>
+{% load i18n %}
+
+<div class="panel">
+    <div class="header" title="{% trans "Show/Hide Users" %}">{% trans "Active Users" %}</div>
```
mitar: _Active_ -> _Online_

mitar:
Waiting for JavaScript ...

mitar:
BTW, fixes from igypigi should be committed by him. Otherwise he will not get any credit for his work towards his course mark.

piplmesh/frontend/static/piplmesh/css/style.css
mitar, line 507: OK, so you are saying 25x25 is OK, then use it.

premik91, line 507: Yes I think it is okay, but that can easily be changed if necessary

piplmesh/frontend/tasks.py
premik91, line 34: Is path now okay? Or should it be just piplmesh/images/logo.png

piplmesh/frontend/tasks.py
```diff
@@ -27,7 +29,12 @@ def check_online_users():
                 {
                     'type': 'userlist',
                     'action': 'JOIN',
-                    'username': user.username,
+                    'user': {
+                        'username': user.username,
+                        'profile_url': urlresolvers.reverse('user', kwargs={'username': user.username}),
+                        # TODO: Get real user image
+                        'image_url': django.contrib.staticfiles.storage.staticfiles_storage.url('/piplmesh/images/logo.png'),
```
mitar: Again, why full import of the function? And please, just try what is the output of your call and see if it is right.

premik91: Now it all works and gives the expected output.

piplmesh/frontend/tasks.py
```diff
@@ -27,7 +29,12 @@ def check_online_users():
                 {
                     'type': 'userlist',
                     'action': 'JOIN',
-                    'username': user.username,
+                    'user': {
+                        'username': user.username,
+                        'profile_url': urlresolvers.reverse('user', kwargs={'username': user.username}),
+                        # TODO: Get real user image
+                        'image_url': staticfiles_storage.url('/static/piplmesh/images/logo.png'),
```
mitar: No, it is still not right. You should not hard-code final URL in. Then you have not done anything. :-)

premik91: haha maybe it would be faster to do TODO :D. 
So I should just leave 'piplmesh/images/logo.png' and add {{ STATIC_URL }} in js?

mitar: Nooo. First, template tags do not work in JS. Because of that we will construct URL in Python.

In fact, concatenating STATIC_URL like that is not the best way to do it. We should use `static` template tag. Please update also templates.

[Read here more](https://docs.djangoproject.com/en/dev/howto/static-files/#staticfiles-in-templates) on the subject.

And then when you will know how to do that in templates, you can call the same function which is used in `static` template tag to get also URL in the code.

premik91:
Is this what you meant? I read all about static files, but about staticfiles_storage there are very little examples. I also googled it and since this works I don't know what else to do. I can't say what igypigi is waiting for...

mitar:
Yes. Also do the same in templates. So instead of using that variable, use tag. (Import it from `staticfiles` library.)

premik91:
I don't understand what to do. In templates tag is already used.
<code>< img src="{{ STATIC_URL }}piplmesh/images/magnifying_glass.png" /></code>

mitar:
Load `staticfiles` library and use `static` template tag. That is not template tag, but variable. We are concatenating strings manually, it is better through tag.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: And this?

piplmesh/frontend/tasks.py
premik91: Is this okay now? Static tag works. Should I do this for all templates?

piplmesh/frontend/static/piplmesh/js/home.js
premik91: Much done, but still some fixes necessary. Should we move all functions inside the document.ready function?
<code>
    $.updates.registerProcessor('home_channel', 'userlist', updateUserList);
    /\* ------------------ USER PANEL ---------------------------- */
    var searchUsers = '';

```
function User(data) {
    var self = this;
    $.extend(self, data.user);
}
```

</code>

piplmesh/account/models.py
```diff
@@ -46,3 +48,7 @@ class User(auth.User):
     connections = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Connection))
     connection_last_unsubscribe = mongoengine.DateTimeField()
     is_online = mongoengine.BooleanField(default=False)
+
+    profile_url = urlresolvers.reverse('user', kwargs={'username': user.username})
+    # TODO: Get real user image
+    image_url = staticfiles_storage.url('piplmesh/images/logo.png')
```
mitar: Change this into two methods on User class, `get_profile_url` and `get_image_url`.

piplmesh/account/models.py
```diff
@@ -46,3 +48,7 @@ class User(auth.User):
     connections = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Connection))
     connection_last_unsubscribe = mongoengine.DateTimeField()
     is_online = mongoengine.BooleanField(default=False)
+
+    profile_url = urlresolvers.reverse('user', kwargs={'username': user.username})
```
mitar: `user` variable is defined here??? Move to method and use `self`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
-	}
-	else if (data.action == 'PART') {
-		// TODO: Improve escape and use more suitable :contains method
-		$('#userlist li:contains(' + escape(data.username) + ')').remove();
-	}
+var searchUsers = '';
+
+function User(data) {
+    var self = this;
+    $.extend(self, data.user);
```
mitar: Ups. Use `data` here. And pass into constructor `data.user`. It is much better (now that we have a separate dict only for user fields).

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+            var div = $('<div/>').attrs({'class': 'info'});
+            div.append(user.info);
+            li.append(div);
+            $('#userlist').append(li);
+        }
+    }
+}
+
+function updateUserList(data) {
+    var user = new User(data.user);
```
mitar, line 36: Oh, you are already passing `data.user` here. :-)

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,28 @@
...
+                        'image_url': '{{ user.image_url }}'
+                    }
+                    {% if not forloop.last %}, {% endif %}
+                {% endfor %}
+            ];
+            /* ]]> */
+        </script>
+{% endblock %}
+
+{% load staticfiles %}
```
mitar: Move all loading to one template tag (you can list multiple libraries to load) on the top.

piplmesh/frontend/tasks.py
mitar: Yes.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: No, have objects outside and then methods inside those objects. For now leave as it is. We will restructure code when we will have more of it.

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,28 @@
-<div>
-    <ul id="userlist">
-        {% for user in online_users %}
-            <li>{{ user }}</li>
-        {% endfor %}
-    </ul>
+{% load i18n %}
+
+{% block js %}
```
mitar: Good. But also first add parent content `{{ block.super }}`, because maybe some other template has already used this block, to not lose its JavaScript code.

piplmesh/account/models.py
premik91: Ups. Fixed :)

piplmesh/frontend/templates/userlist.html
mitar: See documentation on `block` tag for more information. Or documentation on template inheritance.

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,28 @@
...
+        <script type="text/javascript">
+            /* <![CDATA[ */
+            var onlineUsers = [
+                {% for user in online_users %}
+                    {
+                        'username': '{{ user.username }}',
+                        'profile_url': '{{ user.profile_url }}',
+                        'image_url': '{{ user.image_url }}'
+                    }
+                    {% if not forloop.last %}, {% endif %}
```
mitar: Unnecessary space after comma.

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,28 @@
-<div>
-    <ul id="userlist">
-        {% for user in online_users %}
-            <li>{{ user }}</li>
-        {% endfor %}
-    </ul>
+{% load i18n %}
+
+{% block js %}
+        <script type="text/javascript">
```
mitar: Indentation (too much).

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,28 @@
...
+            /* <![CDATA[ */
+            var onlineUsers = [
+                {% for user in online_users %}
+                    {
+                        'username': '{{ user.username }}',
```
mitar: You should always escape values. Django template variables are escaped automatically only for HTML. Not for JavaScript. There is a template filter for this.

piplmesh/frontend/tasks.py
```diff
@@ -7,6 +7,7 @@
 from pushserver.utils import updates
 
 from piplmesh.account import models
+
```
mitar: No.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+        if (key1.toUpperCase() < key2.toUpperCase()) return -1;
+        if (key1.toUpperCase() > key2.toUpperCase()) return 1;
+        return 0;
+    });
+    $('#userlist').empty();
+    for (key in keys) {
+        if (searchUsers === '' || key.indexOf(searchUsers) != -1) {
+            user = onlineUsers[key];
+            var li = $('<li/>');
+            var image = $('<img/>').attrs({'src': user.image_url, 'alt': 'User image'});
```
mitar: Translatable string. Check how translation is done in JavaScript.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+    }
+}
+
+function updateUserList(data) {
+    var user = new User(data.user);
+    if (data.action === 'JOIN') {
+        onlineUsers[user.username] = user;
+        redrawUserList();
+    } else if (data.action === 'PART') {
+        var index = onlineUsers.indexOf(data);
```
mitar: No need for this?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
 $(document).ready(function () {
-	$.updates.registerProcessor('home_channel', 'userlist', updateUserlist);
+    $.updates.registerProcessor('home_channel', 'userlist', updateUserList);
+
+    $('.panel .header').click(function () {
+        $(this).next('ul').slideToggle('slow');
+    });
+
+    $('#search_users').change(function () {
+        searchUsers = $(this).val();
```
mitar: Why do we have this variable? Why not simply call `$('#search_users').val()` in `redrawUserList`?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+        }
+    }
+    keys.sort(function (key1, key2) {
+        if (key1.toUpperCase() < key2.toUpperCase()) return -1;
+        if (key1.toUpperCase() > key2.toUpperCase()) return 1;
+        return 0;
+    });
+    $('#userlist').empty();
+    for (key in keys) {
+        if (searchUsers === '' || key.indexOf(searchUsers) != -1) {
```
mitar: This should be probably case-insensitive search. Fix or add TODO.

piplmesh/frontend/templates/userlist.html
premik91: Fixed

piplmesh/frontend/templates/userlist.html
premik91: Fixed

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+var searchUsers = '';
+
+function User(data) {
+    var self = this;
+    $.extend(self, data.user);
+}
+
+function redrawUserList() {
+    var keys = [];
+    for (key in onlineUsers) {
```
mitar: OK, use [$.each](http://api.jquery.com/jQuery.each/) here.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+            keys.push(key);
+        }
+    }
+    keys.sort(function (key1, key2) {
+        if (key1.toUpperCase() < key2.toUpperCase()) return -1;
+        if (key1.toUpperCase() > key2.toUpperCase()) return 1;
+        return 0;
+    });
+    $('#userlist').empty();
+    for (key in keys) {
```
mitar: OK, use [$.each](http://api.jquery.com/jQuery.each/) here.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,61 @@
...
+
+function User(data) {
+    var self = this;
+    $.extend(self, data.user);
+}
+
+function redrawUserList() {
+    var keys = [];
+    for (key in onlineUsers) {
+        if (onlineUsers.hasOwnProperty(key)) {
```
mitar: No need for this then.

piplmesh/frontend/tasks.py
premik91: Fixed

piplmesh/frontend/templates/userlist.html
premik91: <code>{{ user.username|escapejs }}</code>
Like this?

mitar: Yes!

piplmesh/frontend/templates/userlist.html
premik91: Fixed

premik91: I've read the documentation and tried block.super, but it doesn't work. <code>Error: 'Caught AttributeError while rendering: 'BlockNode' object has no attribute 'context' '</code>
Than I googled the problem and found this and I think it makes sense: 
http://stackoverflow.com/questions/6566347/django-templates-using-block-super-in-included-template-fails-exception

piplmesh/frontend/tasks.py
premik91: I am having problems here, because if I change this
<code>"{{ STATIC_URL }}piplmesh/images/logo-{{ LANGUAGE_CODE }}.png"</code>
into this
<code>"{% static "piplmesh/images/logo-{{ LANGUAGE_CODE }}.png" %}"</code>,
the _{{ LANGUAGE_CODE }}_ variable does't work. Should I use js for this(I think that might work)?

piplmesh/account/models.py
```diff
@@ -7,6 +7,10 @@
 
 from piplmesh.account import fields
 
+from django.contrib.staticfiles.storage import staticfiles_storage
+
```
mitar: No empty line between imports from same package.

piplmesh/frontend/templates/userlist.html
```diff
@@ -1,7 +1,27 @@
...
+    var onlineUsers = [
         {% for user in online_users %}
-            <li>{{ user }}</li>
+            {
+                'username': '{{ user.username|escapejs }}',
+                'profile_url': '{{ user.get_profile_url|escapejs }}',
+                'image_url': '{{ user.get_image_url|escapejs }}'
+            }
+            {% if not forloop.last %},{% endif %}
         {% endfor %}
```
mitar: Move this line up, to the end of previous line.

piplmesh/frontend/templates/userlist.html
mitar: Then move the whole this `js` block to `home.html` template.

piplmesh/frontend/tasks.py
mitar: Of course not. Why use JS?

Anyway, you will have to add new context variable in our context processor and construct the filepath in Python. And use that then in template. Read second paragraph:

https://docs.djangoproject.com/en/dev/howto/static-files/#with-a-template-tag

premik91: It works but you will certainly hate the way I added strings. :)

premik91:
Waiting for approval :)

docs/translation.rst
```diff
@@ -19,14 +20,16 @@ sure your Django applications you want translated have a directory called
 ``locale`` (if not, you should create one).
 
 When all string are marked for translation and settings are prepared run
-command::
+command(the second one if you need JavaScript files translated)::
```
mitar: No need for this comment, because we want to translate everything, of course.

docs/translation.rst
```diff
@@ -19,14 +20,16 @@ sure your Django applications you want translated have a directory called
...
 
     ../manage.py makemessages -l <language code>
+    ../manage.py makemessages -l <language code> -d djangojs
```
mitar, line 26: Change to:

```
../manage.py makemessages -l <language code> -d djangojs
```

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,57 @@
...
-		$('#userlist li:contains(' + escape(data.username) + ')').remove();
-	}
+function User(data) {
+    var self = this;
+    $.extend(self, data);
+}
+
+function redrawUserList() {
+    var keys = [];
+    $.each(onlineUsers, function(key, value){
```
mitar: For anonymous function put spaces around `()`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,57 @@
...
+        keys.push(key);
+    });
+    keys.sort(function (key1, key2) {
```
mitar, line 12: Same as here.

piplmesh/frontend/context_processors.py
```diff
@@ -9,10 +9,13 @@ def global_vars(request):
...
 
     return {
+        # Constants
         'HOME_CHANNEL_ID': views.HOME_CHANNEL_ID,
         'LOGIN_REDIRECT_URL': settings.LOGIN_REDIRECT_URL,
         'REDIRECT_FIELD_NAME': auth.REDIRECT_FIELD_NAME,
         'SEARCH_ENGINE_UNIQUE_ID': settings.SEARCH_ENGINE_UNIQUE_ID,
 
+        # Variables
+        'logo_url': "piplmesh/images/logo-%s.png" % (settings.LANGUAGE_CODE,),
```
mitar: Nop. Don't read from `settings` (that is just default), you have to read from context.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,57 @@
...
+            li.append(div);
+            $('#userlist').append(li);
+        }
+    });
+}
+
+function updateUserList(data) {
+    var user = new User(data.user);
+    if (data.action === 'JOIN') {
+        onlineUsers[user.username.toUpperCase()] = user;
```
mitar: Good idea. But go to lower case. ;-)

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,57 @@
...
+        redrawUserList();
+    } else if (data.action === 'PART') {
+        if (onlineUsers[user.username]) {
```
mitar: But then you have to convert to same case everywhere! And this means it is easy to break.

piplmesh/frontend/static/piplmesh/js/home.js
premik91: Fixed

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,57 @@
...
-	if (data.action == 'JOIN') {
-		$('<li/>').text(data.username).appendTo('#userlist');
-	}
-	else if (data.action == 'PART') {
-		// TODO: Improve escape and use more suitable :contains method
-		$('#userlist li:contains(' + escape(data.username) + ')').remove();
-	}
+function User(data) {
+    var self = this;
+    $.extend(self, data);
```
mitar, line 3: I suggest that you add:

```
self._key = self.username.toLowerCase();
```

And then use `user._key` for keys.

docs/translation.rst
premik91: Fixed

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,57 @@
...
+    $.each(onlineUsers, function(key, value){
+        keys.push(key);
+    });
+    keys.sort(function (key1, key2) {
+        if (key1 < key2) return -1;
+        if (key1 > key2) return 1;
+        return 0;
+    });
+    $('#userlist').empty();
+    var searchUsers = $('#search_users').val().toUpperCase();
```
mitar: Replace this and second line. This goes more with the next block of code.

piplmesh/urls.py
```diff
@@ -10,6 +10,11 @@
 v1_api.register(resources.UserResource())
 v1_api.register(resources.PostResource())
 
+js_info_dict = {
+    'domain': 'djangojs',
```
mitar: This is already default.

docs/translation.rst
premik91, line 26: Fixed

piplmesh/urls.py
```diff
@@ -10,6 +10,11 @@
...
+    'packages': ('piplmesh.frontend',),
```
mitar: Add also `django.conf`.

piplmesh/urls.py
```diff
@@ -10,6 +10,11 @@
...
+    }
```
mitar: Indent.

piplmesh/urls.py
```diff
@@ -37,4 +42,6 @@
 
     # RESTful API
     url(r'^api/', include(v1_api.urls)),
+
+    (r'^jsi18n/$', 'django.views.i18n.javascript_catalog', js_info_dict),
```
mitar: And where do you use it?

piplmesh/frontend/static/piplmesh/js/home.js
premik91: Fixed

piplmesh/urls.py
premik91: Fixed?

piplmesh/urls.py
premik91: Fixed

piplmesh/urls.py
premik91: Fixed

piplmesh/frontend/templates/base.html
```diff
@@ -1,14 +1,14 @@
...
-        <link type="image/x-icon" href="{{ STATIC_URL }}piplmesh/images/favicon.ico" rel="icon" />
-        <link rel="stylesheet" href="{{ STATIC_URL }}piplmesh/css/style.css" type="text/css" media="screen" />
-        <script src="{{ STATIC_URL }}piplmesh/jquery/jquery.js" type="text/javascript"></script>
-        <script type="text/javascript" src="{{ STATIC_URL }}pushserver/updates.js"></script>
-        <script src="{{ STATIC_URL }}piplmesh/js/home.js" type="text/javascript"></script>
+        <link type="image/x-icon" href="{% static "piplmesh/images/favicon.ico" %}" rel="shortcut icon" />
+        <link type="image/x-icon" href="{% static "piplmesh/images/favicon.ico" %}" rel="icon" />
+        <link rel="stylesheet" href="{% static "piplmesh/css/style.css" %}" type="text/css" media="screen" />
+        <script src="{% static "piplmesh/jquery/jquery.js" %}" type="text/javascript"></script>
+        <script type="text/javascript" src="{% static "pushserver/updates.js" %}"></script>
```
mitar: Move `type` to the end of the tag.

piplmesh/urls.py
premik91: Fixed

piplmesh/frontend/templates/base.html
premik91: Fixed

piplmesh/frontend/context_processors.py
premik91: I don't understand what to do. You said I should put it like this :)

mitar: It is very good what you did, only the value you put into the string is wrong. `LANGUAGE_CODE` in the template context is not the same as `settings.LANGUAGE_CODE`, but depends on the current user language. `settings.LANGUAGE_CODE` is only fallback value for that value in template context.

docs/translation.rst
```diff
@@ -8,6 +8,7 @@ To make your strings translatable edit them as follows:
 
 - in Django template: ``{% trans "Some string" %}``
```
mitar: _template_ -> _templates_

piplmesh/frontend/templates/home.html
```diff
@@ -2,6 +2,23 @@
 
 {% block fulltitle %}wlan slovenija{% endblock %}
 
+{% block js %}
+{{ block.super }}
+<script type="text/javascript">
+    /* <![CDATA[ */
+    var onlineUsers = [
+    {% for user in online_users %}
```
mitar: Indentation. Every block indent.

piplmesh/urls.py
```diff
@@ -10,11 +10,16 @@
 v1_api.register(resources.UserResource())
 v1_api.register(resources.PostResource())
 
+js_info_dict = {
+        'packages': ('piplmesh.frontend', 'django.conf',),
```
mitar: Order by alphabet. Split into lines.

piplmesh/urls.py
```diff
@@ -10,11 +10,16 @@
...
+    }
```
mitar: Indentation.

piplmesh/locale/sl/LC_MESSAGES/djangojs.po
```diff
@@ -0,0 +1,22 @@
```
mitar, line 0: Compare with `django.po` and sync style/header.

piplmesh/frontend/templates/base.html
```diff
@@ -1,14 +1,14 @@
...
     <head>
         <title>{% block fulltitle %}{% block title %}{% endblock %} – wlan slovenija{% endblock %}</title>
         <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
-        <link type="image/x-icon" href="{{ STATIC_URL }}piplmesh/images/favicon.ico" rel="shortcut icon" />
-        <link type="image/x-icon" href="{{ STATIC_URL }}piplmesh/images/favicon.ico" rel="icon" />
-        <link rel="stylesheet" href="{{ STATIC_URL }}piplmesh/css/style.css" type="text/css" media="screen" />
-        <script src="{{ STATIC_URL }}piplmesh/jquery/jquery.js" type="text/javascript"></script>
-        <script type="text/javascript" src="{{ STATIC_URL }}pushserver/updates.js"></script>
-        <script src="{{ STATIC_URL }}piplmesh/js/home.js" type="text/javascript"></script>
+        <link type="image/x-icon" href="{% static "piplmesh/images/favicon.ico" %}" rel="shortcut icon" />
```
mitar: Groan. Should I really point everything out. You do not notice that also in just few tags above also things are not ordered in the same way.

premik91: Well it isn't exactly the same. And it could be just the other way around, I would've change it and you would groan why did I do something you didn't want(it is not that I didn't notice it).

mitar: It is not consistent - it is the same. I would not groan if you would make things more consistent.

premik91:
This is now it.

docs/translation.rst
```diff
@@ -21,12 +22,12 @@ sure your Django applications you want translated have a directory called
 When all string are marked for translation and settings are prepared run
 command::
 
-    ../manage.py makemessages -l <language code>
```
mitar, line 24: No, you have to run both, no? Without `-d` (for Python and template) and with `-d` (for JavaScript).

piplmesh/frontend/context_processors.py
```diff
@@ -9,10 +9,14 @@ def global_vars(request):
...
 
     return {
+        # Constants
         'HOME_CHANNEL_ID': views.HOME_CHANNEL_ID,
         'LOGIN_REDIRECT_URL': settings.LOGIN_REDIRECT_URL,
         'REDIRECT_FIELD_NAME': auth.REDIRECT_FIELD_NAME,
         'SEARCH_ENGINE_UNIQUE_ID': settings.SEARCH_ENGINE_UNIQUE_ID,
 
+        # Variables
+        'logo_url': "piplmesh/images/logo-%s.png" % (request.LANGUAGE_CODE),
```
mitar: It is better to use `translation.get_language()` to get language code. It is used also in `django.core.context_processors.i18n`, where template's `LANGUAGE_CODE` is set. So we should probably use the same. `request.LANGUAGE_CODE` is also set from `translation.get_language()`, so the value is the same, but it assumes that we always be using `django.middleware.locale.LocaleMiddleware` which sets it. So it is better to have one less assumption/dependency.

Also, put comma before last `)`, because you want to force tuple.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,58 @@
...
+        if (key1 < key2) return -1;
+        if (key1 > key2) return 1;
+        return 0;
+    });
+    $('#userlist').empty();
+
+    var searchUsers = $('#search_users').val().toLowerCase();
+    $.each(keys, function(index, key){
+        if (searchUsers === '' || key.indexOf(searchUsers) !== -1) {
+            user = onlineUsers[key];
```
mitar: `var user =`, not just `user =`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,58 @@
...
+}
+
+function updateUserList(data) {
+    var user = new User(data.user);
+    if (data.action === 'JOIN') {
+        onlineUsers[user._key] = user;
+        redrawUserList();
+    } else if (data.action === 'PART') {
+        if (onlineUsers[user._key]) {
+            delete onlineUsers[user.username.toLowerCase()];
```
mitar, line 43: Use `user._key`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,58 @@
...
+    });
+    keys.sort(function (key1, key2) {
+        if (key1 < key2) return -1;
+        if (key1 > key2) return 1;
+        return 0;
+    });
+    $('#userlist').empty();
+
+    var searchUsers = $('#search_users').val().toLowerCase();
+    $.each(keys, function(index, key){
```
mitar: Spaces around `(index, key)`. See how we have other anonymous functions.

piplmesh/frontend/templates/base.html
```diff
@@ -1,14 +1,14 @@
...
         <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
-        <link type="image/x-icon" href="{{ STATIC_URL }}piplmesh/images/favicon.ico" rel="shortcut icon" />
-        <link type="image/x-icon" href="{{ STATIC_URL }}piplmesh/images/favicon.ico" rel="icon" />
-        <link rel="stylesheet" href="{{ STATIC_URL }}piplmesh/css/style.css" type="text/css" media="screen" />
-        <script src="{{ STATIC_URL }}piplmesh/jquery/jquery.js" type="text/javascript"></script>
-        <script type="text/javascript" src="{{ STATIC_URL }}pushserver/updates.js"></script>
-        <script src="{{ STATIC_URL }}piplmesh/js/home.js" type="text/javascript"></script>
+        <link href="{% static "piplmesh/images/favicon.ico" %}" rel="shortcut icon" type="image/x-icon" />
+        <link href="{% static "piplmesh/images/favicon.ico" %}" rel="icon" type="image/x-icon" />
+        <link rel="stylesheet" href="{% static "piplmesh/css/style.css" %}" type="text/css" media="screen" />
```
mitar: Put `href` before `rel`?

piplmesh/frontend/templates/home.html
```diff
@@ -2,6 +2,23 @@
 
 {% block fulltitle %}wlan slovenija{% endblock %}
 
+{% block js %}
+{{ block.super }}
+<script type="text/javascript">
+    /* <![CDATA[ */
+    var onlineUsers = [
+        {% for user in online_users %}
+        {
```
mitar: Indent once. Do not care about output look. Care about template readability.

piplmesh/frontend/templates/home.html
```diff
@@ -2,6 +2,23 @@
 
 {% block fulltitle %}wlan slovenija{% endblock %}
 
+{% block js %}
+{{ block.super }}
```
mitar: Indent inside the block.

piplmesh/urls.py
```diff
@@ -10,11 +10,19 @@
...
+        'piplmesh.frontend',
+    ),
+}
+
 urlpatterns = patterns('',
     url('^$', frontend_views.HomeView.as_view(), name='home'),
 
     url(r'^search/', frontend_views.SearchView.as_view(), name='search'),
     url(r'^i18n/', include('django.conf.urls.i18n')),
+    (r'^jsi18n/$', 'django.views.i18n.javascript_catalog', js_info_dict),
```
mitar: Put `url` at the beginning, to be consistent.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,13 +1,58 @@
...
+        }
+    });
+}
+
+function updateUserList(data) {
+    var user = new User(data.user);
+    if (data.action === 'JOIN') {
+        onlineUsers[user._key] = user;
+        redrawUserList();
+    } else if (data.action === 'PART') {
```
mitar: `else if` into new line.

piplmesh/frontend/templates/form/form.html
```diff
@@ -9,31 +9,31 @@
```
mitar, line 8: What did you do with this file? Broke indentation? I see that you cleaned some things. But also broke some things. So we indent always. For template or HTML block/tag.

mitar, line 8: Please improve.

mitar:
It looks very good now.

docs/translation.rst
mitar, line 24: TODO.

piplmesh/frontend/templates/form/form.html
premik91, line 8: I did nothing here? It is the same as it is now?

piplmesh/frontend/context_processors.py
```diff
@@ -9,10 +9,14 @@ def global_vars(request):
...
+        'redirect_to': request.REQUEST.get(auth.REDIRECT_FIELD_NAME),
```
mitar: We have renamed this field (`request_get_next`). Please do not reintroduce it.

piplmesh/frontend/templates/home.html
```diff
@@ -2,6 +2,24 @@
...
+    <script type="text/javascript">
+        /* <![CDATA[ */
+        var onlineUsers = [
+            {% for user in online_users %}
+                {
+                    'username': '{{ user.username|escapejs }}',
+                    'profile_url': '{{ user.get_profile_url|escapejs }}',
+                    'image_url': '{{ user.get_image_url|escapejs }}'
+                }
+                {% if not forloop.last %},{% endif %}
```
mitar: You like this on its own line? You moved it again there. Then OK, let it be. ;-)

piplmesh/frontend/templates/form/form.html
mitar, line 8: Now is OK.

docs/translation.rst
premik91, line 24: Yes you have to run both, but in previous comment you said to delete the other one(I might have understood you wrong).

mitar, line 24: Yes, you need both. So have both here.

(I don't find such comment. But it does not matter.)

piplmesh/frontend/templates/home.html
igypigi: I think it is easier to understand the code and it looks better this way. :)

mitar: OK.

piplmesh/frontend/context_processors.py
mitar: TODO.

piplmesh/frontend/context_processors.py
mitar: TODO.

docs/translation.rst
mitar, line 24: TODO.

piplmesh/frontend/context_processors.py
premik91: Yes the main problem was, I didn't even suspect that igypigi will do the fixes, so I didn't sync the branch, and didn't have igypigi's fixes. And than Github crashed and so on. I hope there are no interferences.

docs/translation.rst
premik91, line 24: http://shrani.si/f/3p/zV/4F3HF7ai/f.png
I thought you meant to replace both with one.

mitar, line 24: No, you had some `-a` parameter there.

docs/translation.rst
```diff
@@ -22,11 +23,12 @@ When all string are marked for translation and settings are prepared run
 command::
 
     ../manage.py makemessages -l <language code>
+    ../manage.py makemessages -l <language code> -d djangojs
 
 *(You should replace <language code> with your preferred language code).*
 Example for Slovenian language::
 
-    ../manage.py makemessages -l sl
```
mitar, line 29: Don't remove this line.

piplmesh/frontend/context_processors.py
```diff
@@ -9,10 +10,14 @@ def global_vars(request):
...
     return {
+        # Constants
         'HOME_CHANNEL_ID': views.HOME_CHANNEL_ID,
         'LOGIN_REDIRECT_URL': settings.LOGIN_REDIRECT_URL,
         'REDIRECT_FIELD_NAME': auth.REDIRECT_FIELD_NAME,
         'SEARCH_ENGINE_UNIQUE_ID': settings.SEARCH_ENGINE_UNIQUE_ID,
 
+        # Variables
+        'logo_url': "piplmesh/images/logo-%s.png" % (translation.get_language(),),
+        'redirect_to': request.REQUEST.get(auth.REDIRECT_FIELD_NAME),
```
mitar: This variable is not needed anymore! Why are you putting it back?

piplmesh/frontend/templates/form/form.html
mitar, line 8: Aaaagh. Again it shows a lot of changes in this file. I am not sure what are you doing! ;-)

premik91, line 8: Haha, me neither I never even had it open. Will fix it back. The problem was I didn't sync igypigi's fixes, and than there was somekind of bug in Github for Mac and it crashed in the middle of commit , ...

docs/translation.rst
premik91, line 29: omg, I didn't. Arrrrgh. I was trying to revert my commit and it crahed again.

piplmesh/frontend/templates/form/form.html
premik91, line 8: It didn't even show in diff

premik91, line 8: But I see no problems still? :D I just copied the current version in, and nothing changed

mitar, line 8: I don't know what to tell you. You do see it in pull request diff?

premik91, line 8: Look you know it is not hard to make another commit to fix it. But I just copied the current version in my file, and github sees no changes. Why would I lie to you? I will do somethin else then, make a stupid commit and then another commit with current code.

mitar, line 8: I am not saying that you are lying. I am saying that also I don't know what is happening. ;-)

What does `git log` says about the file? When it was last time changed? Can you checkout that version of the file? Where have you copied current version from? How do you copy it? Maybe close your IDE. ;-)

premik91, line 8: I think now is okay. I deleted the file and made a new one. Github than recognized as a new file(?!) and was okay.

mitar:
Done. Congrats!

PR #184, opened 2012-05-21
Issue#107

cardawid:
Tests created. Your class "MongoEngineTestCase" unfortunately not working for me.


piplmesh/frontend/tests/tests.py
```diff
@@ -0,0 +1,40 @@
+from django.test.client import Client
+from django.contrib.auth.models import User
```
mitar: ???

We are not using Django User object.

mitar:
What does it mean for you that  MongoEngineTestCase does not work? Database should be cleared between each test, and if seems that you want things to stay.

PR #186, opened 2012-05-23
Minor bug fix.

MartinJakomin:
Fixed bug when login in would result in changing font colour to green.

### DIFF
--- a/piplmesh/account/templates/user/account.html
+++ b/piplmesh/account/templates/user/account.html
@@ -6,8 +6,8 @@
 
 {% block content %}
     <div id="menu_list">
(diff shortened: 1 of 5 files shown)


piplmesh/account/templates/user/account.html
```diff
@@ -16,3 +16,5 @@
         <a href="{% url password_change %}">{% trans "Change your password" %}</a></p>
     </div>
 {% endblock %}
+
+
```
mitar: What are those empty lines?

MartinJakomin: fixed

Issue #173, opened 2012-05-14
Add panel with Google Maps


mitar:
Add panel with Google Maps, with WiFi node's location on it. Display some additional information about the node in the panel (like name and link to node's webpage).

agiz:
Can you provide a list of current nodes as seen on https://nodes.wlan-si.net/network/map.

mitar:
https://nodes.wlan-si.net/feeds/rss/active/

agiz:
Is this available in JSON format?

mitar:
No? Why? XML is not good enough?

I propose you create a temporary document for nodes, a fixture with hard-coded data which we can inject into the database, and use that. So there is no need to sync from real data for now. Because the interface later on will be different anyway (we will simply read directly from the database).

agiz:
I am not sure if I understand this. You want me to create a new document in models.py with fields reflecting mentioned XML, populate the data from https://nodes.wlan-si.net/feeds/rss/active/ and then in some view, query a database for nodes and render it to end user?

mitar:
For now, I would make it very simple, imply include all nodes in JavaScript object in template. Similar to how we are doing on https://nodes.wlan-si.net/network/map/.

agiz:
Please open and specify an issue for Documents that share common fields regarding geolocation. Maybe that is meant for 
#171 (I do not think so :))? These Document(s) are needed for issues #171, #172, #173, #174 and #175.

mitar:
Yes. It is meant with #171. "For now it can be a mock location, later on we will interface it with real data."

But it could even be hard-coded in Python file. This data. ;-)

I will add to my TODO to prepare interface for this. Maybe it will be eaiser like that.

agiz:
Hm.. Not sure we are talking about the same thing. I was thinking of merely creating Document that has field latitude, longitude, address and some other fields that might be needed and are common in multiple issues. Such Document should then be used by these issues as Embedded Document.

mitar:
Yes. I know. But this could also be simply stored in the file for now. Maybe there is no reason for the database. Will do this.

agiz:
But for now... I will pass all the data there is to end user, pick that user's location at random and display zoomed in map of that location.

mitar:
Good.

PR #188, opened 2012-05-24
Issue 173 rev1 (team4)

agiz:


agiz:
For issue #173.

mitar:
Don't use nodewatcher's Google Maps code because it uses deprecated API. Please use new Google Maps API.

We also do not need to display all nodes. We will display only location of current node (the node user is connected to or is closest to the user), the rest is not important. If users wants to browse all nodes then she should go to nodewatcher.

Also status of nodes is not important. Here we are just interested in existence and location of nodes, not their status.

So everything could be much much simplified. We need only one simple map with one location of one node (available in `request.node`). And then on this map we draw also events and other information.

mitar:
Please update to new code from main repository.

agiz:
I will update to Google Maps V3 but please specify this issue in more in details. Do not want to oppose your reasoning but technically this pull request satisfies ticket #173. 

I know that displaying all nodes will not be in final version but I worked with what I had/have. Maybe it is too soon to work on this issue due to yet unpredictable mechanisms to get required data. So in order to avoid further de-railed coding, give more details to this pull request before I update this (or even create new issue) with main repository.

mitar:
You do not have to worry about getting data. This is now taken care of. You have API (node object in `request.node`) and this is all you need, I hope. ;-)

So just  display a small map, put node there. Display some info about the node and link to its webpage. You can differentiate between node to which user is connected to or node to which user is just close to (not much, just some text or something).

If you really want to complicate, you could display a warning somewhere for case when we are displaying a "close" node to the user if real distance between the node and the user is larger then some limit (like more than 100 m).

agiz:
Closing in favor of https://github.com/wlanslovenija/PiplMesh/pull/201.

Issue #175, opened 2012-05-14
Add panel with events from culture.si


mitar:
Add panel with events from [culture.si](http://www.culture.si/). Add locations of those events to Google Maps panel (#173).

Define a background task to fetch regularly this data for all nodes (their locations) in the network.

If needed, work together with them to define a suitable API.

agiz:
How accurate should this be? Is it ok to mark the city? If this is too wide can you provide an idea of locating "Kino Partizan" for example.

mitar:
No. It is far from OK to mark the city. We need direct geo location of the venue.

And they have location displayed on the map, so they do have this information.

You should write to them.

agiz:
They have some, but not all or maybe I am not seeing something. http://www.culture.si/en/Bled_Castle has location, http://www.culture.si/en/Klub_Zakon does not.

mitar:
We will ignore those who do not have location. It should be in owner's interest to update its entry on culture.si.

agiz:
There are literary no locations for any event listed in http://www.culture.si/en/Culture.si:Events. http://www.culture.si/en/Cankarjev_dom_Gallery is one rare exception. I do not know what to do here...

mitar:
You have two options:
- you ignore those events for which venues are not known
- or you do reverse geocoding for as a fallback for those unknown venues from their address, Google has such an API, but we years ago when I was playing with it was not so good (got only the street right, the number was like mostly ignored), maybe it is better now

I see that they have updated [their API](https://developers.google.com/maps/documentation/geocoding/). Maybe you can play a bit with it and see if you can get from the result how precise it. Maybe we can try to do reverse geocoding and if the result is precise enough, we use it. Otherwise we ignore the venue.

(There is a third approach, which is a semi-commercial Slovenian database, which we could might "lend" to run through all addresses once and store that for later. Only new venues would then have problems.)

agiz:
Almost all events fall under first option. We are left with options two and three. I think the third option will work best.

mitar:
Please first try how good Google API is.

mitar:
(We could always have it as a fallback from third option for new venues.)

agiz:
I will, but this is another issue then, no? Reverse geocoding might be used in other issues as well.

mitar:
You can do it here. And if there will be later on some other use too, we can extract it out.

mitar:
[...]

PR #193, opened 2012-05-25
Issue 175 rev1 (team 4)

agiz:
- Events from culture.si
- reverse geo api

For issue: #175

I am using bits from #171


mitar:

Didn't we say that we will be accessing external data sources using background tasks and not from client?

agiz:
Umm.. There was a discusion but we have not come to the conclusion. Plus that will depend on the nature of an API. If it can be accessed by client, there is no need in using our server's cycles. And you do not have to comment this at 3:00 am :) No need to rush, I can wait and I am always on stand-by haha :P

mitar:
I think we decided that we will have a fixed set of locations (our nodes) and prefetch data regularly for them. Because also for clients not connected to our portal through our wireless nodes, we will pretend they are close to some existing node.

agiz:
I do not understand how this changes anything. I am strongly against server prefetching where it is not necessary. As a top contributor I believe I have the right to my own opinion especially now that I have already made the whole API the way it is, including reverse geolocation that will be impossible to do on server side (will promptly overcome the limits). Do with this pull request whatever you like. Also no hard feelings :)

mitar:

mitar:
Of course you have right to your opinion! I also have right to my opinion about your final mark! No hard feelings. ;-)

mitar:
OK, seriously. I do not see any other way than doing it on the server. If venues are missing their geolocation, you have to map all missing geolocation venues' addresses to their geolocation. And if you do this on client and on server, there is same number of requests. But on the server you can do this with different databases, slower, with delays ... Also once you do this, search is much easier (get me all venues 100m around the node).

Also, on server side we can do much better reverse geolocation, because we can also use other services to do that.

Furthermore, we want to support search over all content on the webpage. If we prefetch this data, we can also much easier search over it.

I really do not see a reason for not prefetching it. See prefetching as a caching proxy to their API, where we also clean a bit data, and improve it. If we will have millions of users, do you want that poor their API falls apart because of us?

mitar:
Anything new here? Please update from main repository, too.

Issue #147, opened 2012-04-21
Give user an option to select between Facebook and Gravatar picture


mitar:
When linking will be possible (#76), we should add to option for users to select which user picture they want displayed: Gravatar or Facebook. If user first registered through Facebook, the default should be Facebook, if user first registered directly, then after linking default should change to Facebook if user does not have a Gravatar picture. (Is there a way to test this?)

Blocked by: #76

mitar:
Authentication has moved to separate package, so this should be implemented there: https://github.com/mitar/django-mongo-auth/issues/6

Issue #137, opened 2012-04-19
Keep translations updated


mitar:
Keep translations updated.

Issue #76, opened 2012-03-21
Linking users between Facebook and our registration


mitar:
Users should be able to link accounts registered directly with Facebook (and other authentication backends), so that they can login in any way and get the same account.

mitar:
Authentication has moved to separate package, so this should be implemented there: https://github.com/mitar/django-mongo-auth/issues/5

Issue #69, opened 2012-03-21
Implement django-lazyuser for MongoDB


mitar:
Implement something like django-lazyuser for MongoDB. So that also temporary users would be in MongoDB. Because we are implementing now everything ourselves, we can use only one user model/object with all needed fields.

I would just change decorator django-lazyuser users into additional middleware which would in practice be very similar, but would automatically be applied to all requests (which is I believe what we want).

aljana:
Is it allowed to use some code from django-lazysignup or it must be written from start? I'm not planning to do this (most probably), just curious. :)

mitar:
Oh, good to mention. I have to fix this ticket's description first. I don't believe that the way proposed above is the best anymore. :-)

mitar:
OK, changed description. Now to the answer: you will have to write your own because to many things are different. For example, we can use the same user object, just add a flag (in any case it would be good to add two flags: `is_registered` and `is_email_verified`, currently we do not have any e-mail address verification).

mitar:
You can look at django-lazysignup, but please write your own, because of the licenses issues. We could use their code because it is BSD licensed, but then we will have to add some text to our license that our code incorporates code from them under such and such conditions. We can do that, if really necessary, but try without. :-) If you do, then please [read this](https://www.softwarefreedom.org/resources/2007/gpl-non-gpl-collaboration.html) and accordingly update our `LICENSE` file in a pull request. (We do not use license headers in the code.)

domenkozar:
Why do you use mongodb for everything? Django - batteries included. Except that you are ditching all the batteries you get. It's like getting Pamela Anderson for a girlfriend but then operating her tits to size A.

mitar:
We want to feel pure and close to batteries. :-) Not artificial. :-)

But it is not true. The fundamental goal of Django is [loose coupling](https://docs.djangoproject.com/en/1.4/misc/design-philosophies/#loose-coupling):

> A fundamental goal of Django’s stack is loose coupling and tight cohesion. The various layers of the framework shouldn’t “know” about each other unless absolutely necessary.
> 
> For example, the template system knows nothing about Web requests, the database layer knows nothing about data display and the view system doesn’t care which template system a programmer uses.
> 
> Although Django comes with a full stack for convenience, the pieces of the stack are independent of another wherever possible.

[...]

PR #198, opened 2012-05-28
Account linking

MartinJakomin:
Account linking, avatar selection, translation, support for lazyuser...

Updated Facebook and Twitter registration.
Linking account with Facebook and Twitter. #Issue: #76
Selecting avatar. #Issue: #147
Updated translations. #Issue: #137
Support for lazyuser #Issue: #69


MartinJakomin:
I also fixed bug which allowed to register 2 users with same username. Now users with facebook or twitter username that is already in our database, are redirected to register page forcing them to create an account with different username.

MartinJakomin:
Aded linking for facebook account. Still much work to do, like deleting "old" facebook account and implementing some kind of protection (like password confirmation).
How do you like it so far ??

MartinJakomin:
Ok, Facebook linking should be complete now (except for password confirmation).

And now, I would really need some feedback before continuing with my work :)

mitar:
Merge with latest from main repository please.

piplmesh/account/backends.py
```diff
@@ -45,41 +45,28 @@ def authenticate(self, facebook_token=None, request=None):
...
-        # Retrieve access token
-        url = urllib.urlopen('https://graph.facebook.com/oauth/access_token?%s' % urllib.urlencode(args)).read()
-        response = urlparse.parse_qs(url)
-        access_token = response['access_token'][-1]
-    
-        # Retrieve user's public profile information
-        data = urllib.urlopen('https://graph.facebook.com/me?access_token=%s' % access_token)
-        fb = json.load(data)
 
+        fb, access_token = getFBData(facebook_token, request, 'facebook_callback')
```
mitar: No no no. Please use verbose (long, spelled out) variable and function names. Also you can simply define methods on the class, no need for global functions?

piplmesh/account/backends.py
```diff
@@ -45,41 +45,28 @@ def authenticate(self, facebook_token=None, request=None):
...
+                    'last_name': fb.get('last_name'),
+                    'email': fb.get('email'),
+                    'gender': fb.get('gender'),
+                    'facebook_link': fb.get('link'),
+                    }
+            )
+            user.facebook_token = access_token
+            user.save()
+            return user
+        except Exception:
```
mitar: Never ever just hide exceptions. No. You do not want this.

mitar: (OK, you could do it sometimes, but not really here. Why would this here be good?)

piplmesh/account/fields.py
```diff
@@ -72,10 +72,13 @@ class GenderField(mongoengine.StringField):
     def __init__(self, *args, **kwargs):
         kwargs.setdefault('max_length', 6)
         kwargs.setdefault('choices', GENDER_CHOICES)
-        kwargs.setdefault('default', GENDER_CHOICES[1][0])
+        kwargs.setdefault('default', 'None')
```
mitar: No. This is wrong.

piplmesh/account/fields.py
```diff
@@ -72,10 +72,13 @@ class GenderField(mongoengine.StringField):
...
 
         super(GenderField, self).__init__(*args, **kwargs)
 
+    def _validate(self, value):
+        pass
```
mitar: This too.

piplmesh/account/fields.py
mitar: Default should be ''.

piplmesh/account/templates/user/account.html
```diff
@@ -15,6 +15,18 @@
         {% endwith %}
     </div>
     <div id="password_change">
-        <a href="{% url password_change %}">{% trans "Change your password" %}</a></p>
+        <p><a href="{% url password_change %}">{% trans "Change your password" %}</a></p>
+        {% if not user.facebook_id %}
```
mitar: if-else exists in Django templates. ;-)

piplmesh/account/templates/user/account.html
```diff
@@ -15,6 +15,18 @@
...
+            <p><a href="{% url facebook_link %}">{% trans "Link account with Facebook" %}</a></p>
+        {% endif %}
+        {% if user.facebook_id %}
+            <p><a href="{% url facebook_unlink %}">{% trans "Unlink account with Facebook" %}</a></p>
```
mitar: _from_ instead of _with_?

piplmesh/account/templatetags/twitter.py
```diff
@@ -5,6 +5,7 @@
 
 register = template.Library()
 
+# TODO: Also get larger picture (like facebook)
```
mitar: We are doing this differently. Please merge with main repository.

piplmesh/account/views.py
```diff
@@ -44,13 +44,80 @@ def get(self, request, *args, **kwargs):
         if 'code' in request.GET:
             # TODO: Add security measures to prevent attackers from sending a redirect to this url with a forged 'code'
             user = auth.authenticate(facebook_token=request.GET['code'], request=request)
+            if user is None:
+                messages.error(request, _("Please register with another username and then link your account with Facebook."))
```
mitar: No, we will make sure that username is unique automatically. There is no reason to make people do manual work.

piplmesh/frontend/templates/user/user.html
```diff
@@ -14,6 +14,9 @@
     <div id="picture" >
```
mitar, line 14: For now we will have only 50x50 images of users. We will improve that later.

piplmesh/urls.py
```diff
@@ -34,6 +34,9 @@
     # Facebook
     url(r'^facebook/login/$', account_views.FacebookLoginView.as_view(), name='facebook_login'),
     url(r'^facebook/callback/$', account_views.FacebookCallbackView.as_view(), name='facebook_callback'),
+    url(r'^facebook/linkAccount/$', account_views.FacebookLinkView.as_view(), name='facebook_link'),
```
mitar: I really do not understand why you need multiple views? Why is login not enough? It is very simple. If `request.user` exists (which it will always with lazy login), then you link it. If username is temporary (with lazy login this will be like that), you override it. If any user profile data is missing, you fill it in. Why is there difference between linking and login?

You should only add unlink (good idea). Maybe login should be renamed link?

mitar:
So, good start. But unify login and linking. I will try to do lazy users soon, so this part will be easier.

MartinJakomin:
The funny thing here is, that I initially had linking inside the login view but I thought that you would want seperate views for some reason so I created them :)
Will change it.

piplmesh/account/backends.py
MartinJakomin: The thing is I dont really know which exception I am suppose to catch. Well I know that it is username already exist, but I dont know how to define it.

piplmesh/account/fields.py
MartinJakomin: I dont know what you want me to do here. If I dont have this method overwritten, it checks if value is male or female but we want to have empty value too "".

piplmesh/account/backends.py
mitar: Try it. ;-) Hardcode something and try it. Or go to `./manage.py shell` and run code there manually. Twice. And you will see. ;-)

piplmesh/account/fields.py
mitar: This is why you put required to false and should take care of this. No?

piplmesh/account/templates/user/account.html
MartinJakomin: thats what you get from lazy copy-pasting :)

mitar: And you dare to publicly tell me you are copy-pasting?

piplmesh/account/views.py
MartinJakomin: How will you make sure that someone facebook username is not the same like some username in our database ?

piplmesh/account/fields.py
MartinJakomin: Ok, I will make this method check if value is either male, female or "". Would that be good ?

mitar: Doesn't it do this automatically?

piplmesh/account/templates/user/account.html
MartinJakomin: Someone just hacked into my computer and made that post, I swear :D

But seriously its hard to overcome old habits :)

piplmesh/account/views.py
mitar: We will automatically change that username to something else. Like we could add the sequence number at the end.

piplmesh/account/fields.py
MartinJakomin: It just checks if value is one that is listed (in our case male or female) if you leave it empty it would raise an exception, because empty value is not equal to male or female.

piplmesh/account/templates/user/account.html
mitar: There is punishment for everything and everybody!

piplmesh/account/fields.py
mitar: Then add '' to choices???

MartinJakomin: Then there would be third choice for "empty" gender in registration. Unless I change how the gender field is render :)

mitar: I think it would be much easier if you would simply change:

```
user.gender=form.cleaned_data['gender']
```

to:

```
if form.cleaned_data['gender']:
    user.gender=form.cleaned_data['gender']
```

MartinJakomin: Omg I am cursing myself how on earth I havent came up with this idea myself :)
Btw: I also had to remove setting its default value to "" (otherwise you would still get an exception). So now default value is "the default value :)" which is None.

MartinJakomin:
Facebook login and linking is now complete. Moving on Twitter.

MartinJakomin:
Twitter linking complete.

MartinJakomin:
Updated translations.

MartinJakomin:
Fixed exceptions.

MartinJakomin:
Added profile image selection. (get_image_url function is still the same, will change it in next pull request, and also image selection need to have dynamic choices).

MartinJakomin:
Added dynamic choices. Now everything "should" be completed.

mitar:
Please update from main repository. I have just changed something in this part. Bugfixing.

MartinJakomin:
Updated.

MartinJakomin:
fixed minor bugs.

mitar:
Please update from main repository. There is now basic support for #69, so I think some things change here. Now every request always have a user object stored in the database, an authenticated one (this is, with password, or some third-party authentication). So registration is in fact just linking this user with password (and opportunity to set additional fields at the same time, like username). Facebook authentication is linking it with Facebook. And so on. So this makes everything unified. Would you adapt the rest of the code accordingly? And score also #69 if you do so?

The only tricky thing is what to do with objects lazy user created before login (for example, forgets to log in before posting). Because of this we should have both for login or registration (this could become one page) a checkbox (default to true) asking the user if she wants to retain posts and other objects under this logined user account. If true, posts should be transferred to logined account.

MartinJakomin:
I dont really like this concept. I think it would be better and much much easier to do, if we left thing the way they are and address the lazy users differently. 
What if registration and login would just delete (instead of merge or something) the lazy user and create a new one (or log in to existing one). So basicly the lazy user would just be there only for users who are infact to lazy to sign up :)

mitar:
You should not look at what is easy to do, but what is best for the user. And I also believe this now is much easier to do. Mostly instead of `get_or_create` you just update `request.user`. ;-) Or you can also create a simple class-based form to update the instance.

Anyway, we want to allow people to post without registration or third-part authentication. And now imagine a common scenario. Users comes, starts posting, and then discovers that this is really great and that she would like to keep those posts/profile. So she decided to register. And we should keep those things when she registers.

And now registration is really simple. Just setting username and password.

Then another scenario. Existing user comes and forgets that she is not logged in. So she posts something. And then she sees, ups, this is not me. So she decides to log in. But of course posts should be transferred to this new user.

But sometimes on shared computer you will not want to want to take over posts from somebody else when you will be loggin in. (Ha, we need also "reset" button somewhere for those lazy users. So that if you do not want to register, but still want a new session.)

So, don't worry about those merging of content. I will make a simple API for that. For now just make this registration and linking for users. And forget about the content. This will be done afterwards. Now just make that user can keep or not keep the user instance.

MartinJakomin:
Sorry for the delay, didnt have any spare time this weekend. 
Now I think everything should be done (along with support for lazyuser).

Btw: How should we detect and delete unused lazyuser account ? We don't want to end up with hundreds of unused accounts in our database. But that is a problem for future :)

mitar:
Yes, this is similar to the problem of old registered users. But here we have more of them. One solution is to not create lazy users for robots. ;-) But because we will not have much content accessible from the Internet, this is not really a problem here. And then you simply define `./manage.py` command to cleanup old lazy users. We can also clean old normal users in similar manner. And then you run this command daily. Hm, or we could have daily celery task. This would be better in our case, because it is easier to parallelize. I will open ticket for this.

piplmesh/account/backends.py
```diff
@@ -6,6 +6,7 @@
 
 from mongoengine import queryset
 from mongoengine.django import auth
+from mongoengine.queryset import DoesNotExist, OperationError
```
mitar: GROAN. Import only modules.

piplmesh/account/backends.py
```diff
@@ -95,18 +100,113 @@ def authenticate(self, twitter_token, request):
...
+                    username = twitter_user.screen_name
+                    username += str(i)
+                    i+=1
+                    continue
+                else:
+                    raise
+
         return user
 
+def getFacebookData(facebook_token, request):
```
mitar: Don't we have function names with underscores?

piplmesh/locale/sl/LC_MESSAGES/django.po
```diff
@@ -273,15 +304,15 @@ msgstr "Rezultati za:"
 
 #: .\frontend\templates\userlist.html.py:4
 msgid "Show/Hide Users"
-msgstr "Prikaži/Skrij uporabnike"
+msgstr "Pokaži/Skrij uporabnike"
 
 #: .\frontend\templates\userlist.html.py:4
 msgid "Online Users"
-msgstr "Povezani uporabniki"
+msgstr "Prijavljeni uporabniki"
```
mitar: Online = on-line = na-vezi. Closer to "povezani" then "prijavljeni" (which is "logged in", which is not that).

piplmesh/locale/sl/LC_MESSAGES/django.po
```diff
@@ -250,15 +290,6 @@ msgstr "Poglej zadetke na "
 msgid "Logout"
 msgstr "Odjava"
 
-#: .\frontend\templates\outside.html.py:8
```
mitar: Why is this removed?

piplmesh/locale/sl/LC_MESSAGES/django.po
```diff
@@ -177,13 +230,21 @@ msgstr "Registriraj se"
 msgid "Change Password"
 msgstr "Spremeni geslo"
 
+#: .\account\templates\user\password_check.html.py:5
+msgid "Password Confirmation"
```
mitar: Should maybe also English version be: _Password (repeat)_?

piplmesh/locale/sl/LC_MESSAGES/django.po
```diff
@@ -153,6 +190,22 @@ msgstr "Pošlji"
...
+msgid "Unlink account from Facebook"
+msgstr "Razveži račun s Facebook-om."
+
+#: .\account\templates\user\account.html.py:25
+msgid "Link account with Twitter"
+msgstr "Poveži račun s Twitter-jem"
+
+#: .\account\templates\user\account.html.py:27
+msgid "Unlink account from Twitter"
+msgstr "Razveži račun s Twitter-jem."
```
mitar: Dot at the end or not?

What now, do we have accounts or users? What are we linking?

piplmesh/locale/sl/LC_MESSAGES/django.po
mitar: This is not "internal error". ;-)

piplmesh/account/views.py
```diff
@@ -151,16 +231,17 @@ def form_valid(self, form):
         user.first_name=form.cleaned_data['first_name']
         user.last_name=form.cleaned_data['last_name']
         user.email=form.cleaned_data['email']
-        user.gender=form.cleaned_data['gender']
+        user.gender=form.cleaned_data['gender'] or None
```
mitar: `form.cleaned_data.get('gender')`

piplmesh/account/views.py
```diff
@@ -132,8 +213,7 @@ def form_valid(self, form):
         return super(RegistrationView, self).form_valid(form)
 
     def dispatch(self, request, *args, **kwargs):
-        # TODO: Is this really the correct check? What is user is logged through third-party authentication, but still wants to register with us?
-        if request.user.is_authenticated():
+        if not request.user.is_anonymous():
```
mitar: You do know that `is_anonymous` is defined as `not is_authenticated`? :-)

piplmesh/account/backends.py
```diff
@@ -49,39 +50,43 @@ def authenticate(self, facebook_token, request):
...
-        # Retrieve access token
-        url = urllib.urlopen('https://graph.facebook.com/oauth/access_token?%s' % urllib.urlencode(args)).read()
-        response = urlparse.parse_qs(url)
-        access_token = response['access_token'][-1]
-    
-        # Retrieve user's public profile information
-        data = urllib.urlopen('https://graph.facebook.com/me?%s' % urllib.urlencode({'access_token': access_token}))
-        fb = json.load(data)
 
+        fb, access_token = getFacebookData(facebook_token, request)
```
mitar: OK, I will complicate, because I know you will manage. ;-)

Please name things better. `fb` does not tell anything. _facebook data_ neither. (And use underscores.) So function name should clearly tell what it does. You probably have a function which returns Facebook user profile data? Then write that!

piplmesh/account/backends.py
```diff
@@ -49,39 +50,43 @@ def authenticate(self, facebook_token, request):
...
-                'last_name': fb.get('last_name'),
-                'email': fb.get('email'),
-                'gender': fb.get('gender'),
-                'facebook_link': fb.get('link'),
-            }
-        )
-        user.facebook_token = access_token
-        user.save()
+
+        username = fb.get('username', fb.get('first_name') + fb.get('last_name'))
```
mitar: If we are precise, `get` return by default `None`. So `+` could fail. ;-) Better would be:

```
username = fb.get('username', fb.get('first_name', '') + fb.get('last_name', '')) or None
```

In this way if also `first_name` and `last_name` are empty, then username becomes `None`, what you can check later on and simply leave randomly generated username as it is.

mitar:
Before continuing with comments, I still don't know why you have linking and login to Facebook separate? Is it not possible to do this at one place? So linking or login with Facebook is simply connecting _current user_ (being it lazy or not) with Facebook. You login to Facebook, fetch user profile data, fill any empty user profile fields, save link, token, and this is it.

The only question is about username. For this I would add one flag which tells if username was automatically generated or is this set value. And if it is automatically generated, you override it from Facebook's username.

There is two open questions:
- do we try to generate some username from name and/or surname, or simply leave it at automatically generated value if Facebook does not provide us with username
- do we try to generate unique username if user with such username provided by Facebook already exists, or we just leave it at automatically generated value for user to manually pick it later

piplmesh/account/backends.py
```diff
@@ -95,18 +100,113 @@ def authenticate(self, twitter_token, request):
...
+                    user.username = username
+                    user.first_name = twitter_user.name
+                    user.twitter_token_key = twitter_token.key
+                    user.twitter_token_secret = twitter_token.secret
+                    user.twitter_link = "http://twitter.com/#!/" + twitter_user.screen_name
+                    user.save()
+                    break
+            except OperationError, e:
+                msg = str(e)
+                if 'E11000' in msg and 'duplicate key error' in msg and 'User' in msg:
```
mitar: Check for `username`, not for `User` (`User` is document name, we want to see if error is about username field).

piplmesh/account/backends.py
```diff
@@ -95,18 +100,113 @@ def authenticate(self, twitter_token, request):
...
+                    user.save()
+                    break
+                except DoesNotExist:
+                    user = request.user
+                    user.twitter_id = twitter_user.id
+                    user.username = username
+                    user.first_name = twitter_user.name
+                    user.twitter_token_key = twitter_token.key
+                    user.twitter_token_secret = twitter_token.secret
+                    user.twitter_link = "http://twitter.com/#!/" + twitter_user.screen_name
```
mitar: No, don't do this. If data was not given, don't construct data at some random point in time. What if Facebook will change their URLs? Then you will go through all database and change URLs?

But I agree that this is good to have on user object. So you can define such property:

```
@property
def twitter_link(self):
    return 'http://twitter.com/#!/%s' % user.twitter_screen_name
```

Maybe it would be useful to store all usernames given from third-party authentication mechanisms (and even display them).

BTW, user can change their username on Twitter, so one more reason not to hard-code twitter_link (and you should also make it so that `twitter_screen_name` is saved every time user logs in through Twitter. Probably this goes for all fields from the third-party authentication provider.

piplmesh/account/fields.py
```diff
@@ -10,11 +10,25 @@
     ('female', _("female")),
 )
 
+DEFAULT_IMAGE_CHOICES = (
+    ('gravatar', _("Gravatar")),
+    ('twitter', _("Twitter")),
+    ('facebook', _("Facebook")),
+    )
```
mitar: Indentation.

mitar: I would not hard-code this!

I would extend our backend classes with a method which returns a tuple for the image for that backend (or None, if not supported). And then I would iterate over all authentication backends here and construct this choices.

mitar: The order of backends would also set the preference for images.

piplmesh/account/models.py
```diff
@@ -54,6 +57,7 @@ class User(auth.User):
     connection_last_unsubscribe = mongoengine.DateTimeField()
     is_online = mongoengine.BooleanField(default=False)
 
+
```
mitar, line 85: Unnecessary line?

mitar:
OK. Enough comments for now. ;-)

piplmesh/account/backends.py
MartinJakomin: Tell that to the pycharm automatic import :P

mitar: Tell [them yourself](http://youtrack.jetbrains.com/dashboard/PY#tab=0). ;-)

piplmesh/locale/sl/LC_MESSAGES/django.po
MartinJakomin: Well I havent seen "Povezani" on any slovenian site so far only "Prijavljeni", but I can change it. It just sounds so weird for some reason.

piplmesh/account/views.py
MartinJakomin: Found out that too late, sorry.

piplmesh/locale/sl/LC_MESSAGES/django.po
mitar: And on how many Slovenian sites do they have list of currently online users in real-time?

mitar: Hm, then maybe "Omreženi uporabniki"? And we also change English to "Networked Users"? If this is our tagline, we could use it ...

piplmesh/account/views.py
mitar: In fact I have noticed, that it will not be anymore. I have to fix it.

mitar:
I am sorry, but you will have to update from main repository again. We just got Google authentication in. ;-)

MartinJakomin:
Btw: Do you really want me to merge facebook linking and log in. The code is really quite a mess now, and merging it to one function would really overcomplicate it, in my opinion. But heck, like my opinion counts anyway :)
Long story short, if you really want to have one function I can do it.

mitar:
No, not one function. You can have things split into multiple functions. I am just talking about one code path. Don't you agree that linking and login are very very similar? Almost the same? The only difference is does it set username or not? No?

MartinJakomin:
Yes, now that we have lazy user it is almost the same :)
I just hope you feel my frustration here, how I spent doing linking firstly without lazyuser in mind, and now completely changing it, because of it :)

piplmesh/locale/sl/LC_MESSAGES/django.po
MartinJakomin: It is so confusing when more people are doing translates, and than merging this file is just nightmare :)

piplmesh/locale/sl/LC_MESSAGES/django.po
MartinJakomin: I think we agree on, that we have users with accounts or something like that (so unlink "account" should be good)

piplmesh/account/views.py
MartinJakomin: That is not good, we need it like that to prevent exception:
"ValidationError, Exception Value: Errors encountered validating document: gender: Value must be one of ['male', 'female']("gender")"

mitar:
But I warned you about that! That things will change once we will have lazy users. But you still played with that. ;-)

piplmesh/locale/sl/LC_MESSAGES/django.po
mitar: OK.

piplmesh/account/views.py
mitar: OK.

mitar:
I just pushed some changes to Facebook authentication. To show you what I have in mind. Please comment.

piplmesh/account/backends.py
MartinJakomin: How does "in" actually works, does it only search for words (separated with "space"), or does it actually search for all substrings.
Because error is: E11000 duplicate key error index: PiplMesh.user.$_types_1_username_1

mitar: Google that? Try it in Python shell?

piplmesh/account/fields.py
MartinJakomin: I already have getImageChoices(user), which gets the available choices. Default choices are just for defining field class.

mitar: I know what you have. But as I said, also those default choices should be based on backends. So that if somebody adds a new backend, they are automatically extended. Not that you have to search around the code where all you have also to change stuff.

MartinJakomin:
I saw your facebook and twitter changes. Will create new login for facebook and twitter based on your changes tomorrow.
Btw: Are u sure you replaced all of the "if facebook.id", because I used them alot :D

mitar:
Not sure, verify. ;-)

mitar:
Some more changes for you. Enjoy. ;-)

mitar:
Even more changes. Please update from main repository. ;-) Be faster. ;-)

mitar:
If you are asking what I am not sure: is checking for third-party user profile data always false if account is not linked, or should be check for the content, for example in `is_authenticated` method.

MartinJakomin:
Hi,
after a month of doing other stuff for school and having a little vacation :), I'm back with some newly updated code. I have managed to finish (hopefully) Facebook linking. If you still remember our conversation about multiple accounts with same fb data, or accounts with multiple fb, etc.... Now everything should be (in theory) considered and rightly managed. 
Please check out my code, and for now focus only on Facebook login/linking (because Twitter login/linking will look the same, after you are completely happy with Facebook login :)  ). Also ignore other stuff (image choice and translations), they will be renewed in the near future...

MartinJakomin:
Fixed minor bug.

mitar:

piplmesh/account/backends.py
```diff
@@ -37,6 +37,18 @@ def get_user(self, user_id):
     def user_class(self):
         return models.User
 
+    def generateUsername(self,usrnm):
```
mitar, line 40: What is this? Camel case? Where is consistency?

`usrnm`??? You want me to get a heart attack?

piplmesh/account/backends.py
```diff
@@ -37,6 +37,18 @@ def get_user(self, user_id):
...
+        i = 1;
+        original = usrnm
+        while True:
+            try:
+                user = models.User.objects.get(username=usrnm)
```
mitar, line 45: You cannot do it like this. This is not correct. There is a race condition here. Between checking with `get` and later on saving. This is low chance, but still a chance. And with many users, chance increases.

There are two ways to do this:
- set username in atomic MongoDB operation in a loop until it succeeds (probably in the first try), in every loop fixing username in some way
- saving mongoengine document in a loop until it succeeds, doing the same as above if it fails

So, in fact there is only one way: to try to store data and see if database says that everything is OK. The only problem with second approach is that it could be from some other reason that saving fails. If you can determine from the error message that it is really because of non-unique username, then this is OK.

piplmesh/account/backends.py
mitar, line 40: I would rename it to `unique_username`. But as I said, you will not do it like this. :-)

piplmesh/account/fields.py
```diff
@@ -10,11 +10,34 @@
     ('female', _("female")),
 )
 
+DEFAULT_IMAGE_CHOICES = (
```
mitar, line 13: O no. We will not do it like this. You should read this from used backends. You should never assume that somebody who will define a backend, will go all around the code and check where it has also to fix something. New backend is installed with settings file. And this is it. The rest should be done automatically.

So add some method to backends which returns you wanted tuples, and then create those choices with iterating through that.

piplmesh/account/backends.py
```diff
@@ -52,18 +64,16 @@ def authenticate(self, facebook_access_token, request):
         try:
             user = self.user_class.objects.get(facebook_profile_data__id=facebook_profile_data.get('id'))
         except self.user_class.DoesNotExist:
-            # TODO: Based on user preference, we might create a new user here, not just link with existing, if existing user is lazy user
```
mitar, line 55: Why removing this TODO?

piplmesh/account/backends.py
```diff
@@ -52,18 +64,16 @@ def authenticate(self, facebook_access_token, request):
...
             user = request.user
-            # TODO: Is it OK to override Facebook link if it already exist with some other Facebook user?
 
         user.facebook_access_token = facebook_access_token
         user.facebook_profile_data = facebook_profile_data
 
         if user.lazyuser_username and facebook_profile_data.get('username'):
             # TODO: Does Facebook have same restrictions on username content as we do?
-            user.username = facebook_profile_data.get('username')
+            user.username = self.generateUsername(facebook_profile_data.get('username'))
```
mitar, line 76: And why are you doing this only for Facebook backend? You should fix it everywhere.

piplmesh/account/backends.py
```diff
@@ -52,18 +64,16 @@ def authenticate(self, facebook_access_token, request):
         try:
             user = self.user_class.objects.get(facebook_profile_data__id=facebook_profile_data.get('id'))
         except self.user_class.DoesNotExist:
-            # TODO: Based on user preference, we might create a new user here, not just link with existing, if existing user is lazy user
             # We reload to make sure user object is recent
             request.user.reload()
             user = request.user
-            # TODO: Is it OK to override Facebook link if it already exist with some other Facebook user?
```
mitar, line 59: What we decided about this?

piplmesh/account/fields.py
```diff
@@ -10,11 +10,34 @@
...
+    ('foursquare', _("Foursquare")),
+    ('google', _("Google")),
+)
+
 ERROR_MESSAGES = {
     'bounds': _("Value is out of bounds."),
     'callable': _("Callable did not return datetime.date or datetime.datetime object."),
 }
 
+def getImageChoices(user):
```
mitar, line 27: You have really programmed in some other language in between, haven't you?

piplmesh/account/forms.py
```diff
@@ -96,6 +96,7 @@ class UserBasicInfoForm(forms.Form):
         label=_("Gender"),
         choices=fields.GENDER_CHOICES,
         widget=forms.RadioSelect(),
+        required=False,
```
mitar, line 99: Hm?

piplmesh/account/forms.py
```diff
@@ -118,6 +119,17 @@ class UserAdditionalInfoForm(forms.Form):
...
     """
 
+    profile_image = forms.ChoiceField()
+
+    def __init__(self, *args, **kwargs):
+        super(UserAdditionalInfoForm, self).__init__(*args, **kwargs)
+        user_choices = fields.getImageChoices(self.user)
+        self.fields['profile_image'] = forms.ChoiceField(
+            label=_("Profile image"),
+            required=False,
```
mitar, line 129: Why false?

piplmesh/account/forms.py
```diff
@@ -118,6 +119,17 @@ class UserAdditionalInfoForm(forms.Form):
     Class with user additional information form.
     """
 
+    profile_image = forms.ChoiceField()
+
+    def __init__(self, *args, **kwargs):
+        super(UserAdditionalInfoForm, self).__init__(*args, **kwargs)
+        user_choices = fields.getImageChoices(self.user)
```
mitar, line 126: `user_choices` -> better name.

piplmesh/account/models.py
```diff
@@ -116,20 +117,26 @@ def has_usable_password(self):
...
+        elif self.profile_image == 'foursquare' and self.foursquare_profile_data and 'photo' in self.foursquare_profile_data:
             return self.foursquare_profile_data['photo']
-        
-        elif self.google_profile_data and 'picture' in self.google_profile_data:
+
+        elif self.profile_image == 'google' and self.google_profile_data and 'picture' in self.google_profile_data:
             return self.google_profile_data['picture']
 
-        elif self.email:
+        elif self.email and self.profile_image == 'gravatar':
```
mitar, line 139: Keep the same order of checks, first checking `profile_image`, then conditions to use that.

piplmesh/account/models.py
```diff
@@ -116,20 +117,26 @@ def has_usable_password(self):
...
 
+    def get_twitter_link(self):
+        return 'http://twitter.com/#!/%s' % self.twitter_profile_data['screen_name']
+
     def get_image_url(self):
-        if self.twitter_profile_data and 'profile_image_url' in self.twitter_profile_data:
+
+        # TODO: Save images after each login, so you don't have to contact facebook or twitter on each request
+
+        if self.profile_image == 'twitter' and self.twitter_profile_data and 'profile_image_url' in self.twitter_profile_data:
```
mitar, line 127: Can you move all this to backends? So all this backend-specific checks to some method there, which returns true if image url can be given. And then another method which gives it?

piplmesh/account/views.py
```diff
@@ -35,48 +41,151 @@ def get_redirect_url(self, **kwargs):
...
-
+        # TODO: Check if 'client_id' is same as ours
         if 'code' in request.GET:
             args = {
                 'client_id': settings.FACEBOOK_APP_ID,
                 'client_secret': settings.FACEBOOK_APP_SECRET,
                 'redirect_uri': request.build_absolute_uri(urlresolvers.reverse('facebook_callback')),
                 'code': request.GET['code'],
-            }
+                }
```
mitar, line 98: Unnecessary tab.

piplmesh/frontend/templates/user/user.html
```diff
@@ -20,10 +20,15 @@
...
-        <p>{% trans "Last login:" %} {{ object.last_login }}</p>
-        {% if object.facebook_profile_data.link %}
-            <p><a href="{{ object.facebook_profile_data.link }}"><img src="{{ STATIC_URL }}piplmesh/images/fb_logo.png" height="30" width="30" alt="{% trans "Facebook link" %}" /></a></p>
-        {% endif %}
+        <p>{% trans "Last login:" %} {{ object.last_login }}
+        <p>
+            {% if object.facebook_profile_data.link %}
+                <p><a href="{{ object.facebook_profile_data.link }}"><img src="{{ STATIC_URL }}piplmesh/images/fb_logo.png" height="30" width="30" alt="{% trans "Facebook link" %}" /></a></p>
+            {% endif %}
+            {% if object.twitter_id %}
```
mitar: `twitter_id`? We don't have that anymore.

piplmesh/account/views.py
```diff
@@ -267,21 +420,23 @@ class AccountChangeView(edit_views.FormView):
...
         user.save()
         messages.success(self.request, _("Your account has been successfully updated."))
         return super(AccountChangeView, self).form_valid(form)
 
     def dispatch(self, request, *args, **kwargs):
-        # TODO: With lazy user support, we want users to be able to change their account even if not authenticated
-        if not request.user.is_authenticated():
-            return shortcuts.redirect('login')
+        if not request.user.password:
+            messages.error(request, _("Before proceeding to your account page, you must set up your password."))
```
mitar, line 438: Why?

piplmesh/account/views.py
```diff
@@ -267,21 +420,23 @@ class AccountChangeView(edit_views.FormView):
 
     def form_valid(self, form):
         user = self.request.user
-        user.first_name = form.cleaned_data['first_name']
-        user.last_name = form.cleaned_data['last_name']
+        user.first_name=form.cleaned_data['first_name']
```
mitar: Why no spaces around `=`?

piplmesh/account/views.py
```diff
@@ -267,21 +420,23 @@ class AccountChangeView(edit_views.FormView):
...
+        user.last_name=form.cleaned_data['last_name']
         if user.email != form.cleaned_data['email']:
             user.email_confirmed = False
             user.email = form.cleaned_data['email']
-        user.gender = form.cleaned_data['gender']
-        user.birthdate = form.cleaned_data['birthdate']
+        user.gender=form.cleaned_data['gender'] or None
+        user.birthdate=form.cleaned_data['birthdate']
+        user.profile_image=form.cleaned_data['profile_image']
```
mitar: Why those changes?

piplmesh/account/views.py
```diff
@@ -106,17 +221,56 @@ def get(self, request, *args, **kwargs):
             twitter_auth.set_request_token(request_token.key, request_token.secret)
             twitter_auth.get_access_token(verifier=oauth_verifier)
 
+            if request.user.is_authenticated():
+                if request.user.twitter_id:
+                    messages.error(self.request, _("Your account is already linked with Twitter."))
+                else:
+                    backends.twitterLink(twitter_token=twitter_auth.access_token, request=request)
```
mitar, line 228: `twitterLink`? What is this?

piplmesh/account/views.py
```diff
@@ -233,16 +385,18 @@ class RegistrationView(edit_views.FormView):
...
-            email = form.cleaned_data['email'],
-            gender = form.cleaned_data['gender'],
-            birthdate = form.cleaned_data['birthdate'],
-        )
-        new_user.set_password(form.cleaned_data['password2'])
-        new_user.save()
+
+        user = self.request.user
+        user.username = form.cleaned_data['username']
+        user.first_name = form.cleaned_data['first_name']
```
mitar, line 391: I would set those values only if they were specified in registration? Or do you prepopulate from existing user object? So if fields have values from linking?

piplmesh/account/templates/user/link_choice.html
```diff
@@ -0,0 +1,17 @@
```
mitar, line 0: Where is this file used?

piplmesh/account/backends.py
mitar, line 55: OK, I see it bellow. You ask user what to do ...

piplmesh/account/backends.py
mitar, line 59: OK, you ask user what to do.

piplmesh/account/views.py
```diff
@@ -35,48 +41,151 @@ def get_redirect_url(self, **kwargs):
...
+                }
 
             # Retrieve access token
             response = urlparse.parse_qs(urllib.urlopen('https://graph.facebook.com/oauth/access_token?%s' % urllib.urlencode(args)).read())
             # TODO: Handle error, what if response does not contain access token?
             access_token = response['access_token'][0]
+            facebook_data = json.load(urllib.urlopen('https://graph.facebook.com/me?%s' % urllib.urlencode({'access_token': access_token})))
+            self.fb_token = access_token
+
+            if request.user.is_authenticated():
```
mitar, line 107: Uh is this code ugly. :-)

piplmesh/account/views.py
```diff
@@ -35,48 +41,151 @@ def get_redirect_url(self, **kwargs):
...
+                if request.user.facebook_profile_data:
+                    if request.user.facebook_profile_data['id'] == facebook_data.get('id'):
+                        messages.error(self.request, _("Your account is already linked with Facebook."))
+                        return http.HttpResponseRedirect(self.get_success_url())
+                    else:
+                        try:
+                            models.User.objects.get(facebook_profile_data__id=facebook_data.get('id'))
+                            self.choices = (('nothing', _("Do nothing")),
```
mitar, line 115: Format this a bit better. Newlines and such.

piplmesh/account/views.py
```diff
@@ -35,48 +41,151 @@ def get_redirect_url(self, **kwargs):
...
+            self.fb_token = access_token
+
+            if request.user.is_authenticated():
+                if request.user.facebook_profile_data:
+                    if request.user.facebook_profile_data['id'] == facebook_data.get('id'):
+                        messages.error(self.request, _("Your account is already linked with Facebook."))
+                        return http.HttpResponseRedirect(self.get_success_url())
+                    else:
+                        try:
+                            models.User.objects.get(facebook_profile_data__id=facebook_data.get('id'))
```
mitar, line 114: Please use `count()` here. Instead of try/except.

piplmesh/account/views.py
```diff
@@ -35,48 +41,151 @@ def get_redirect_url(self, **kwargs):
...
+                            self.choices = (('nothing', _("Do nothing")),
+                                            ('unlink_and_overwrite', _("Unlink other account and overwrite this account with new information")),
+                                )
+                            messages.error(self.request, _("There is already one account linked with this Facebook account and your account is already linked with another Facebook account."))
```
mitar, line 118: OK. This error message is completely and really completely comprehensible. :-)

BTW, messages is not useful to use if you are displaying a form. Much better is to simply set some flag and then based on that flag in the view template display the message.

messages are used when doing redirects and you want to display something on the target page.

I would all this code move to form validation. And this could be simply a form validation error. So you check all this things in the form and if there is no value for choices, you display a validation error and ask user to select the value. If value is already there, you process the form.

mitar, line 118: You could make a form which simply have the same hidden name as comes from Facebook (`code`). And this form would then automatically be populated from GET data. Which would be missing choice. So it would fail. And after resubmit it would succeed.

mitar:
OK. It is getting better. So after some though, I like the idea of form in callback. I don't really like the exposed Facebook parameters, but this is Facebook quilt of not doing POST request. So I would just create a form mapping to those GET parameters. And if form is validated correctly, go through and redirect. Or display an error message.

For images, please move that to backend. I know that we have on other places names hardcoded, but we will try to clean that later.

If you want to do something extra, you could create your own generic class-based view for this linking workflow, and then for each backend just extend it and set some properties like name, backend class and such. But maybe this could be another pull request or a ticket for somebody else. But there is a lot of repetitive stuff here.

mitar:
Ping?

mitar:
(Merge with latest from the main repository.)

MartinJakomin:
Hi,

Dont worry, I'm still here. I just have a ton of work to do right now, will try to do some changes asap.

mitar:
Please merge with main repository.

And: ping!

MartinJakomin:
I have merged with main repository. 
Sadly I do not have a any spare time so I wont be able to work on this till the next week :(
But I will finish this before September :)

Btw:  lxml 2.3.5 does NOT work on Windows. So I put lxml version 2.3 into requirements (that version works fine with Windows) for all our Windows developers :)

mitar:
There is a ticket for lxml: #266

So you are saying that using the 2.3 version which has binary egg available, allows direct installation without compilation?

MartinJakomin:
Firstly I thought it is all about version, but now I tried to install it again using pip and it did not work.
However if you install it like it says on lxml page (http://lxml.de/installation.html) using easy install, everything works perfectly (not the version 2.3.5 though) 
"easy_install --allow-hosts=lxml.de,*.python.org lxml==2.3"

MartinJakomin:
Since when do we have to add "" for url variables. Nevermind fixed it :) Will try to finish this thing till the end of the week.

mitar:
Please merge with latest from main repository.

mitar:
> Since when do we have to add "" for url variables.

Since I everywhere in the code loaded `future` template library. This is future proof. In the future, `url` tag will behave like this. Because currently it is the exception. All other tags should have `"` for constant strings. And I think it is nicer if things are consistent.

mitar:
I am very sad, but you know that you will have to this pull request again. You shouldn't push whole lxml.

mitar:
(You could always fix this before the push locally.) Now is too late. But maybe you could try to delete it from GitHub.

mitar:
And about lxml, this is why there is a ticket for documentation. So this above, how lxml is installed, should be added to documentation, that before pip install you should manually install lxml on Windows.

MartinJakomin:
I noticed too late that I pushed whole lxml (I was testing how to install it and it failed and left the whole directory there).
But I removed it later, so why is this a problem ?
Nvm, will make a new request when everything will be fixed.

mitar:
> But I removed it later, so why is this a problem ?

No, you didn't remove it. It is still stored in the repository. You just removed it from the latest version. But whole history is forever stored in the repository. Together with your lxml blunder.

mitar:
Hm, what will we do about this?

MartinJakomin:
working on that ... :)

mitar:
Is there any pull request to watch? :-)

mitar:
I moved account code to separate Python package. So you should continue there:

https://github.com/mitar/django-mongo-auth

The idea is that this can be reusable also elsewhere. Because we did quite nice job.

PR #202, opened 2012-05-29
Issue 175 rev2 (team 4)

agiz:
- NOT Final
- Updates to issue #171 (mongoengine support for nodes)
- Tasks for fetching nodes and venues from odpiralncasi.com


agiz:
I do not know how to pass venue data to end user and I am afraid to do it wrong. Please suggest. If you want me to do it via API you must know that you must either introduce sessions to API which is a no no in my opinion or pass user node with API call. Also please help me fix tests and other files to completely move nodes to mongoengine. Thank you.

piplmesh/nodes/models.py
```diff
@@ -1,16 +1,39 @@
...
+    
+class NodeModel(object):
     def __init__(self, id, name, location, latitude, longitude, url):
         self.id = id
         self.name = name
-        self.location = location
-        self.latitude = latitude
-        self.longitude = longitude
+        self.address = location
+        self.location = [latitude, longitude]
```
mitar, line 30: Tuples.

PR #203, opened 2012-05-29
Issue 174 rev2 (team 4)

agiz:
- NOT Final
- Updates to issue #171 (mongoengine support for nodes)
- Tasks for fetching nodes and venues from odpiralncasi.com


agiz:
I do not know how to pass venue data to end user and I am afraid to do it wrong. Please suggest. If you want me to do it via API you must know that you must either introduce sessions to API which is a no no in my opinion or pass user node with API call. Also please help me fix tests and other files to completely move nodes to mongoengine. Thank you.

mitar:

mitar:
No, we really do not need to have nodes stored in the database. Assume node ids are unique strings. In reality, for nodewatcher data, they will be UUIDs, each node has its own UUID. So you can safely store this into the database as a key and use `piplmesh.nodes.get_node` to get node object when needed.

mitar:
OK, for example, for odpiralni casi is maybe a good idea to simply do not do any prefetching. No? Because results really depend on the time of the request.

agiz:
Closing in favor of https://github.com/wlanslovenija/PiplMesh/pull/225.

Issue #205, opened 2012-05-30
Show only published posts


mitar:
Only posts with `is_published` flag set to true should be visible to other (non-author) users.

[Here](http://django-tastypie.readthedocs.org/en/latest/authentication_authorization.html) **could** be some answers to how to do this.

PR #220, opened 2012-06-06
Pull request for #205

matevzmihalic:
I added custom authorization, that hides non published posts from other users.

### DIFF
--- a/piplmesh/api/authorization.py
+++ b/piplmesh/api/authorization.py
@@ -1,10 +1,13 @@
 from tastypie import authorization as tastypie_authorization
 
-class PiplMeshAuthorization(tastypie_authorization.Authorization):
+class PostAuthorization(tastypie_authorization.Authorization):
     def is_authorized(self, request, object=None):
         try:
             obj = self.resource_meta.queryset[0]
         except (IndexError, TypeError):
             return True
         
-        return True if obj.is_published else obj.author == request.user
+        if obj.is_published:
+            return True
+        else:
+            return obj.author == request.user
--- a/piplmesh/api/authorization.py
+++ b/piplmesh/api/authorization.py
@@ -1,0 +1,10 @@
+from tastypie import authorization as tastypie_authorization
+
+class PiplMeshAuthorization(tastypie_authorization.Authorization):
+    def is_authorized(self, request, object=None):
+        try:
+            obj = self.resource_meta.queryset[0]
+        except (IndexError, TypeError):
+            return True
+        
+        return True if obj.is_published else obj.author == request.user
--- a/piplmesh/api/authorization.py
+++ b/piplmesh/api/authorization.py
@@ -4,8 +4,8 @@ from tastypie import authorization as tastypie_authorization
 class PostAuthorization(tastypie_authorization.Authorization):
     def apply_limits(self, request, object_list):
         if request and hasattr(request, 'user'):
-            if request.user.is_authenticated():
-                object_list = object_list.filter((queryset.Q(author=request.user) & queryset.Q(is_published=False)) | queryset.Q(is_published=True))
-                print object_list
+            object_list = object_list.filter((queryset.Q(author=request.user) & queryset.Q(is_published=False)) | queryset.Q(is_published=True))
+        else:
+            object_list = object_list.filter(is_published=True)
 
         return object_list
--- a/piplmesh/api/authorization.py
+++ b/piplmesh/api/authorization.py
@@ -4,7 +4,7 @@ from tastypie import authorization as tastypie_authorization
 class PostAuthorization(tastypie_authorization.Authorization):
     def apply_limits(self, request, object_list):
         if request and hasattr(request, 'user'):
-            object_list = object_list.filter((queryset.Q(author=request.user) & queryset.Q(is_published=False)) | queryset.Q(is_published=True))
+            object_list = object_list.filter(queryset.Q(author=request.user) | queryset.Q(is_published=True))
         else:
             object_list = object_list.filter(is_published=True)
 
--- a/piplmesh/api/authorization.py
+++ b/piplmesh/api/authorization.py
@@ -1,13 +1,11 @@
+from mongoengine import queryset
 from tastypie import authorization as tastypie_authorization
 
 class PostAuthorization(tastypie_authorization.Authorization):
-    def is_authorized(self, request, object=None):
-        try:
-            obj = self.resource_meta.queryset[0]
-        except (IndexError, TypeError):
-            return True
-        
-        if obj.is_published:
-            return True
-        else:
-            return obj.author == request.user
+    def apply_limits(self, request, object_list):
+        if request and hasattr(request, 'user'):
+            if request.user.is_authenticated():
+                object_list = object_list.filter((queryset.Q(author=request.user) & queryset.Q(is_published=False)) | queryset.Q(is_published=True))
+                print object_list
+
+        return object_list
--- a/piplmesh/api/resources.py
+++ b/piplmesh/api/resources.py
@@ -38,7 +38,6 @@ class ImageAttachmentResource(AuthoredResource):
 
     class Meta:
         object_class = api_models.ImageAttachment
-        authorization = tastypie_authorization.Authorization()
 
 class LinkAttachmentResource(AuthoredResource):
     link_caption = tastypie_fields.CharField(attribute='link_caption', default='', null=False, blank=True)
@@ -46,7 +45,6 @@ class LinkAttachmentResource(AuthoredResource):
 
     class Meta:
         object_class = api_models.LinkAttachment
-        authorization = authorization.PiplMeshAuthorization()
 
 class AttachmentResource(AuthoredResource):
     class Meta:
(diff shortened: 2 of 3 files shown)


mitar:
Without tests this is useless. First make tests which shows scenarios we want to allow and scenarios we do not want. So we want users to be able to post new posts, only authors should be able to add attachments, only authors should be able to modify anything before publishing. Everybody should be able to comment. Only authors of comments and post owner should be able to delete the comment. Nobody should be allowed to change their comments. Before publishing, author can change the post. After publishing nobody can change the post anymore.

I think we want to define also our own authentication class, which allows everything, but for `get_identifier` returns username.

mitar:
Ah, sorry. I was thinking that you are doing the authorization for what we discussed at the class. Of all necessary checks we still have to do.

OK, if this is just for #205, then it looks good enough. But tests are still missing.

mitar:
I will open another ticket for the rest.

piplmesh/api/resources.py
```diff
@@ -5,6 +5,13 @@
 from piplmesh.account import models as account_models
 from piplmesh.api import models as api_models
 
+class PiplMeshAuthorization(tastypie_authorization.Authorization):
```
mitar: Please move this to another Python module.

piplmesh/api/resources.py
```diff
@@ -30,7 +37,7 @@ class Meta:
         object_class = api_models.Comment
         allowed_methods = ('get', 'post', 'put', 'patch', 'delete')
         # TODO: Make proper authorization, current implementation is for development use only
-        authorization = tastypie_authorization.Authorization()
+        authorization = PiplMeshAuthorization()
```
mitar: This is not good. User does not have `is_published`.

piplmesh/api/resources.py
```diff
@@ -51,7 +58,7 @@ class Meta:
         object_class = api_models.Attachment
         allowed_methods = ('get', 'post', 'put', 'patch', 'delete')
         # TODO: Make proper authorization, current implementation is for development use only
-        authorization = tastypie_authorization.Authorization()
+        authorization = PiplMeshAuthorization()
```
mitar: This is not good. Attachment does not have `is_published` flag.

piplmesh/api/resources.py
```diff
@@ -1,9 +1,9 @@
-from tastypie import authorization as tastypie_authorization, fields as tastypie_fields
+from tastypie import fields as tastypie_fields
 
 from tastypie_mongoengine import fields, resources
 
 from piplmesh.account import models as account_models
-from piplmesh.api import models as api_models
+from piplmesh.api import authorization, models as api_models
```
mitar, line 7: Please split this into multiple lines. You already have if-block. So just continue using it.

matevzmihalic, line 7: You mean like this:
`from piplmesh.api import authorization`
`from piplmesh.api import models as api_models`

mitar:
I moved above comment to #231.

piplmesh/api/resources.py
mitar, line 7: UPS. This is completely wrong space for this comment. I wonder why is here??? You made a change while I was commenting. ;-)

This is for the if-block inside the authorization class. Where you have a inline if-then-else block.

mitar, line 7: (The import here is of course correct. Don't change anything.)

piplmesh/api/authorization.py
```diff
@@ -0,0 +1,10 @@
+from tastypie import authorization as tastypie_authorization
+
+class PiplMeshAuthorization(tastypie_authorization.Authorization):
+    def is_authorized(self, request, object=None):
+        try:
+            obj = self.resource_meta.queryset[0]
+        except (IndexError, TypeError):
+            return True
+        
+        return True if obj.is_published else obj.author == request.user
```
mitar: This. This is unnecessary to do inline.

piplmesh/api/authorization.py
```diff
@@ -0,0 +1,10 @@
+from tastypie import authorization as tastypie_authorization
+
+class PiplMeshAuthorization(tastypie_authorization.Authorization):
+    def is_authorized(self, request, object=None):
+        try:
+            obj = self.resource_meta.queryset[0]
```
mitar: Why is this needed and `object` is not good enough?

piplmesh/api/resources.py
```diff
@@ -45,13 +46,14 @@ class LinkAttachmentResource(AuthoredResource):
...
         object_class = api_models.LinkAttachment
+        authorization = authorization.PiplMeshAuthorization()
 
 class AttachmentResource(AuthoredResource):
     class Meta:
         object_class = api_models.Attachment
         allowed_methods = ('get', 'post', 'put', 'patch', 'delete')
         # TODO: Make proper authorization, current implementation is for development use only
-        authorization = tastypie_authorization.Authorization()
+        authorization = authorization.PiplMeshAuthorization()
```
mitar: Please see my comments from before. Attachments do not have is published flag?

piplmesh/api/authorization.py
matevzmihalic: I don't know what's up with object, but it's always None ;)

piplmesh/api/resources.py
matevzmihalic: Just commited this.

piplmesh/api/authorization.py
mitar: A bug in our Tastypie implementation? Don't workaround. Fix (or at least make a test there which fails and I will fix it).

piplmesh/api/resources.py
```diff
@@ -45,6 +46,7 @@ class LinkAttachmentResource(AuthoredResource):
 
     class Meta:
         object_class = api_models.LinkAttachment
+        authorization = authorization.PiplMeshAuthorization()
```
mitar: Ehm. This does not exist anymore. And this line you can remove, it was not here before.

piplmesh/api/resources.py
```diff
@@ -38,6 +38,7 @@ class ImageAttachmentResource(AuthoredResource):
 
     class Meta:
         object_class = api_models.ImageAttachment
+        authorization = tastypie_authorization.Authorization()
```
mitar: This line you can remove, it was not here before.

piplmesh/api/authorization.py
matevzmihalic: Not our bug, that is just how tastypie works (currently object is always none; nowhere in tastypies code is is_authorized called with object argument).
I replaced is_authorized with apply_limits which is more suited for this kind of work.

piplmesh/api/authorization.py
```diff
@@ -0,0 +1,11 @@
+from mongoengine import queryset
+from tastypie import authorization as tastypie_authorization
+
+class PostAuthorization(tastypie_authorization.Authorization):
+    def apply_limits(self, request, object_list):
+        if request and hasattr(request, 'user'):
```
mitar, line 6: Why would `request` and request user not be available? When you user OAuth? But we will have to change this limits anyway then.

piplmesh/api/authorization.py
```diff
@@ -0,0 +1,11 @@
...
+            if request.user.is_authenticated():
```
mitar: This check is not needed anymore. All our users are always real users, lazy or non-lazy. We are misusing `is_authenticated` now for telling between lazy and non-lazy users. So this is not needed here.

We should just apply filter so that only published and only those authored are visible. The query can be simplified accordingly. Think about it.

piplmesh/api/authorization.py
matevzmihalic, line 6: Don't know, I just copied it from documentation. If it's in there it should be right ;-)

piplmesh/api/authorization.py
matevzmihalic: Yes, this is not needed, I will remove it.

piplmesh/api/authorization.py
mitar, line 6: See in the code when request is not passed in. Because sometimes we could maybe rather fail then return unrestricted access.

(Or return simply empty list if request is not available.)

(So the difference here is between deny by default or allow by default.)

matevzmihalic, line 6: I decided to return only published posts when request or request.user aren't available.

mitar, line 6: Good decision!

piplmesh/api/authorization.py
```diff
@@ -0,0 +1,11 @@
+from mongoengine import queryset
+from tastypie import authorization as tastypie_authorization
+
+class PostAuthorization(tastypie_authorization.Authorization):
+    def apply_limits(self, request, object_list):
+        if request and hasattr(request, 'user'):
+            object_list = object_list.filter((queryset.Q(author=request.user) & queryset.Q(is_published=False)) | queryset.Q(is_published=True))
```
mitar: This can be simplified! ;-)

matevzmihalic: Done!

mitar:
Good!

PR #221, opened 2012-06-06
fixed button cursor and translated

boco:
fixed login button and translated Foursquare and Google login


mitar:
Comment message is wrong. Please do new new pull request with new commit, same content, but correct commit message.

Otherwise content is OK.

boco:
OK

Issue #219, opened 2012-06-04
Form feedback with JavaScript


mitar:
For some forms we can provide feedback even before they are submitted. One example is password matching when user has to repeat it twice. OK, I don't remember any other, but maybe somebody else will. ;-) So things which are not strictly validation (we are doing this still on the server) but improves user experience.

mitar:
Some things were done with #227, but some more things could be done. For example, for password we could check minimal length, for username minimal and maximal length and that it matches our regex. Don't hardcode those constants in those checks, but get them from Python code (move them there to top-level constant variables, then get them included in Django template, where it sets global JavaScript variables, and use that in those checks).

PR #223, opened 2012-06-07
Add password check match

domenc:
Issue #219

### DIFF
--- a/piplmesh/account/templates/user/registration.html
+++ b/piplmesh/account/templates/user/registration.html
@@ -1,6 +1,11 @@
 {% extends "plain.html" %}
 
-{% load i18n %}
+{% load static i18n %}
+
+{% block js %}
+    {{ block.super }}
+    <script src="{% static "piplmesh/js/register.js" %}" type="text/javascript"></script>
+{% endblock %}
 
 {% block title %}{% trans "Registration" %}{% endblock %}
 
--- a/piplmesh/frontend/static/piplmesh/js/home.js
+++ b/piplmesh/frontend/static/piplmesh/js/home.js
@@ -62,5 +62,23 @@ $(document).ready(function () {
 
     $('#search_users').change(redrawUserList).keyup(redrawUserList);
 
-    redrawUserList();
+    redrawUserList();    
 });
+
+$(window).bind("load", function() {
+    $('#id_password2').keyup(function() { checkPassword(); });
+});
+
+function checkPassword() {
+    var pass1 = document.getElementById('id_password1');
+    var pass2 = document.getElementById('id_password2');
+
+    var goodColor = "#66cc66";
+    var badColor = "#ff6666";
+
+    if(pass1.value == pass2.value){
+        pass2.style.backgroundColor = goodColor;
+    } else {
+        pass2.style.backgroundColor = badColor;
+    }
+}
--- a/piplmesh/frontend/static/piplmesh/js/register.js
+++ b/piplmesh/frontend/static/piplmesh/js/register.js
@@ -1,5 +1,6 @@
 $(document).ready(function () {
     $('#id_password2').keyup(checkPassword);
+    $('#id_email').keyup(checkEmail);
 });
 
 function checkPassword() {
@@ -14,5 +15,19 @@ function checkPassword() {
     {
         $(pass1).css('background-color','#ff6666');
         $(pass2).css('background-color','#ff6666');
+    }
+}
+
+function checkEmail() {
+    var email = "#id_email";
+    var emailReg = /^([\w-\.]+@([\w-]+\.)+[\w-]{2,4})?$/;
+
+    var emailValue = $(email).val();
+    if (!emailReg.test(emailValue)) {
+        $(email).css('background-color','#ff6666');
+    }
+    else
+    {
+        $(email).css('background-color','#66cc66');
     }
 }
--- a/piplmesh/frontend/static/piplmesh/js/register.js
+++ b/piplmesh/frontend/static/piplmesh/js/register.js
@@ -1,0 +1,18 @@
+$(document).ready(function () {
+    $('#id_password2').keyup(checkPassword);
+});
+
+function checkPassword() {
+    var pass1 = '#id_password1';
+    var pass2 = '#id_password2';
+
+    if($(pass1).val() == $(pass2).val()) {
+        $(pass1).css('background-color','#66cc66');
+        $(pass2).css('background-color','#66cc66');
+    }
+    else
+    {
+        $(pass1).css('background-color','#ff6666');
+        $(pass2).css('background-color','#ff6666');
+    }
+}


piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -62,5 +62,23 @@ $(document).ready(function () {
 
     $('#search_users').change(redrawUserList).keyup(redrawUserList);
 
-    redrawUserList();
+    redrawUserList();
```
mitar, line 65: Some unnecessary space?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -62,5 +62,23 @@ $(document).ready(function () {
 
     $('#search_users').change(redrawUserList).keyup(redrawUserList);
 
-    redrawUserList();
+    redrawUserList();    
 });
+
+$(window).bind("load", function() {
```
mitar, line 68: Static string.

mitar, line 68: We are putting spaces around `()` in anonymous functions. We always list all arguments, even if we do not use them, so that later developers know that they exist. Does this anonymous function gets any arguments on load?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -62,5 +62,23 @@ $(document).ready(function () {
...
+    $('#id_password2').keyup(function() { checkPassword(); });
```
mitar, line 69: Bind to both keyup and change (what about copy/paste into the input box).

There is no reason to have an anonymous function which just calls `checkPassword`. Just bind `checkPassword`.

piplmesh/frontend/static/piplmesh/js/home.js
mitar, line 68: Oh, why load? jQuery proposes `ready`, which is before `load`. (Load means all dependencies are loaded, ready is when DOM is ready, which is much sooner.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -62,5 +62,23 @@ $(document).ready(function () {
...
+});
+
+function checkPassword() {
+    var pass1 = document.getElementById('id_password1');
```
mitar, line 73: Please use jQuery.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -62,5 +62,23 @@ $(document).ready(function () {
...
+    var pass2 = document.getElementById('id_password2');
+
+    var goodColor = "#66cc66";
```
mitar, line 76: Don't hardcode colors. Use CSS classes.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -62,5 +62,23 @@ $(document).ready(function () {
...
+    var badColor = "#ff6666";
+
+    if(pass1.value == pass2.value){
```
mitar, line 79: Missing spaces.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -62,5 +62,23 @@ $(document).ready(function () {
...
+        pass2.style.backgroundColor = goodColor;
+    } else {
```
mitar, line 81: Go into new line.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -62,5 +62,23 @@ $(document).ready(function () {
...
+
+function checkPassword() {
+    var pass1 = document.getElementById('id_password1');
+    var pass2 = document.getElementById('id_password2');
+
+    var goodColor = "#66cc66";
+    var badColor = "#ff6666";
+
+    if(pass1.value == pass2.value){
+        pass2.style.backgroundColor = goodColor;
```
mitar, line 80: Use jQuery.

piplmesh/frontend/static/piplmesh/js/home.js
domenc, line 68: ready doenst work in FF and Opera

mitar, line 68: I am very very sure that it does. ;-) Check how we are doing this elsewhere in our code.

domenc, line 68: I add $('#id_password2').keyup(checkPassword()); to $(document).ready(function () { but doesn't work

mitar, line 68: `$('#id_password2').keyup(checkPassword);`

mitar, line 68: You should pass reference to the function, not function result. ;-)

domenc:
$('#id_password2').keyup(checkPassword) doesn't work under ready

mitar:
Please show me exactly the code which does not work. And in which why it does not work?

domenc:
$('#id_password2').keyup(checkPassword); under doucment ready

mitar:
Concrete code. ;-)

I really cannot believe that this does not work as I have very similar code working in many other projects. Can you find me any reference on the Internet about this? Can you reference my other concerns and let us see if it still does not work when we change for `ready`?

domenc:
I tested it:

I add next code to the $(document).ready: alert('LOL');
When i was at http://127.0.0.1:8000/ is work fine, but when i waas at http://127.0.0.1:8000/register/ doesn't work. (i use opera and FF). Please, test it on your machine

mitar:
Yes, but this is not the problem with ready but with our broken JavaScript. ;-)

If you would run Firebug you would see that on register page our JavaScript fails (`map.js` expects `node` global variable and it cannot find any).

I will fix this now.

domenc:
ok

mitar:
Merge from main repository.

domenc:
Doesn't work for me

mitar:
Yes, `home.js` is now not included anymore in registration template. You have to define your own JavaScript file for there.

domenc:
Eeee, i see

domenc:
Can you check last two commit?

mitar:
Aha, I didn't notice. You have to write something here for me to be notified. If you just add new commits to the pull request, I am not noticed.

Why you would like to read CSS from CSS file?

You still have event on windows load?

piplmesh/account/templates/user/registration.html
```diff
@@ -1,6 +1,11 @@
 {% extends "plain.html" %}
 
-{% load i18n %}
+{% load static i18n %}
+
+{% block js %}
+    {{ block.super }}
+    <script src="{% static "piplmesh/js/register.js" %}" type="text/javascript"></script>
+{% endblock %}
```
mitar, line 8: Good. Just name it the same as template, _registration_.

piplmesh/frontend/static/piplmesh/js/home.js
mitar, line 65: You added some unnecessary space at the end of this line.

piplmesh/frontend/static/piplmesh/js/home.js
mitar, line 68: TODO.

piplmesh/frontend/static/piplmesh/js/home.js
mitar, line 69: TODO.

mitar:
On pull request, click on "diff" on the top of the page and you will see all comments inline in the code.

mitar:
You have to pull and merge new changes from the main repository.

mitar:
Aha, I see. You left your changes in `home.js` as they were. You have to remove them. And just leave those in new file you made.

piplmesh/frontend/static/piplmesh/js/register.js
```diff
@@ -0,0 +1,33 @@
+$(document).ready(function () {
+    $('#id_password2').keyup(checkPassword);
```
mitar, line 2: Bind to both `keyup` and `change`.

piplmesh/frontend/static/piplmesh/js/register.js
```diff
@@ -0,0 +1,33 @@
...
+    $('#id_email').keyup(checkEmail);
+});
+
+function checkPassword() {
+    var pass1 = '#id_password1';
```
mitar, line 7: No need for those variables. Just do `$('#id_password1')`. We will not rename element IDs so often.

domenc:
OK, i will create a new pull request on this all changes (then I edit all file and commit on github)

piplmesh/frontend/static/piplmesh/js/register.js
```diff
@@ -0,0 +1,33 @@
...
+    var pass2 = '#id_password2';
+
+    if($(pass1).val() == $(pass2).val()) {
+        $(pass1).css('background-color','#66cc66');
```
mitar, line 11: No, in CSS file define CSS class. And then just add or remove those class from from the element.

piplmesh/frontend/static/piplmesh/js/register.js
```diff
@@ -0,0 +1,33 @@
...
+    else
+    {
+        $(pass1).css('background-color','#ff6666');
+        $(pass2).css('background-color','#ff6666');
+    }
+}
+
+function checkEmail() {
+    var email = "#id_email";
+    var emailReg = /^([\w-\.]+@([\w-]+\.)+[\w-]{2,4})?$/;
```
mitar, line 23: This is much much too simplified. E-mail addresses can be much more complex.

piplmesh/frontend/static/piplmesh/js/register.js
```diff
@@ -0,0 +1,33 @@
...
+
+    var emailValue = $(email).val();
+    if (!emailReg.test(emailValue)) {
+        $(email).css('background-color','#ff6666');
+    }
+    else
```
mitar, line 29: `else` and `{` should be on the same line. So:

```
if () {
    ...
}
else {
    ...
}
```

piplmesh/frontend/static/piplmesh/js/register.js
domenc, line 11: A try with addClass and removeClass but for me doesn't work ($('pass1').removeClass('input_ok input_error').addClass('input_ok')) or is this wrong?

mitar, line 11: There is difference between your:

```
$('pass1').removeClass('input_ok input_error').addClass('input_ok');
```

and

```
$(pass1).removeClass('input_ok input_error').addClass('input_ok');
```

domenc, line 11: Sorry, I thought: $(pass1).removeClass('input_ok input_error').addClass('input_ok');

mitar, line 11: That should work. Probably you just have to remove the old class, not both. But still, it should work. Check if Firebug if it really adds classes. Also, you can run your code in Firebug manually. So open Firebug, run `$('#id_password1').removeClass('input_ok input_error').addClass('input_ok');` and see what happens. Is there an error? Does DOM change?

domenc:
I add new issue #227

PR #199, opened 2012-05-28
Pull request for issue #161 and #162

aljana:
The most basic wall with input box and post listing. Not finished yet but so that you can see we are moving forward. Ideas welcome. :)

PR #189, opened 2012-05-24
Issue#160 - Main page panels (team 1)

NejcZupec:
I've added panels on front-end page. Boxes on the right side are collapsible. They can also be dragged-and-dropped. Wall on the left is fluid. 

Am I on the right way? Now I'd like to save the position but I'm not sure how and where. Where these settings should be saved in database or cookies?

Issue #174, opened 2012-05-14
Add panel with open venues from Odpiralni Casi


mitar:
Add panel with open venues from [Odpiralni Casi](http://odpiralnicasi.com/) in node's vicinity. Add locations of those venues to Google Maps panel (#173).

Define a background task to fetch regularly this data for all nodes (their locations) in the network.

If needed, work together with them to define a suitable API.

There should be some link to report errors with the information displayed (directly to Odpiralni casi).

mitar:
Description updated a bit.

agiz:
Not related with (just) this issue but does it matter how we address associative arrays in JS? I used dot notation so far which cannot always be used. If key is integer the only way to access it is by square brackets `foo['123']`.

mitar:
Keys can be only strings. But yes, if it is not JavaScript identifier, it cannot be used with dot syntax. But I have never had this problem: if I was accessing values through hard-coded keys manually, I used keys that can be used in dot format, otherwise I was mostly doing something automatic, through variables, when you anyway have to use bracket syntax.

Can you make a more concrete example?

PR #225, opened 2012-06-07
Issue 174 rev3 (team 4)

agiz:
- API for displaying stores around nodes from odpiralnicasi.com
- Provided short demonstration how to integrate api information into template.


agiz:
- Changed zoom properties on map for purposes of this api (must be changed back to 15 or 16).
- Does square brackets notation I used for arrays needs to be changed for dotted one?
- This pull request is up to date (https://github.com/wlanslovenija/PiplMesh/commit/12e518f8b1379db026a58da261cc67fc0e75dd24)
- This pull request addresses issue #174.

mitar:
Please update from main repository. You have panels API available. Check it out. ;-)

agiz:
Please support laziness and tell how to enable a panel. Can it be determined what panels should load first?

mitar:
Currently all panels under `piplmesh.panels` are added automatically (if you register them into panels_pool, of course). For the rest, there is a pending pull request #189. So ordering them in display and so on.

API also supports defining dependencies between panels. But as currently all panels are always displayed, this is not a problem.

Also order of loading on the page is not important and _should_ not be important if you do it correctly. So, first, you do everything through jquery.ready in every panel.

Then, I suggest you define custom event which you will trigger when map is ready. And other panels can then in their jquery.ready subscribe to that event. Of course, if map is already ready when they subscribe to the event, they should be immediately run, not just waiting for the event which will never come anymore. So it should behave like jquery.ready does, only for maps. I am not really sure if jQuery supports something like this already, but it maybe does. Check it out.

So, you should define your own JavaScript API in the map in the way that other panels can use the map, without being afraid of the loading order.

agiz:
By 

> (if you register them into panels_pool, of course)

do you mean create them under panels directory? Because I have done that. Can underscore in panel name be an issue?

mitar:
Can you show me the code?

No, underscore should not be a problem. But name is automatically derived from the class name otherwise, and it is probably best to just leave it as it is?

mitar:
You have restarted runserver? Really restart, not just leave to its autoreload.

agiz:
Grrr. This is the stupidest github part. If you rename files it stops tracking them or what? I do not even have option to add them manually.

mitar:
Github? Or git? You should be renaming always with `git mv`.

agiz:
I have known this issue before, I was just never affected by it. Anyway, please check this pull request now, it seems to represent my structure correctly.

mitar:
And where is Python code which goes with the panel? There is where you register the panel into the pool.

mitar:
And please name it `odpiralnicasi`. ;-)

mitar:
And `OdpiralniCasiPanel`.

agiz:
Aaaa.So that is what is mean by registering into the pool. Nonetheless I have added that and panel still does not load.

mitar:
Or even name it `openhours`. That is their English name:

http://odpiralnicasi.com/en/

mitar:
Please check everything precisely. You do have `__init__.py` file?

agiz:
Are panels connected by context header? I mean does Map panel name connect with map directory name?

agiz:
I did have `__init__.py` before just forgot to add it, but `panel.py` was indeed missing.

agiz:
Ok, I am able to load panel (named openhours) now, underscore is indeed an issue.

agiz:
This is working at the moment. There are syntactical issues you have to decide what to do with them.

mitar:
No, it was not underscore an issue, but that you have left to the default naming of your panel in Python code. And by default it was generated without underscore. So or you change the panel name, or you change directories. I suggest the later, what you did. ;-)

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -1,3 +1,4 @@
+var map;
```
mitar: No no no. This is not good. What if openhours panel's ready function is run before the this one? You have to fix this as I described before.

piplmesh/panels/openhours/panel.py
```diff
@@ -0,0 +1,12 @@
+from django.utils.translation import ugettext_lazy as _
+
+from piplmesh import panels
+
+class OpenhoursPanel(panels.BasePanel):
```
mitar: It seems they are named _OpenHours_.

mitar: Check `BasePanel` class. You should define dependency on map panel.

piplmesh/panels/openhours/panel.py
```diff
@@ -0,0 +1,12 @@
...
+    def get_context(self, context):
+        context.update({
+            'header': _("Openhours"),
```
mitar: _OpenHours_

piplmesh/panels/openhours/templates/panel/openhours/panel.html
```diff
@@ -0,0 +1,8 @@
+{% extends "panel/panel.html" %}
+
+{% load i18n static sekizai_tags %}
```
mitar, line 3: `i18n` is probably not needed?

piplmesh/panels/openhours/templates/panel/openhours/panel.html
```diff
@@ -0,0 +1,8 @@
...
+
+{% block content %}
+    {% addtoblock "js" %}<script src="{% static "piplmesh/panel/openhours/panel.js" %}" type="text/javascript"></script>{% endaddtoblock %}
+    <div id="odpiralni_casi_info"></div>
```
mitar: Rename.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,66 @@
+$(document).ready(function() {
```
mitar: Space before `()` in anonymous functions.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,66 @@
...
+    $.getJSON(
+        'http://odpiralnicasi.com/spots.json?callback=?',
+        {
+            lat: node.latitude,
+            lng: node.longitude
+        },
+        function(data) {
+            $.each(data['spots'], function (i, venue) {
```
mitar: OK. I will check this later. Please fix the dependency on the map.

piplmesh/panels/openhours/panel.py
agiz: Uff, am I following - I have to define dependencies and implement api to signal map ready or one or the other?

piplmesh/panels/openhours/templates/panel/openhours/panel.html
agiz: openhours_info?

piplmesh/panels/openhours/panel.py
mitar: You have to define dependency in Python. From this we will be able to forbid using only OpenHours panel without the map panel.

Then, on the JavaScript side, you have to make some way in map panel for other panels to get callback when map is ready.

piplmesh/panels/openhours/templates/panel/openhours/panel.html
mitar: Or just `openhours_content`.

piplmesh/panels/openhours/panel.py
agiz: All clear, thanks.

piplmesh/panels/openhours/templates/panel/openhours/panel.html
mitar: OK. I see now that you made it differently than I imagined. ;-)

mitar: I thought that panel will display a list (similar to user list) of all venues around the location, and its open hours. And when you would click on the map, you would get the element in the list scrolled into the view, enhanced somehow to be visible. So panel should be a list of venues, in the order of proximity to the location (order which is given by OpenHours).

Also a search/filter box could be useful.

mitar: And also in other direction. If you click on the venue in the list, you get the venue on the map enhanced (Google Map icon bouncing?).

mitar: Each element on the list should have its open hours displayed, maybe some logo/photo, name, and link to the more info on OpenHours website.

mitar: (This is just an idea, but somewhere around this lines I think it would be most useful to users. Just having a panel to display info is unnecessary.)

agiz: Ok, no problem. Given that will be just a simple list, where and how to display additional information of the venue?

mitar: I would link to the OpenHours? Or you propose something else? We could display a list as collapsed list and when you click on it, it expands with the content. Maybe:

http://jqueryui.com/demos/accordion/

agiz: Accordion look good. Yes, I have to provide link back to odpiralnicasi.com somewhere.

mitar: I thought of linking to it for every venue, for additional information.

agiz: I was thinking the same, sorry for confusion.

mitar: OK, I was thinking. Lets limit this pull request in its scope. So that also others can do something. So please just do basic panel, define dependency in Python, and create JavaScript API for map event handler. And just a simple list of venues with links to OpenHours. And then we will create a ticket for somebody else to implement accordion and stuff.

I think this is also better for you, to concentrate on hard and interesting problems and leave simple user interface tweaking to others. ;-)

agiz: Again, I was thinking the same thing, really :). But I am almost done with all the bling bling. I will use hover instead of accordion and maybe later someone should change it.

mitar: Ah, then say, if you think. ;-)

Hover? Hmm. OK, let's see. ;-)

agiz:
Those sneaky details :). I did not yet implement dependency and `onLoad` event for signaling Google Map is ready.

piplmesh/panels/openhours/templates/panel/openhours/panel.html
agiz: It is not challenging after I know how it is done, I hate finishing and polishing :P.

mitar: Yes, so you should leave this to others. You see that they are complaining that they do not have anything do to. ;-)

mitar:
Then what have you been doing? Exactly those hard problems are for you. :-)

piplmesh/panels/openhours/panel.py
agiz: What do you suggest here? Remove `ready` events from panels that depends on map panel and do initialization from there?

agiz: Also, I cannot get past `dependencies = ()`. Should not this be a list?

mitar: No, you should, ready event is cool. You can do something like:

```
    // other stuff ...

    $(window).bind('piplmesh.mapready', function (event) {
        ...
    });
});
```

But this does not really work if event is triggered before callback is binded. This should still be researched.

mitar: And in map initialization callback you do:

```
$(window).trigger('piplmesh.mapready');
```

mitar: (Changed above code a bit.)

mitar: Why list? Why not tuple? Will you dynamically change values of dependencies? I think not. ;-) So tuple is perfect for this!

mitar:
Aha, this is is:

http://stackoverflow.com/questions/9172742/jquery-ready-like-event-possible

piplmesh/panels/openhours/panel.py
agiz: I do not follow. We have multiple ready events. If openhours event is triggered before map is rendered there is a problem. How do you intend to prevent this? Can you delay some ready events but not others? I know how to catch on Map ready event:

```
google.maps.event.addListenerOnce(map, 'idle', function () ...
```

Can I use this and call panels depending on map from there?

mitar: No, you use ready events for local-to-panel initialization. Then you should create ready-like custom event/API (through Deferred) for other panels to wait for map to load.

OK, please read a bit more around, what I wrote and what I send links. And you can combine it together and you will understand. ;-) If not, tell me and I will be more concrete in explanation.

agiz: Ok, I will look into it tomorrow. 

I tried to add dependency but do not know how :). I overloaded get_dependencies event and tried to add MapPanel there but to no avail. A little hint please.

mitar: > get_dependencies event

Event? Which event? How event?

You just override `dependencies` field in your Python class with a name of the panel (or panels) your panel depends on.

In your case:

```
dependencies = (map_panel.MapPanel.get_name(),)
```

piplmesh/panels/openhours/templates/panel/openhours/panel.html
```diff
@@ -0,0 +1,14 @@
...
+
+{% load i18n static sekizai_tags %}
+
+{% block content %}
+    {% addtoblock "css" %}<link href="{% static "piplmesh/panel/openhours/panel.css" %}" rel="stylesheet" type="text/css" media="screen" />{% endaddtoblock %}
+    {% addtoblock "js" %}<script src="{% static "piplmesh/panel/openhours/jquery-ui-1.8.21.custom.min.js" %}" type="text/javascript"></script>{% endaddtoblock %}
+    {% addtoblock "js" %}<script src="{% static "piplmesh/panel/openhours/panel.js" %}" type="text/javascript"></script>{% endaddtoblock %}
+    <ul id="openhours_list" class="oh_clearfix"></ul>
+    <form action="" id="openhours_filter">
+        <img src="{% static "piplmesh/panel/onlineusers/magnifying_glass.png" %}" alt="{% trans "Filter venues" %}" />
```
mitar: Hm, if this is clipart we are using more times, move it among main PiplMesh images.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,176 @@
...
+        } else if ('company' in self && 'name' in self) {
+            self.name = self.company + " " + self.name;
+        } else if ('company' in self || 'name' in self) {
+            self.name = self.company || self.name;
+        }
+        self.today = self.is_open ? gettext("open") : gettext("closed");
+        self.location = new google.maps.LatLng(self.lat, self.lng);
+        self.image = 'http://www.google.com/intl/en_us/mapfiles/ms/micons/yellow-dot.png';
+        if ('is_open' in self && self.is_open == true) {
+            self.image = 'http://www.google.com/intl/en_us/mapfiles/ms/micons/green-dot.png';
```
mitar: Always use HTTPS versions. What if site will run under HTTPS.

mitar:
Or are you asking how to implement callback for map? Everything you need is here:

http://stackoverflow.com/questions/9172742/jquery-ready-like-event-possible

agiz:
in base.py, line 19
`if not panels_pool.panels_pool.has_panel(dependency):`
to
`if not panels_pool.has_panel(dependency):`
?

piplmesh/panels/base.py
```diff
@@ -16,7 +16,7 @@ def __init__(self):
         from . import panels_pool
 
         for dependency in self.get_dependencies():
-            if not panels_pool.panels_pool.has_panel(dependency):
+            if not panels_pool.has_panel(dependency):
```
mitar, line 19: So this fail at run-time or you listened to PyCharm? Because PyCharm was quite confused with relative imports?

mitar:
Should I review this pull request? Or wait?

piplmesh/panels/base.py
agiz, line 19: Fails at run-time. I do not use PyCharm.

agiz:
Umm.. review please :).

mitar:
You should move jQuery UI to be used by whole PiplMesh as we will be using it probably also elsewhere. Don't use minimized version. Name it `jquery-ui.js`.

mitar:
Where is CSS and icons which are given together with jQuery UI?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.css
```diff
@@ -0,0 +1,69 @@
...
+    color: #006600;
+}
+
+#openhours_list dl {
+    display: none;
+}
+
+.oh-clearfix:after {
+    content: ".";
+    display:block;
```
mitar: Space after `:`.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -1,3 +1,6 @@
+var deferred = $.Deferred();
```
mitar: I don't like name `deferred` for this. This should be something like `map_loaded`.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -20,4 +23,10 @@ $(document).ready(function () {
         map: map,
         title: node.name
     });
+    google.maps.event.addListenerOnce(map, 'idle', function () {
```
mitar: `idle`? Is not there a callback setting to `google.maps.Map` which is run at the end?

mitar: OK, it seems there is really nothing better.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -1,3 +1,6 @@
+var deferred = $.Deferred();
+var map;
+var map_edge;
```
mitar: What is this?

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -20,4 +23,10 @@ $(document).ready(function () {
         map: map,
         title: node.name
     });
+    google.maps.event.addListenerOnce(map, 'idle', function () {
+        var bounds = map.getBounds();
+        var boundCoord = new google.maps.LatLng(node.latitude, bounds.ea.b);
```
mitar: Be more verbose with variable names, _boundCoord_.

mitar:
Ping? Please update from main repository.

mitar:
Ups. No need. It says it can be merged. But ping anyway. ;-)

agiz:
One step at the time. I want to finish this pull first, ok?

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
agiz: This is distance of node from the rendered map's edge.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
agiz: This is point on map's right edge. It is a helper to determine distance from the centered node. I am just wasting time here, I have spent 10 minutes brainstorming what should more verbose name be :). I am not good at this.
`farRightEdge` ?

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
agiz: This is needed to avoid creating markers that do not fit in shown area.

agiz:
Regarding pushing complete jquery UI into root static files... Directory `images` has to be in the same folder as stylesheet file. What do you suggest?

mitar:

mitar:
For jQuery I have in my projects simply a directory inside the static:

```
static/jquery/
  jquery.js
  jquery-ui.js
  css/
    jquery-ui.css
    images/
      ...
```

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
mitar: Aaa. OK, some other name is needed. :-)

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
mitar: OK, I would use "bound" instead of "edge", same as Google uses.

So `mapBound` instead of `map_edge`. And `eastBoundCoordinate` above. So I was not saying that `boundCoord` is wrong as it is, just that you should spell it fully, `boundCoordinate`. Now that I know that it is east, use that. :-)

mitar:
I have documented JavaScript code style here:

https://github.com/wlanslovenija/PiplMesh/pull/199#commitcomment-1707793

mitar:
TODO.

mitar:

agiz:
BUMP! Updated from main repository.

mitar:
Vuii, you are alive!!! :-)

agiz:
Hehe, yes hardly :). I wish I was relaxing all this time as it might be seen from lack of my activity here. Unfortunately that was not the case :(.

agiz:
Umm hi again. Have I overlooked something that needs to be done here?

mitar:
No time on my side. There are others which need to pass the class.

BTW, please update from main repository, again. Sorry!

mitar:
(We got now also big map functionality. The idea is to provide layers there! And that people can move around!)

agiz:
No problem Mitar. I understand how priorities work. Also updated. I have not checked big map yet, I am on a rough schedule myself.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+            self.name = self.combined_name;
+        } else if ('company' in self && 'name' in self) {
+            self.name = self.company + " " + self.name;
+        } else if ('company' in self || 'name' in self) {
+            self.name = self.company || self.name;
+        }
+        self.today = self.is_open ? gettext("open") : gettext("closed");
+        self.location = new google.maps.LatLng(self.lat, self.lng);
+        self.image = 'https://www.google.com/intl/en_us/mapfiles/ms/micons/yellow-dot.png';
+        if ('is_open' in self && self.is_open == true) {
```
mitar: `self.is_open == true`? Why not just `self.is_open`?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+        if ('full_address' in self) {
+            self.address = self.full_address;
+        } else if ('street' in self && 'city' in self) {
+            self.address = self.street + ", " + self.city;
+        }
+        if ('combined_name' in self) {
+            self.name = self.combined_name;
+        } else if ('company' in self && 'name' in self) {
+            self.name = self.company + " " + self.name;
+        } else if ('company' in self || 'name' in self) {
```
mitar: We go into new line after `}`.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    function Venue(data) {
+        var self = this;
+        $.extend(self, data);
+        if ('distance' in self) {
+            self.distance = get_distance(self.distance);
+        }
+        self.address = '';
```
mitar: address is user exposed string? Then `""`.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+        }
+        return parseFloat(d) * multiplier;
+    }
+
+    function toggleBounce(marker) {
+        if (marker.getAnimation() != null) {
+            marker.setAnimation(null);
+        } else {
+            marker.setAnimation(google.maps.Animation.BOUNCE);
+            setTimeout(function(){ marker.setAnimation(null); }, 750);
```
mitar: Spaces around `()`. You can go into newline after {`and before`}`.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+        d += '';
+        var multiplier = 1.0;
+        if (d.toLowerCase().indexOf('km') > -1) {
+            multiplier = 1000.0;
+        }
+        return parseFloat(d) * multiplier;
+    }
+
+    function toggleBounce(marker) {
+        if (marker.getAnimation() != null) {
```
mitar: `!== null`? Or would it be enough to do `if (marker.getAnimation()) ...`?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+                });
+                div.append(venue.address).append(" | ");
+                var days = ['open_mon', 'open_tue', 'open_wed', 'open_thu', 'open_fri', 'open_sat', 'open_sun'];
+                var day_of_week = new Date().getDay();  //0=Sun, 1=Mon, ..., 6=Sat
+                var open_hours = $('<dl/>').prop({
+                    'class': 'oh_tooltip'
+                });
+                $.each(days, function (j, day) {
+                    if (day in venue && venue[day].length > 0) {
+                        if (day_of_week == (j+1)%7) {
```
mitar: Spaces around binary operators.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+                            venue.today = venue[day];
+                        }
+                        open_hours.append($('<dt/>').text(gettext(day.substr(5))).append(":"));
```
mitar, line 116: `gettext(day.substr(5))).append(":")`? That smells wrong!

This will not be possible to translate as strings will not be extracted in translation catalog. What exactly you want to do here? Format date?

Also, just appending `:` might be wrong. For example in French, there is space before `:`.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+                    }
+                });
+                var today_label = $('<span/>').text(gettext("Today:")).append(" ");
+                today_label.append(venue.today);
+                today_label.hover(
+                    function () {
+                        $(this).parent().find('dl:hidden').css('top', venue.li.offset().top + 'px').fadeIn(500);
+                    },
+                    function () {
+                        $(this).parent().find('dl:visible').fadeOut(500);
```
mitar: Use slow or fast instead of 500.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+                    }
+                );
+                div.append(today_label);
+                if (open_hours.children().length > 0) {
+                    div.append(open_hours);
+                }
+                if ('sid' in venue) {
+                    var website = 'http://odpiralnicasi.com/spots/' + venue.sid;
+                    var website_image = $('<img/>').prop({
+                        'src': '/static/piplmesh/panel/openhours/external_website.png',
```
mitar, line 137: Hardcoded path? Ugly! This should not be like this.

You can define HTML fragment in template, use CSS to hide it, and then clone it here from the DOM, and inject it whenever you want.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+                    icon: venue.image
+                });
+                venue.marker = marker;
+                google.maps.event.addListener(marker, 'click', function (event) {
+                    if (venue.li && venue.li.offset() && $('#openhours_list > li:first') && $('#openhours_list > li:first').offset()) {
+                        $('#openhours_list').animate(
+                            {
+                                scrollTop: venue.li.offset().top - $('#openhours_list > li:first').offset().top
+                            },
+                            300,
```
mitar: Use fast or slow.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+                            function () {
+                                if (venue.li.is(':visible')) {
+                                    venue.li.effect('highlight', {}, 2000);
```
mitar: This is to show which venue was clicked?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,182 @@
...
+        });
+    }
+
+    $.getJSON(
+        'http://odpiralnicasi.com/spots.json?callback=?',
+        {
+            lat: node.latitude,
+            lng: node.longitude
+        },
+        function(data) {
```
mitar: In anonymous functions we put space around `()`. Write out all arguments this callback receives, so that future developers will know what is available.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz: Even if function argument? That javascript is actually the first argument of setTimeout.

```
setTimeout(
    function () {
        marker.setAnimation(null);
    },
    750
);
```

?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz: Definitely `!==`.
I am not sure about the second part. The way you put it will execute inner section only if condition is `true` (`not false`) so you are missing the case when condition is `false`. Whether this is enough or not I do not know :).

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz, line 116: Why? I am just trimming `open_` here? The end result looks like `gettext('mon')` for example. No formatting is done here.

Deja vu. I am positive you said that is ok at one time here. Should I change it to `gettext(":")` then?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz: Yes.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz, line 116: `gettext(day.substr(5).append(":"))`?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz: Are you ok with exception here :)?

```
venue.li.effect(
    'highlight',
    {
    },
    slow
);
```

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz, line 137: Yeah this static image should be displayed via CSS.

Which fragment?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
mitar: That's nice, yes. It could be also like:

```
setTimeout(function () {
    marker.setAnimation(null);
}, 750);
```

mitar: BTW, does not Google Maps API something to bounce just few times? I think I remember this.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
mitar: (So, you don't have `!==`, but `!=`.)

mitar: For the second part, OK. Leave it then like this.

I see that in their example they have `!=`. :-) So I don't know. Probably we can leave it.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
mitar, line 137: In general. You can define a HTML fragment in template. And then use that as a base for populating it dynamically. Sometimes (especially for static things) it is easier than just constructing everything with JavaScript. So, instead of creating DOM in JavaScript, you have a hidden/not-used HTML fragment in the template and then you extract/clone that from the DOM, whenever you need it, and fill it with data.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
mitar: Yes, your original code style is OK here. :-)

(Is it really necessary to specify second empty argument? Sometimes jQuery is smart and knows how to detect which parameter is which.

"slow" is a string constant.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
mitar: Hm, you know that there exists `scrollintoView`? Or you want a nice easing effect. :-)))

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
mitar, line 116: I am positive to not said that. Because I am particularly picky about `:`. Because it is a great example to show that you have to think about that, too.

I see that you do not see the main point here. The main point is that for translations to work, catalog of strings for translation should be made and then each such string should be translated. `colllectstrings` `manage.py` command does the former. And it is quite stupid, it parses code and searches for `gettext` calls. So, if you have `gettext("mon")` it extracts this string out.

In your case, it has nothing to extract. It cannot know which all strings you will be translating.

What is common to do in such cases is to pregenerate dictionary/map between your known values and translations. For example, you could have:

```
days = {
    'open_mon': gettext("Monday:"),
    ...
}

And then simply do `text(days[day])` there, this would work.

Now, this is a general approach. In your case, because you are translating week days names, those translations are probably already existing. It might be interesting to research if Django supports such translations in JavaScript by itself, or how to leverage existing translations for days or localized date time formatting.
```

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz, line 137: I kind of understand hidden HTML part. What I do not know is, how high up the hierarchy should this fragment go? As I see it, this would be common approach for mostly static content. If I remember correctly this part is very dynamic (meaning content is likely to change when moving on map) but more importantly this is generated in a loop which implies loading HTML fragment and populating HTML fragment for every item. Have I overlooked something?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz, line 116: Why is `gettext` so stupid :)

> Now, this is a general approach. In your case, because you are translating week days names, those translations are probably already existing. It might be interesting to research if Django supports such translations in JavaScript by itself, or how to leverage existing translations for days or localized date time formatting.

Is this a hint, because your suggestion with associative array looks very convenient.

agiz:
Fixed some minor mistakes, somewhat clueless how to combat others.

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
agiz: I did not know that. Can I keep it this way?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
mitar, line 137: Hm, I defined it often as a one big hidden div just before `</body>`. (We will probably not have only this as a fragment in the future.)

Define another Sekizai block at the end, because multiple panels could all want to add something there. So similar to how we handle JavaScript and CSS blocks.

I do not understand what has moving a map with this? You just use fragment of one entry, then you clone it every time, populating it with data from JSON. So yes, in the loop you clone the fragment, populate it, and insert into the list.

agiz, line 137: I know where to put it. I do not know what should be the content of this fragment.
`<a href="..."></a>`?
Fake edit: I sense that by fragment you mean all the code that constructs item?

I think values used in sekizai blocks must be known before rendering. I am saying that there is little to no gains using preloaded HTML as content of the item is very dynamic so you would still have to go through all this procedure:
1. Load HTML
2. Populate variables
3. Show/hide parts of not yet rendered HTML that are missing for current item (if node does not have website for example)
4. Insert item into list
Plus you would still need to either track all items in an array or enumerate/uniquely label/ these loaded HTML blocks so that you know which to add/remove when switching nodes.

mitar, line 137: This could be a fragment:

```
<a><img src="/static/piplmesh/panel/openhours/external_website.png" /></a>
```

For the rest, I am not sure what we are talking about. Maybe I do not understand the pull request. But I understood, that we are not yet moving map around and the only thing is that we based on node location fetch after page loads in user browser Odpiralni Casi data from their server and display it. And the question is how we render that data we fetch from them. And I am suggesting a simple fragment you have in template prepared and is thus in DOM and you clone it, fill in data, and inject into DOM.

So where you do 2.), which parts are 3.)?

So I am not sure what are you talking about when you are talking about switching nodes? We do not yet support that? Or do we? (We currently support only mocking nodes, but this is just for staff users.)

Anyway, see how we are doing posts or comments to see how you "enumerate/uniquely label/ these loaded HTML blocks".

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
```diff
@@ -0,0 +1,186 @@
...
+            else $(this).addClass('hover');
+        },
+        function() {
+            if (classes) $(this).attr('class', classes);
+            else $(this).removeClass('hover');
+        });
+    });
+}
+
+map_loaded.done(function () {
```
mitar, line 25: Where is `map_loaded` defined?

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
mitar, line 116: > Why is `gettext` so stupid :)

Because we do not know how to solve halting problem. :-)

> Is this a hint, because your suggestion with associative array looks very convenient.

I would suggest you look into `django-missing` package. ;-)

piplmesh/panels/openhours/static/piplmesh/panel/openhours/panel.js
mitar: Check `scrollintoView`. See if goes to there nicely (so that user can see that it moved) or it jumps so that user is not sure what happened. I prefer using existing functions because they are less error prone. And probably work across browsers.

mitar:
So about which things you are clueless? I can try to help.

PR #227, opened 2012-06-08
Add registration email and password JS checker

domenc:
Issue #219 - add JS checker for registration (password matcher and email validation)


domenc:
New pull request - please comment :)

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,29 @@
+$(document).ready(function () {
+    $('#id_password2').keyup(checkPassword).change(checkPassword);
+    $('#id_email').keyup(checkEmail).change(checkEmail);
+});
+
+function checkPassword() {
+    $('#id_password1').keyup(checkPassword).change(checkPassword);
```
mitar: What is this? Why?

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+    else {
+        $('#id_password1').removeClass('input_valid input_invalid').addClass('input_invalid');
+        $('#id_password2').removeClass('input_valid input_invalid').addClass('input_invalid');
+    }
+}
+
+function checkEmail() {
+    var emailReg = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+(\.([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+)*)|((\x22)((((\x20|\x09)*(\x0d\x0a))?(\x20|\x09)+)?(([\x01-\x08\x0b\x0c\x0e-\x1f\x7f]|\x21|[\x23-\x5b]|[\x5d-\x7e]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(\\([\x01-\x09\x0b\x0c\x0d-\x7f]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]))))*(((\x20|\x09)*(\x0d\x0a))?(\x20|\x09)+)?(\x22)))@((([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])([a-z]|\d|-|\.|_|~|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])*([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])))\.)+(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])([a-z]|\d|-|\.|_|~|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])*([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])))$/;
```
mitar: You found this somewhere? Then you could add attribution in the comment above.

mitar: _emailReg_ -> _emailRegex_

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,29 @@
...
+
+    var emailValue = $('#id_email').val();
```
mitar: This variable is probably not needed, you are using it only once?

piplmesh/frontend/static/piplmesh/js/registration.js
domenc: It's not okey this add to document ready because they every time return false at the first time (beacause second input is empty) so i add on keyup on second input if they go back and retype password on in password1 field

piplmesh/frontend/static/piplmesh/js/registration.js
domenc: Source: http://jzaefferer.github.com/jquery-validation/jquery.validate.js
ok

mitar:
Otherwise it looks good now. I would just add some text feedback to the left of the input box with error. So not just colors, but that some text is displayed in the same color. Like _Input invalid._ or something.

domenc:
OK, i add some text after input.. Add source link for email validation, changed name for regex.
Remove $('#id_password1').keyup(checkPassword).change(checkPassword); ??

domenc:
But text in javascript - how to translate it with Django? Or I put text in html?

piplmesh/frontend/static/piplmesh/js/registration.js
mitar: Then add a comment. Otherwise it would be probably better to port `django.core.validators.email_re` to JavaScript. It probably does not require match to port, just some correct escaping. Maybe there is even something on Internet about that.

Because if would be better to use the same regex, so that it succeeds or fails exactly in same cases.

mitar:
I wrote you above my comments. :-) I do not remember what all I wrote. Use Github pull request interface to check it. ;-)

mitar:
Django supports translations in JavaScript. Check our documentation. It is written there how to do it.

domenc:
Add text for invalid email - like for password

piplmesh/frontend/static/piplmesh/js/registration.js
mitar: I would do it like this:
- run one check if both fields are equal only if they are both unempty, and I would do this on blur (so when users goes somewhere else)
- if previous check fails, I would redo the check on every change, to remove the error as soon as it is possible
- also if password is changed further, i would check again immediately and immediately display the error

domenc: Will be this OK? 
Code: blur(checkPassword).keyup(checkPassword).change(checkPassword);

function checkPassword() {
    if ($('#id_password1').val().trim().length > 0 && $('#id_password2').val().trim().length > 0) {
        if ($('#id_password1').val() == $('#id_password2').val()) {
            $('#id_password1').removeClass('input_valid input_invalid').addClass('input_valid');
            $('#id_password2').removeClass('input_valid input_invalid').addClass('input_valid');
            $("#password_info").remove();
        }
        else {
            $('#id_password1').removeClass('input_valid input_invalid').addClass('input_invalid');
            $('#id_password2').removeClass('input_valid input_invalid').addClass('input_invalid');
            $("#password_info").remove();
            var password_invalid_message = gettext('Please re-enter your password.');
            $("#id_password2").after('<span id="password_info" class="input_invalid">' + password_invalid_message + '</span>');
        }
    }
    else {
        $("#password_info").remove();
        $("#password_info").remove();
        $('#id_password1').removeClass('input_valid input_invalid');
        $('#id_password2').removeClass('input_valid input_invalid');
    }
}

domenc:
Waiting cor comment :)

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,50 @@
...
+        $('#id_password2').removeClass('input_valid input_invalid');
+    }
+}
+
+function checkEmail() {
+    // Source: http://jzaefferer.github.com/jquery-validation/jquery.validate.js
+    var emailRegex = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+(\.([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+)*)|((\x22)((((\x20|\x09)*(\x0d\x0a))?(\x20|\x09)+)?(([\x01-\x08\x0b\x0c\x0e-\x1f\x7f]|\x21|[\x23-\x5b]|[\x5d-\x7e]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(\\([\x01-\x09\x0b\x0c\x0d-\x7f]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]))))*(((\x20|\x09)*(\x0d\x0a))?(\x20|\x09)+)?(\x22)))@((([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])([a-z]|\d|-|\.|_|~|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])*([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])))\.)+(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])([a-z]|\d|-|\.|_|~|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])*([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])))$/;
+    
+    if ($('#id_email').val().trim().length > 0) {
```
mitar: I would just do:

```
if (($('#id_email').val()) {
...
```

No trimming, because we are not do any trimming on Python side.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -79,6 +79,14 @@ form input[type="submit"], form input[type="reset"] {
     margin-left: -13em;
 }
 
+.input_valid {
```
mitar: OK, just one side comment, I would only mark fields as invalid, not also as valid. So just adding invalid class or removing it. To simplify things. Because sometimes Python will do also additional checks and field could still be invalid. (Like invalid hostname in e-mail.)

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,50 @@
+$(document).ready(function () {
+    $('#id_password1').blur(checkPassword).keyup(checkPassword).change(checkPassword);
```
mitar: I have not tried it, but I think this does not work OK. So immediately after the first letter of second password input box, error will be displayed, because both will be of non-zero length and non-matching?

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,50 @@
...
+    $('#id_password2').blur(checkPassword).keyup(checkPassword).change(checkPassword);
+
+    $('#id_email').keyup(checkEmail).change(checkEmail);
+});
+
+function checkPassword() {
+    if ($('#id_password1').val().trim().length > 0 && $('#id_password2').val().trim().length > 0) {
```
mitar: Simplify.

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,50 @@
...
+    if ($('#id_email').val().trim().length > 0) {
+        if (emailRegex.test($('#id_email').val())) {
+            $('#id_email').removeClass('input_valid input_invalid').addClass('input_valid');
+            $("#email_info").remove();
+        }
+        else {
+            $('#id_email').removeClass('input_valid input_invalid').addClass('input_invalid');
+            $("#email_info").remove();
+            var email_invalid_message = gettext('Invalid e-mail.');
+            $("#id_email").after('<span id="email_info" class="input_invalid">' + email_invalid_message + '</span>');
```
mitar: Construct HTML elements through jQuery. See how we are doing for user list panel, for example.

piplmesh/frontend/static/piplmesh/js/registration.js
domenc: Ok, i remove keyup event

domenc:
Simplify and remove valid mark

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,60 @@
...
+
+    if ($('#id_email').val()) {
+        if (emailRegex.test($('#id_email').val())) {
+            $('#id_email').removeClass('input_invalid');
+            $("#email_info").remove();
+        }
+        else {
+            $('#id_email').addClass('input_invalid');
+            $("#email_info").remove();
```
mitar: Move this bellow, just before `after` line so that there is smallest delay between removal and adding. (You can construct `message_info` before.)

domenc:
Moved some line

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,58 @@
...
+        }
+        else {
+            var message_info = $('<span/>').prop({
+                'id': 'password_info',
+                'class': 'input_invalid'
+            }).text(gettext('Please re-enter your password.'));
+
+            $('#id_password1').addClass('input_invalid');
+            $('#id_password2').addClass('input_invalid');
+            $("#password_info").remove();
```
mitar: Error should be _Passwords don't match._ -> this is the only thing we are checking here.

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,58 @@
+$(document).ready(function () {
+    $('#id_password1').blur(checkPassword).change(checkPassword);
+    $('#id_password2').blur(checkPassword).change(checkPassword);
```
mitar: OK, other things now all look good. But this we will still have to improve. ;-) So this is still not done correctly. Again: we want that no error is displayed until user enters something in both password input fields and leave (blur) them both. Then, after that, we want that fields are checked on any change/keystroke.

So I would propose you leave here only `blur` handler. This is when we want first checks to be done. Only on blur. And then, in `checkPassword`, I would attach to change/keyup event first time check fails, after `$("#id_password2").after(message_info);` for example.

domenc:
Fixed

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,61 @@
+$(document).ready(function () {
+    $('#id_password1').blur(checkPassword);
+    $('#id_password2').blur(checkPassword);
+
+    $('#id_email').blur(checkEmail).change(checkEmail);
```
mitar: The same goes probably for e-mail. First time we want to check it only on blur, no?

mitar: When is change called? For every key press too?

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,61 @@
...
+                'id': 'password_info',
+                'class': 'input_invalid'
+            }).text(gettext('Passwords do not match.'));
+
+            $('#id_password1').addClass('input_invalid');
+            $('#id_password2').addClass('input_invalid');
+            $("#password_info").remove();
+            $("#id_password2").after(message_info);
+            
+            $('#id_password1').keyup(checkPassword);
```
mitar: It is necessary to handle also change so that copy-paste is handled.

domenc:
Change: first time only check on blur; if they error then check on change and on key up

mitar:
It would be great if you would write this into commit messages. ;-) Such great descriptions as you have written above.

domenc:
:D

mitar:
It looks good. I will test how it works in reality now. ;-)

mitar:
OK. It crashed my browser. We have to do it better. ;-)

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,63 @@
...
+            $("#email_info").remove();
+        }
+        else {
+            var message_info = $('<span/>').prop({
+                'id': 'email_info',
+                'class': 'input_invalid'
+            }).text(gettext('Invalid e-mail.'));
+
+            $('#id_email').addClass('input_invalid');
+            $("#email_info").remove();
```
mitar: OK. This is not good. We cannot remove and add like this, on every event. We should just add if message info text is not yet there.

domenc:
What?!? On opera is working normal

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,63 @@
...
+
+    if ($('#id_email').val()) {
+        if (emailRegex.test($('#id_email').val())) {
+            $('#id_email').removeClass('input_invalid');
+            $("#email_info").remove();
+        }
+        else {
+            var message_info = $('<span/>').prop({
+                'id': 'email_info',
+                'class': 'input_invalid'
```
mitar: Please use some other class for this text. It looks really ugly with red background. Probably reuse the class which is used for marking invalid forms after Django validates.

mitar:
I did some heavy testing. ;-)

domenc:
Please, test it

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -79,6 +79,10 @@ form input[type="submit"], form input[type="reset"] {
     margin-left: -13em;
 }
 
+.input_invalid {
+    color: #ff0000;
```
mitar, line 87: Hm, for input boxes it was OK that background was red. But OK, it is also OK like this now.

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,81 @@
...
+        else {
+            var message_info = $('<span/>').prop({
+                'id': 'password_info',
+                'class': 'input_invalid'
+            }).text(gettext('Passwords do not match.'));
+
+            $('#id_password1').addClass('input_invalid');
+            $('#id_password2').addClass('input_invalid');
+
+            if ($('#password_info')) {
```
mitar: Ehm, what if you would add only if it does not yet exist? Not that you remove and add if it exists?

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,81 @@
...
+            }
+            $('#id_email').after(message_info);
+
+            $('#id_email').keyup(checkEmail).change(checkEmail);
+        }
+    }
+    else {
+        $('#id_email').removeClass('input_invalid');
+        
+        if ($('#email_info')) {
```
mitar: No, this was OK. `remove` will anyway only remove if it exists. You adding should be done only if it does not yet exists.

domenc: ok

piplmesh/frontend/static/piplmesh/js/registration.js
domenc: i try this:

```
        if (!$('#email_info')) {
            $('#id_email').after(message_info);
        }
```

but doesnt work. Isn't ! for not equal? Ca i use $('#email_info').length <= 0)??

mitar: `if (!$('#email_info').length) {`

domenc:
Change JS. Please, test it

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,69 @@
...
+            $('#id_password1').removeClass('input_invalid');
+            $('#id_password2').removeClass('input_invalid');
+
+            $('#password_info').remove();
+        }
+        else {
+            var message_info = $('<span/>').prop({
+                'id': 'password_info',
+                'class': 'input_invalid_text'
+            }).text(gettext('Passwords do not match.'));
```
mitar: We use `" ... "` for user-displayed (translatable) strings. (This way it is easy to search for them.)

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,69 @@
```
mitar: I just now noticed. This should go to `account/static/piplmesh/js/registration.js`. ;-)

piplmesh/frontend/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,69 @@
...
+            }).text(gettext('Passwords do not match.'));
+
+            $('#id_password1').addClass('input_invalid');
+            $('#id_password2').addClass('input_invalid');
+
+            if (!$('#password_info').length) {
+                $('#id_password2').after(message_info);
+            }
+
+            $('#id_password1').keyup(checkPassword).change(checkPassword);
```
mitar: OK, after some testing I discovered why my Firefox crashes. Because new and news event handlers are registered. It seems jQuery does not bind handlers only once.

mitar: OK, so first unbind the handler and then rebind it. But be careful to unbind only your handlers.

domenc: ok

piplmesh/frontend/static/piplmesh/js/registration.js
domenc: OK, but i dont know how to get right url in html file: from this

{% static "piplmesh/js/registration.js" %}

to what?

I try: /static/admin/iplmesh/js/registration.js but admin doesnt exits

domenc:
Only add unbind.

domenc:
OK, i find a way to call registration.js under admin folder..
I run commend manage.py collecstatic and they find my js under admin static folder. How to commit this?? Or only change path to this folder and manage.py when run server automatic add this JS to static file?

mitar:
What admin? Account! We have `piplmesh.account` app with everything for users, registration, etc.

mitar:
So instead of code being in `piplmesh.frontend`, it should be in `piplmesh.account`.

domenc:
Yes, yes. I write wrong :)

domenc:
Is this OK, what i did it?

domenc:
Move file under piplmesh.account and change

 $('#id_email').unbind('blur.checkEmail').blur(checkEmail);

to

 $('#id_email').unbind('blur', checkEmail).blur(checkEmail);

piplmesh/account/static/piplmesh/js/registration.js
```diff
@@ -0,0 +1,69 @@
+$(document).ready(function () {
+    $('#id_password1').unbind('blur', checkPassword).blur(checkPassword);
```
mitar: In `ready` you do not have to unbind anything! This was OK.

domenc: Fixed

PR #229, opened 2012-06-09
Added comment and INTERNAL_IPS.

boco:
Deleted TODO comment.

### DIFF
--- a/piplmesh/settings.py
+++ b/piplmesh/settings.py
@@ -108,7 +108,7 @@ DEFAULT_FILE_STORAGE = 'piplmesh.utils.storage.GridFSStorage'
 # URL prefix for internationalization URLs
 I18N_URL = '/i18n/'
 
-#List of internal IPs from which django-pushserver is accessible
+# List of configured IPs from which django-pushserver passthrough callbacks are allowed
 INTERNAL_IPS = (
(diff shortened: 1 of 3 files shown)


piplmesh/settings.py
```diff
@@ -108,6 +108,11 @@
 # URL prefix for internationalization URLs
 I18N_URL = '/i18n/'
 
+#List of internal IPs from which django-pushserver is accessible
```
mitar: Not really true, only from where _django-pushserver passthrough callbacks_ are accessible.

Also this setting configures other things, so comment should reflect that. Like _This list also configures IPs from which django-pushserver passthrough callbacks are allowed._

mitar: We are doing space just after the `#`. See how other comments are done.

boco:
Changed comment.

mitar:
Now you have to update `requirements.txt` with new version of django-pushserver so that this setting will have an effect.

boco:
So I just change "django-pushserver==0.1.8" to, let's say, "django-pushserver==0.1.9" ?

mitar:
Yes, I made 0.1.9 version with your changes.

boco:
Updated.

PR #232, opened 2012-06-10
jQuery update

domenc:
jQuery update to version 1.7.2 and replace develement version to production version (it's 153KB less)


mitar:
What is this? Don't upgrade to production lib, because we are developing. For production, we can always combine all JavaScript together and minimize it ourselves. But it is good that you can debug also JavaScript libraries when you have problems. So you can update it, in separate pull request, but leave it non-minimized. And same filename. Just file replace.

domenc:
Yes, i forget :) too fast for pull request - i delete it but i can't

mitar:
No problem, leave it as it is. Closed. ;-)

domenc:
ok:)

Issue #160, opened 2012-05-14
Main page panels


mitar:
For main page (home) we should need some layout with panels. I am not sure how dynamic they should be, so how much users could interact with them, but a bit is probably good. So it is not good that user could completely change layout, so that people would get distracted.

For now I would do something similar to what is [shown here](http://dev.wlan-si.net/wiki/PiplMesh). I would have a wider (two "block" lengths) left column with walls. And then multiple columns to the right (narrower, one "block" length), depending on the width of the window. So what people could change is order of panels in the right column. And then those panels would reflow in left-to-right manner to multiple columns based on how wide the window is. One would be to have only left and right column. If wider, more columns would get displayed, panels reflowed.

User could collapse the panel, so only titlebar of the panel would be displayed. (This, together with reordering those smaller panels, would be how much dynamic we would allow for now.)

Later on, user could remove and add panels (if we will support third-party panels, for example). Also we could support that user can create additional wall columns. So you would be able to have one (zero?) or more wall panels on the left, and then reflowed smaller panels on the right.

Even more later on, users could pop-out wall panels into smaller individual windows.

User's layout (order of panels, and which panels are collapsed/opened) should be automatically stored.

But all this is open for discussion (and examples). Anybody using iGoogle and has some feedback from there?

PR #252, opened 2012-06-18
Issue #160 - Main page panels, rev2

NejcZupec:
This is a new pull request for main page panels. Panels are now draggable and droppable. Their position and collapsed status is saved in database. Everything is made with ajax. When window size is changed, all panels are reallocated.


piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
+function howManyColumns() {
+    var panelsWidth = $('#panels').width();
+    var columnPanelsWidth = $('.panels_column').width() + parseInt($('.panels_column').css('margin-left'));
+
+    return parseInt(panelsWidth / columnPanelsWidth);
+}
+
+function movePanel(id, columnId) {
+    $('#'+id).appendTo($('#panels').children().eq(columnId));
```
mitar: Spaces around `+`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+}
+
+function resetColumns() {
+    $('#panels').children().each(function () {
+        $(this).children().each( function () {
```
mitar: Space before `function`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
+function howManyColumns() {
+    var panelsWidth = $('#panels').width();
+    var columnPanelsWidth = $('.panels_column').width() + parseInt($('.panels_column').css('margin-left'));
+
+    return parseInt(panelsWidth / columnPanelsWidth);
+}
+
+function movePanel(id, columnId) {
```
mitar: `columnId` is what? Index? Or id?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+function resetColumns() {
+    $('#panels').children().each(function () {
+        $(this).children().each( function () {
+            $(this).appendTo($('#panels').children().eq(0))
+        });
+    });
+
+    var count = 0;
+    $('#panels').children().each(function () {
+        if (count != 0)
```
mitar: Do it inline.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+            $(this).remove();
+        count++;
+    });
+}
+
+function fillWithColumns() {
+    var currentColumns = $('#panels').children().length;
+
+    for(i = currentColumns; i < howManyColumns(); i++) {
```
mitar: Space after `for`.

mitar: Why you call `howManyColumns` every time in the loop?

piplmesh/frontend/static/piplmesh/js/home.js
mitar: Write out all arguments function in `each` gets. Everywhere.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+        $('#panels').append('<div class="panels_column"></div>');
+    }
+}
+
+function orderPanelsDefault() {
+    var num_of_panels = $('.panels_column').children().length;
+    var num_of_columns = howManyColumns();
+
+    for (i=0; i < num_of_panels; i++) {
```
mitar: Spaces around `=`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+        var toColumn     = i % num_of_columns;
```
mitar: Spaces?

mitar: Please write variables camel case.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+        $('.panels_column').children().eq(num_of_panels-i-1).appendTo($('#panels').children().eq(toColumn));
```
mitar: Spaces around arithmetic operations.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+        $(this).children().each( function () {
+            var item = {
+                id: $(this).attr('id'),
+            };
+            column.push(item);
+        });
+        items.push(column);
+    });
+    
+    $.get('/panels/order/', 'data=' + JSON.stringify( {panels: items} ));
```
mitar: Do I see a hard-coded URL?

mitar: You are doing query string passing of variables here?

GET request should never change anything in the database. Please update to POST.

mitar: No need for space around `{}`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+}
+
+function orderPanels() {
+    $.get('/panels/order/get', 'data=' + JSON.stringify( {noOfColumns: howManyColumns()} ), function (data) {
```
mitar: Hardcoded URL? `/panels/order/get/` instead of `/panels/order/get`?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+        for (var i = 0; i < data['panels'].length; i++) {
```
mitar: Should you use `var` in for loop or not? What is best practice? Please use it everywhere if yes.

For iterating over lists and dicts you can use jQuery's `each` method.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+    $('#spinner').css('top', '100px');
+    $('#spinner').show();
+
+    fillWithColumns();
+    orderPanels();
+    collapsePanels();
+    makeColumnsSortable();
+    makePanelsUpdatable();
+
+    setTimeout('hideSpinner()', 800);
```
mitar: This is ugly. You should use callbacks when everything finish. Not just some arbitrary timeout.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+}
+
+function makeColumnsSortable() {
+    $('.panels_column').sortable({
+        connectWith: '.panels_column',
+        handle: '',
+        cursor: 'move',
+        placeholder: 'placeholder',
+        forcePlaceholderSize: true,
+        opacity: 0.6,
```
mitar: Sadly, JavaScript does not allow comma at the end like Python does.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,144 @@
...
+
 $(document).ready(function () {
+    preparePanels();
+
     $('.panel .header').click(function (event) {
+        var visible = $(this).next().is(':visible');
         $(this).next('.content').slideToggle('fast');
+
+        if (visible) {
+            var param = "1";
```
mitar: String? Static?

piplmesh/frontend/templates/home.html
```diff
@@ -18,27 +18,17 @@
...
-        SOME MORE JUICY CONTENT
-        </p>
-        <p>
-        HOME PAGE CONTENT HERE
-        </p>
-        <p>
-        SOME MORE JUICY CONTENT
-        </p>
+    <div id="wall">
+
```
mitar: Unnecessary line.

piplmesh/frontend/templates/home.html
```diff
@@ -18,27 +18,17 @@
...
-        HOME PAGE CONTENT HERE
-        </p>
-        <p>
-        SOME MORE JUICY CONTENT
-        </p>
+    <div id="wall">
+        
+    </div>
+    <div id="spinner" style="display: block;">
+        <img src="../static/piplmesh/images/ajax-loader.gif" />
```
mitar: NOT GOOD!

piplmesh/frontend/views.py
```diff
@@ -1,6 +1,9 @@
 from django import http, template
 from django.conf import settings
+from django.core import serializers
 from django.core.files import storage
+from django.http import HttpResponse
```

piplmesh/frontend/views.py
```diff
@@ -73,3 +76,47 @@ def forbidden_view(request, reason=''):
         'reason': reason,
         'no_referer': reason == csrf.REASON_NO_REFERER,
     })))
+
+def panels_collapse(request):
+    user = account_models.User.objects.get(id=request.user.id)
+    message = simplejson.loads(request.GET['data'])
+
+    if (message['collapsed'] == "1"):
```
mitar: Constant string.

mitar: Probably better to use true and false values? Especially if you pass this through JSON?

piplmesh/panels/dummy4/panel.py
```diff
@@ -0,0 +1,17 @@
...
+class DummyPanel4(panels.BasePanel):
+    def get_context(self, context):
+        context.update({
+            'header': _("Dummy panel 4"),
+            'content': '\n\n'.join(lorem_ipsum.paragraphs(random.randint(1, 1))),
+            'id': 'dummy4',
+        })
+        return context
+
+panels.panels_pool.register(DummyPanel4)
```
mitar: Can you do that those panels are registered only if we are running in debug mode?

piplmesh/frontend/views.py
```diff
@@ -73,3 +76,47 @@ def forbidden_view(request, reason=''):
...
+
+def get_panels_order(request):
+    user = account_models.User.objects.get(id=request.user.id)
+    order = simplejson.loads(request.GET['data'])
+    noOfColumns = str(order['noOfColumns'])
+
+    if user.panels_order.has_key(noOfColumns):
+        order = user.panels_order[noOfColumns]
+    else:
+        order = { "panels": [] }
```
mitar: Unnecessary space.

piplmesh/frontend/templates/home.html
```diff
@@ -18,27 +18,17 @@
...
+    </div>
+    <div id="panels" style="visibility: hidden;">
```
mitar: Hm. Never use inline CSS.

Why you hide by default? You should render only those which are activated anyway?

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Index. I changed columnId to columnIndex.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Done.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/templates/home.html
NejcZupec: Fixed.

piplmesh/frontend/views.py
NejcZupec: Done.

piplmesh/frontend/views.py
NejcZupec: Fixed.

piplmesh/panels/dummy4/panel.py
NejcZupec: How can I do that? Should I read DEBUG value in settings.py and wrapped panels.panels_pool.register(DummyPanel4) with if statement?

mitar:
Please update with latest from main repository.

piplmesh/panels/dummy4/panel.py
mitar: Yes!

mitar: You could also not define four different panels, but simply register four times inside the dummy panel. Of course not four times the same, but slightly modified dynamically created classes. ;-)

mitar:
Ping.

piplmesh/account/models.py
```diff
@@ -82,6 +82,9 @@ class User(auth.User):
     email_confirmed = mongoengine.BooleanField(default=False)
     email_confirmation_token = mongoengine.EmbeddedDocumentField(EmailConfirmationToken)
 
+    panels_collapsed = mongoengine.DictField()
```
mitar, line 88: You cannot use embedded documents here?

mitar:
Tried it out. It is getting better. :-)

I don't like the loading sign when you are resizing the window. This is just ugly. Why would that be useful? I know that I am resizing the window. :-) Just leave it displayed and rewrap the windows as you can, you can do also a very quick animation to move them around, if you want.

Don't tell me that you are doing GET Ajax requests to store state on the server?

Drag and dropping does not always work. Sometimes it simply does not allow dropping on columns.

For collapsing, add some icon on the edge of the title space, not just collapse on every click. Change mouse into dragging when you are over title space.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

mitar: And `var i`.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: What is the best way to remove hard-coded URL here? Constant variable?

Yes, I'm doing GET requests because POST requests doesn't work. CSRF doesn't allow POST requests. I don't know how to fix this.

Spaces fixed.

mitar: > Constant variable?

Yes.

For CSRF, read [documentation](https://docs.djangoproject.com/en/dev/ref/contrib/csrf/#ajax).

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: I've just removed this spinner.

piplmesh/frontend/templates/home.html
NejcZupec: Spinner removed.

piplmesh/panels/dummy4/panel.py
NejcZupec: Now I read DEBUG value from settings.

piplmesh/frontend/templates/home.html
NejcZupec: Removed.

NejcZupec:
I'll fix other comments in next days... ;)

mitar:
Will wait for that. :-)

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: I've changed string values with boolean values.

piplmesh/frontend/views.py
NejcZupec: I've changed string values with boolean values. Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: GET requests changed with POST requests.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Added / at the end of URL.

NejcZupec:
I've fixed almost everything, except:

**1) Hard-coded URLs**
What is the best way to remove hard-coded URL? Constants defined in settings.py? And how can I get them in JS script?

**2) Hiding div panels, while AJAX is moving panels around**
I'd like to hide div panels, but I don't know how. I've already tried to hide div and then show it back with CSS visibility value. But this is not working, because JavaScript is asynchronous. I've also tried callback functions. How can I fix that?

piplmesh/frontend/static/piplmesh/js/home.js
mitar: Still, why hard-coded URL?

mitar:
> What is the best way to remove hard-coded URL? Constants defined in settings.py? And how can I get them in JS script?

No, you don't need constants. Django can resolve them for you. You just name them in `urls.py` and then Django knows how to give you the URL. You then a global JavaScript variable in the template where you store that URL into a variable. The best is to define a dict:

```
        /* <![CDATA[ */
        var piplMeshAPI = {
            'panels': '{% filter escapejs %}{% url "panels" %}{% endfilter %}'
        };
        /* ]]> */
```

mitar:
> Hiding div panels, while AJAX is moving panels around

I don't understand this question.

piplmesh/panels/dummy/panel.py
```diff
@@ -9,8 +10,40 @@ class DummyPanel(panels.BasePanel):
...
         context.update({
             'header': _("Dummy panel"),
-            'content': '\n\n'.join(lorem_ipsum.paragraphs(random.randint(1, 3))),
+            'content': '\n\n'.join(lorem_ipsum.paragraphs(random.randint(1, 1))),
+            'id': 'dummy',
         })
         return context
 
-panels.panels_pool.register(DummyPanel)
+class DummyPanel2(panels.BasePanel):
```
mitar: Better, but could be done dynamically. So dynamically define those panels classes. :-)

mitar:
It looks nice. But still I cannot always drag and drop to some columns. If I move away, it sometimes does not allow to move to another column.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Hard-coded URLs are removed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/panels/horoscope/panel.py
```diff
@@ -17,6 +17,7 @@ def get_context(self, context):
 
         context.update({
             'header': _("Today's horoscope"),
+            'id': 'horoscope',
```
mitar: O no no. No no. No no no.

If you want something so repetitive, this should go to the panel's base class. And then this can be populated automatically.

piplmesh/frontend/views.py
```diff
@@ -98,3 +99,47 @@ def forbidden_view(request, reason=''):
...
+
+    user.panels_order[noOfColumns] = order
+    user.save()
+
+    return http.HttpResponse()
+
+def get_panels_order(request):
+    user = account_models.User.objects.get(id=request.user.id)
+    order = simplejson.loads(request.POST['data'])
+    noOfColumns = str(order['noOfColumns'])
```
mitar: We don't use camelCase in Python. Also, we use_underscore in JavaScript. [See the style](https://github.com/wlanslovenija/PiplMesh/pull/199#commitcomment-1707793).

piplmesh/frontend/views.py
```diff
@@ -98,3 +99,47 @@ def forbidden_view(request, reason=''):
...
+    message = request.POST['data']
+    order = simplejson.loads(message)
+    noOfColumns = str(len(order['panels']))
+
+    user.panels_order[noOfColumns] = order
+    user.save()
+
+    return http.HttpResponse()
+
+def get_panels_order(request):
```
mitar: I don't understand why you use two views here? Why you don't have one view, where on GET you return order. On POST you change it?

piplmesh/frontend/views.py
```diff
@@ -98,3 +99,47 @@ def forbidden_view(request, reason=''):
...
+    user = account_models.User.objects.get(id=request.user.id)
```
mitar: What is difference between your `user` and `request.user`? You already have user object in `request.user`. Why fetching it again?

piplmesh/panels/templates/panel/base_panel.html
```diff
@@ -1,4 +1,4 @@
-<div class="panel">
+<div class="panel" id="{{ id }}">
```
mitar: I would not use id, because there might be a conflict with some other XHTML id. So, use class (`class="panel {{ id }}") or prefix the id, like`id="panel-{{ id }}"`.

piplmesh/frontend/templates/home.html
```diff
@@ -13,32 +13,25 @@
         $.updates.subscribe({
             'home_channel': '{% filter escapejs %}{% channel_url HOME_CHANNEL_ID %}{% endfilter %}'
         });
+
+        var urls = {
+            'panels_collapse': '{% url frontend.views.panels_collapse %}',
```
mitar: JavaScript escape!

piplmesh/account/models.py
mitar, line 88: TODO.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
+function howManyColumns() {
+    var panelsWidth = $('#panels').width();
+    var columnPanelsWidth = $('.panels_column').width() + parseInt($('.panels_column').css('margin-left'));
```
mitar: Be careful. Computing sizes in one of the most non-cross-browser-compatibile thing there is. :-) This is why jQuery provides so many different methods do that. Maybe you should use one of them, like [outerWidth](http://api.jquery.com/outerWidth/)?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -575,14 +578,57 @@ input.gsc-search-button {
...
+    min-height: 500px;
+    width: 520px;
+    float: left;
+}
+
+#panels {
+    margin-left: 520px;
+}
+
+#spinner {
```
mitar: We don't use spinner anymore?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+}
+
+function movePanel(id, columnIndex) {
+    $('#' + id).appendTo($('#panels').children().eq(columnIndex));
+}
+
+function resetColumns() {
+    $('#panels').children().each(function (index, value) {
+        $(this).children().each(function (index, value) {
+            $(this).appendTo($('#panels').children().eq(0))
```
mitar: Uh, you are complicating here. And ... moving even bit too much around.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+
+    return parseInt(panelsWidth / columnPanelsWidth);
+}
+
+function movePanel(id, columnIndex) {
+    $('#' + id).appendTo($('#panels').children().eq(columnIndex));
+}
+
+function resetColumns() {
+    $('#panels').children().each(function (index, value) {
```
mitar: Rename `value` to the what you are iterating over (like `panel` in this case).

Do you have to do `$(this)`? Is not `$(this) == value`?

piplmesh/frontend/static/piplmesh/js/home.js
mitar: Anyway, use [deatch](http://api.jquery.com/detach/). And not append&remove.

So this could be really simplified. jQuery is great to work on sets of elements. So you should just find a selector, which selects all your panels and move them to the first column. So no need for for loops if you are working with jQuery. `.each` is mostly needed when you want to do something completely else, not connected with DOM manipulation.

Hint: [:gt](http://api.jquery.com/gt-selector/).

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+
+    var count = 0;
+    $('#panels').children().each(function (index, value) {
+        if (count != 0)
+            $(this).remove();
+        count++;
+    });
+}
+
+function fillWithColumns() {
```
mitar: Improve name.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+    var currentColumns = $('#panels').children().length;
+    var noOfColumns = howManyColumns();
```
mitar: Variable names.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+
+    for (i = currentColumns; i < noOfColumns; i++) {
```
mitar: `var i`!!!

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+        $('#panels').append('<div class="panels_column"></div>');
```
mitar: Don't use hard-coded html. Construct elements through jQuery.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+    }
+}
+
+function orderPanelsDefault() {
+    var numOfPanels = $('.panels_column').children().length;
+    var numOfColumns = howManyColumns();
+
+    for (var i = 0; i < numOfPanels; i++) {
+        var toColumn = i % numOfColumns;
+        $('.panels_column').children().eq(numOfPanels - i - 1).appendTo($('#panels').children().eq(toColumn));
```
mitar: Ehm, why you loop over number of panels and then access the indexed one? Why you just simply don't loop over panels themselves?

And appendTo moves the panel to new column? Or it duplicates it?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+}
+
+function orderPanelsUpdate() {
+    var items = [];
+
+    $('#panels').children().each(function (index, value) {
+        var column = [];
+        $(this).children().each(function (index, value) {
+            var item = {
+                id: $(this).attr('id'),
```
mitar: Use `prop` instead of `attr`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+    var numOfPanels = $('.panels_column').children().length;
+    var numOfColumns = howManyColumns();
+
+    for (var i = 0; i < numOfPanels; i++) {
+        var toColumn = i % numOfColumns;
+        $('.panels_column').children().eq(numOfPanels - i - 1).appendTo($('#panels').children().eq(toColumn));
+    }
+}
+
+function orderPanelsUpdate() {
```
mitar: I must say that I don't understand from function names what are they doing.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+        $(this).children().each(function (index, value) {
+            var item = {
+                id: $(this).attr('id'),
+            };
+            column.push(item);
+        });
+        items.push(column);
+    });
+
+    $.post(urls['panels_order'], 'data=' + JSON.stringify({panels: items}));
```
mitar: Why exactly you have to push this as a JSON?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+    }
+}
+
+function orderPanelsUpdate() {
+    var items = [];
+
+    $('#panels').children().each(function (index, value) {
+        var column = [];
+        $(this).children().each(function (index, value) {
+            var item = {
```
mitar: Why do you create a dict here? Why it cannot be simply a string? So that the result is a list of list of strings?

piplmesh/frontend/static/piplmesh/js/home.js
mitar: Why use `data=`? Decide, are you POSTing JSON, or you want to POST it as a normal POST.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+
+function orderPanels() {
+    $.post(urls['get_panels_order'], 'data=' + JSON.stringify( {noOfColumns: howManyColumns()} ), function (data) {
+        for (var i = 0; i < data['panels'].length; i++) {
+            for (var j = 0; j < data['panels'][i].length; j++) {
+                movePanel(data['panels'][i][j]['id'],i);
+            }
+        }
+
+        if (data['panels'].length == 0) {
```
mitar: Move this to the top, so if == 0, then do this, otherwise go into a loop. Code be much more readable in this way.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
+function howManyColumns() {
+    var panelsWidth = $('#panels').width();
+    var columnPanelsWidth = $('.panels_column').width() + parseInt($('.panels_column').css('margin-left'));
+
+    return parseInt(panelsWidth / columnPanelsWidth);
+}
+
+function movePanel(id, columnIndex) {
+    $('#' + id).appendTo($('#panels').children().eq(columnIndex));
```
mitar, line 12: This moves or it copies?

piplmesh/frontend/static/piplmesh/js/home.js
mitar: (Hm, OK, so append moves? And you use remove only for columns?)

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+    });
+
+    $.post(urls['panels_order'], 'data=' + JSON.stringify({panels: items}));
+}
+
+function orderPanels() {
+    $.post(urls['get_panels_order'], 'data=' + JSON.stringify( {noOfColumns: howManyColumns()} ), function (data) {
+        for (var i = 0; i < data['panels'].length; i++) {
+            for (var j = 0; j < data['panels'][i].length; j++) {
+                movePanel(data['panels'][i][j]['id'],i);
```
mitar: Space before `i`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+        });
+        items.push(column);
+    });
+
+    $.post(urls['panels_order'], 'data=' + JSON.stringify({panels: items}));
+}
+
+function orderPanels() {
+    $.post(urls['get_panels_order'], 'data=' + JSON.stringify( {noOfColumns: howManyColumns()} ), function (data) {
+        for (var i = 0; i < data['panels'].length; i++) {
```
mitar: Use jQuery `.each` to iterate over. You don't need to play with all those indexes. This is not so readable.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+            column.push(item);
+        });
+        items.push(column);
+    });
+
+    $.post(urls['panels_order'], 'data=' + JSON.stringify({panels: items}));
+}
+
+function orderPanels() {
+    $.post(urls['get_panels_order'], 'data=' + JSON.stringify( {noOfColumns: howManyColumns()} ), function (data) {
```
mitar: GET.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+
+        if (data['panels'].length == 0) {
+            orderPanelsDefault();
+        }
+    });
+}
+
+function collapsePanels() {
+    $.get(urls['get_panels_collapse'],'/panels/collapse/get/', function (data) {
+        for (var panel in data) {
```
mitar: Use jQuery `.each`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+        }
+
+        if (data['panels'].length == 0) {
+            orderPanelsDefault();
+        }
+    });
+}
+
+function collapsePanels() {
+    $.get(urls['get_panels_collapse'],'/panels/collapse/get/', function (data) {
```
mitar: Hard-coded URL. Spaces.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+        for (var panel in data) {
+            if (data[panel] == true)
+                $('#'+ panel +' .content').css('display','none');
```
mitar: Spaces.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+        if (data['panels'].length == 0) {
+            orderPanelsDefault();
+        }
+    });
+}
+
+function collapsePanels() {
+    $.get(urls['get_panels_collapse'],'/panels/collapse/get/', function (data) {
+        for (var panel in data) {
+            if (data[panel] == true)
```
mitar: `== true`???

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+                $('#'+ panel +' .content').css('display','none');
+        }
+    });
+}
+
+function preparePanels() {
```
mitar: Prepare? For what?

mitar: Initialize? Initialize panels layout?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+    fillWithColumns();
+    orderPanels();
+    collapsePanels();
+    makeColumnsSortable();
+    makePanelsUpdatable();
+}
+
+function makeColumnsSortable() {
+    $('.panels_column').sortable({
+        connectWith: '.panels_column',
```
mitar: Make key names string.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+        handle: '',
+        cursor: 'move',
+        placeholder: 'placeholder',
+        forcePlaceholderSize: true,
+        opacity: 0.6
+    }).disableSelection();
+}
+
+function makePanelsUpdatable() {
+    $('.panels_column').bind("sortstop", function (event, ui) {
```
mitar: Constant string.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
+    });
+}
+
 $(document).ready(function () {
+    preparePanels();
+
     $('.panel .header').click(function (event) {
+        var visible = $(this).next().is(':visible');
         $(this).next('.content').slideToggle('fast');
+        $.post(urls['panels_collapse'], 'data=' + JSON.stringify( {panel_id: $(this).parent().attr('id'), collapsed: (visible) ? true : false }));
```
mitar: Why `(visible)` and not `visible`? Wouldn't it be better if you would do this in the line above? Not in this long line?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,161 @@
...
     });
+
+    $(window).resize(function () {
+        resetColumns();
+        preparePanels();
+    });
+});
+
+$(document).ajaxSend(function(event, xhr, settings) {
```
mitar: Create `common.js` and move this there.

piplmesh/frontend/templates/base.html
```diff
@@ -12,6 +12,7 @@
         {% render_block "css" %}
         {% block js %}
             <script src="{% static "piplmesh/jquery/jquery.js" %}" type="text/javascript"></script>
+            <script src="{% static "piplmesh/jquery/jquery-ui.js" %}" type="text/javascript"></script>
```
mitar, line 16: You added just JavaScript and not also the CSS and icons?

mitar:
More work is needed.

piplmesh/frontend/static/piplmesh/css/style.css
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec, line 12: Moves. http://api.jquery.com/appendTo/

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/templates/home.html
NejcZupec: Fixed.

mitar:
Please tell me when you will fix everything. :-)

NejcZupec:
Ok. I need a few hours ;) I'll notify you.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Moved.

piplmesh/frontend/templates/base.html
NejcZupec, line 16: Added CSS file and images.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Now I just simply loop over panels ;) appendTo just moves the panel. Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: I don't understand this.

piplmesh/account/models.py
NejcZupec, line 88: I'd like to, but I don't know how. Where can I initialize this embedded document?

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: If I create dict here, I can simply save it to database on server side.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/views.py
NejcZupec: Huh.. I like more functions :)

piplmesh/frontend/views.py
NejcZupec: Because, I was young and stupid :) Fixed.

piplmesh/panels/templates/panel/base_panel.html
NejcZupec: Fixed.

piplmesh/frontend/views.py
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed. I have removed resetColumns() and now I'm using detach(). This function is exactly what I need :)

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Done.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/panels/horoscope/panel.py
NejcZupec: Hm, I think I am lost here :S

NejcZupec:
Ok, I think, I've fixed almost everything... actually, I hope so :)

piplmesh/frontend/static/piplmesh/js/home.js
mitar:
```
connectWith: '.panels_column',
```

into:

```
'connectWith': '.panels_column',
```

mitar:
> Fixed. Why is GET better than POST in this example?

Because it allows you to have the same view both for reading and updating the state. GET for reading, POST for updating.

piplmesh/panels/horoscope/panel.py
mitar: Do you think it is nice API which require programmer to manually add ID of the panel to the context? And if she forgets this, everything collapses?

So how would you make that ID is automatically in the context? Not that you had to change all panel code?

piplmesh/account/models.py
mitar, line 88: I don't understand what you don't know. Read documentation. Check examples. See how we have this.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: But you will not have dict's anymore in the database. Uahaha!

So please, simplify this.

piplmesh/account/models.py
```diff
@@ -82,6 +86,11 @@ class User(auth.User):
...
+    panels_order = mongoengine.DictField()
+
+    panels_settings = mongoengine.EmbeddedDocumentField(PanelsSettings)
```
mitar: This is a good approach. But why is there is still mongoengine.DictField() defined on user document?

piplmesh/account/models.py
```diff
@@ -44,6 +44,10 @@ def check_token(self, confirmation_token):
         else:
             return True
 
+class PanelsSettings(mongoengine.EmbeddedDocument):
+    panels_collapsed = mongoengine.DictField()
```
mitar: Please read about sets in MongoDB.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,119 @@
+function howManyColumns() {
+    var panelsWidth = $('#panels').width();
+    var columnPanelsWidth = $('.panels_column').outerWidth() + parseInt($('.panels_column').css('margin-left'));
```
mitar: TODO. Please read my comment from before.

mitar:
I will not read and comment until you fix everything what I have already commented. Things are still not good. So, JavaScript style, too much for loops, spaces, how you do HTTP requests, database schema. All needs polishing. Please make a comment when you fix all of those.

mitar:

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/panels/horoscope/panel.py
mitar: TODO.

piplmesh/frontend/views.py
```diff
@@ -3,8 +3,9 @@
 from django import http, template
 from django.conf import settings
 from django.contrib import messages
-from django.core import mail, urlresolvers
+from django.core import mail, urlresolvers, serializers
```
mitar: Where is this used?

piplmesh/frontend/templates/home.html
```diff
@@ -7,38 +7,30 @@
 {% block js %}
     {{ block.super }}
     <script src="{% static "pushserver/updates.js" %}" type="text/javascript"></script>
+    <script src="{% static "piplmesh/js/common.js" %}" type="text/javascript"></script>
```
mitar: The idea of curse is that you put this into base.

piplmesh/account/models.py
mitar: I don't like the document schema. I would do something like:

```
class PanelsSettings(mongoengine.EmbeddedDocument):
    collapsed = mongoengine.BooleanField()
    order = mongoengine.IntField()

...

panels_settings = mongoengine.MapField(mongoengine.EmbeddedDocumentField(PanelsSettings))
```

piplmesh/frontend/templates/home.html
NejcZupec: Fixed.

mitar:
Now you will have to update from main repository, because the file is already there. :-)

piplmesh/frontend/views.py
NejcZupec: It's not used anymore. Removed.

piplmesh/account/models.py
NejcZupec: And where can I save number of columns and id of a panel?

mitar: ID of the panel is key to mapfield.

Yes, I didn't think about the number of columns. So let us think a bit. What all should we have in panel settings:
- if panel is enabled or disabled (currently all panels are disabled)
- if enabled, it can be collapsed or not
- and moved around, where we store different position based on number of columns

The question is, if the position/order is really a panel setting or not?

NejcZupec: Can we have mapfield of mapfield? One for columns and the second one for IDs?

mitar: We could have. The questions is if it is reasonable. :-) I would like to have semantic structure. But yes, we could have order be a map from number of columns to order for this number of columns. But as I am asking myself (and you) is, is order really a part of settings? Probably yes. Maybe we can do it like that.

NejcZupec: Ok, but view functions will be very complicated :S

mitar: You don't have to send JSON to them. :-)

mitar: And you can propose a better semantic-full model.

NejcZupec: And why is this model so much better than mine? What are benefits?

mitar: Because MongoEngine checks what you are storing in the database. And on reading from database makes sure it is in this format. So it is a ages old question of typed or non-typed data. And I like typed (and semantically full) data more. You can define some other model, but please do not use free-form dicts.

NejcZupec: Can I talk with you via Skype? I have million of questions right now :S

mitar: I am barely awake. :-)

piplmesh/panels/map/panel.py
```diff
@@ -4,8 +4,11 @@
 
 class MapPanel(panels.BasePanel):
     def get_context(self, context):
+        context = super(MapPanel, self).get_context(context)
+        
         context.update({
             'header': _("Map"),
+            'id': 'map',
```
mitar: What is this?

piplmesh/panels/base.py
```diff
@@ -46,6 +46,10 @@ def get_template(self):
         return templates
 
     def get_context(self, context):
+        context.update({
+            'id': self.get_name()
```
mitar: Please rename `id` to `name`.

piplmesh/panels/map/panel.py
```diff
@@ -4,8 +4,11 @@
 
 class MapPanel(panels.BasePanel):
     def get_context(self, context):
+        context = super(MapPanel, self).get_context(context)
+
```
mitar, line 8: Decide, do you want empty lines or not after super call.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -589,17 +592,61 @@ button.browserid-signin:hover {
...
     height: 20px;
     text-align: center;
+    background-position: right;
+    background-color: #68AE2C;
+    cursor: move;
+}
+
+.panel .header:hover {
+    background: url("../images/maximize_icon.png") no-repeat;
+    background-position: 97% 50%;
```
mitar: 97%? Why exactly this value?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+    return parseInt(panelsWidth / columnPanelsWidth);
+}
+
+function movePanel(id, columnIndex) {
+    $('#' + id).appendTo($('#panels').children().eq(columnIndex));
+}
+
+function initializeEmptyColumnsForPanels() {
+    var currentColumns = $('#panels').children().length;
+    var noOfColumns = howManyColumns();
```
mitar: `noOfColumns` -> `numberOfColumns`

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+
+    for (var i = currentColumns; i < noOfColumns; i++) {
+        var newColumn = $(document.createElement('div'));
```
mitar: `$(document.createElement('div'));` -> `$('<div/>')`

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+        newColumn.addClass('panels_column');
+        $('#panels').append(newColumn);
+    }
+}
+
+function orderPanelsDefault() {
+    var numOfPanels = $('.panels_column').children().length;
```
mitar: Once you have `no`, then `num`. Please make it verbose and consistent.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+        var columns = $('#panels').children();
+
+        $(panel).appendTo(columns.eq(toColumn));
+    });
+}
+
+function sendOrderOfPanelsToServer() {
+    var items = [];
+
+    $('#panels').children().each(function (index, value) {
```
mitar: What is stored in `value`? Name the variable after that.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+        var column = [];
+        $(this).children().each(function (index, value) {
```
mitar: What is stored in `value`? Name the variable after that.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+            };
+            column.push(item);
+        });
+        items.push(column);
+    });
+
+    $.ajax({
+        type: 'POST',
+        url: urls['panels_order'],
+        data: {json: JSON.stringify(items)}
```
mitar: You are still using JSON?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+        type: 'GET',
+        url: urls['panels_order'],
+        data: {noOfColumns: howManyColumns()},
+        success: function (data) {
+            if (data['panels'].length == 0) {
+                orderPanelsDefault();
+            } else {
+                for (var i = 0; i < data['panels'].length; i++) {
+                    for (var j = 0; j < data['panels'][i].length; j++) {
+                        movePanel(data['panels'][i][j]['id'],i);
```
mitar: Space after `,`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+
+function orderPanels() {
+    $.ajax({
+        type: 'GET',
+        url: urls['panels_order'],
+        data: {noOfColumns: howManyColumns()},
+        success: function (data) {
+            if (data['panels'].length == 0) {
+                orderPanelsDefault();
+            } else {
```
mitar: Newline before `else`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+        url: urls['panels_order'],
+        data: {json: JSON.stringify(items)}
+    });
+}
+
+function orderPanels() {
+    $.ajax({
+        type: 'GET',
+        url: urls['panels_order'],
+        data: {noOfColumns: howManyColumns()},
```
mitar: Wrap.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+                }
+            }
+        }
+    });
+}
+
+function collapsePanels() {
+    $.get(urls['panels_collapse'], function (data) {
+        $.each(data, function (panelId, collapsed) {
+            if (collapsed == true) {
```
mitar: `(collapsed == true)` -> `(collapsed)`

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+        'cursor': 'move',
+        'placeholder': 'placeholder',
+        'forcePlaceholderSize': true,
+        'opacity': 0.6,
+        'helper': 'clone'
+    }).disableSelection();
+}
+
+function makePanelsOrderUpdatable() {
+    $('.panels_column').bind("sortstop", function (event, ui) {
```
mitar: Constant string.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+        var visible = $(this).next().is(':visible');
         $(this).next('.content').slideToggle('fast');
+
+        var panel_id = $(this).parent().attr('id');
+        var collapsed =  visible ? true : false;
+
+        $.ajax({
+            type: "POST",
+            url: urls['panels_collapse'],
+            data: {panel_id: panel_id, collapsed: collapsed}
```
mitar: Wrap.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,133 @@
...
+
 $(document).ready(function () {
+    initializePanels();
+
     $('.panel .header').click(function (event) {
+        var visible = $(this).next().is(':visible');
         $(this).next('.content').slideToggle('fast');
+
+        var panel_id = $(this).parent().attr('id');
+        var collapsed =  visible ? true : false;
```
mitar: This is really needed?

piplmesh/frontend/templates/base.html
```diff
@@ -8,10 +8,12 @@
         {% block extrahead %}{% endblock %}
         {% block css %}
             <link href="{% static "piplmesh/css/style.css" %}" rel="stylesheet" type="text/css" media="screen" />
+            <link href="{% static "piplmesh/jquery/ui-lightness/jquery-ui.css" %}" rel="stylesheet" type="text/css" media="screen" />
```
mitar, line 10: Move before `style.css` so that we can override things in `style.css`.

piplmesh/frontend/views.py
```diff
@@ -98,3 +99,39 @@ def forbidden_view(request, reason=''):
...
+            request.user.panels_collapsed[request.POST['panel_id']] = True
+        else:
+            request.user.panels_collapsed[request.POST['panel_id']] = False
+        request.user.save()
+
+        return http.HttpResponse()
+    else:
+        message = {}
+
+        for panel in request.user.panels_collapsed:
```
mitar: This code is really necessary? You are just creating a dict from ... dict?

piplmesh/frontend/views.py
```diff
@@ -98,3 +99,39 @@ def forbidden_view(request, reason=''):
...
+            message[panel] = request.user.panels_collapsed[panel]
+
+        return http.HttpResponse(simplejson.dumps(message), mimetype='application/json')
+
+def panels_order(request):
+    if request.method == 'POST':
+        panels = simplejson.loads(request.POST['json'])
+        no_of_columns = str(len(panels))
+
+        request.user.panels_order[no_of_columns] = panels
```
mitar: We said that you will be sending only changes? But you can send everything, just not as JSON.

piplmesh/panels/map/panel.py
NejcZupec: Forgot to remove :S Removed.

piplmesh/panels/base.py
NejcZupec: Renamed.

piplmesh/panels/map/panel.py
NejcZupec, line 8: Added empty lines.

piplmesh/frontend/static/piplmesh/css/style.css
NejcZupec: It looks nice :) What would be better?

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: I've decided for number. Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Wrapped.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: ?

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Wrapped.

piplmesh/frontend/templates/base.html
NejcZupec, line 10: Moved.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Because Django doesn't like other formats. I have a lot of troubles with sending array of arrays with post for example.

piplmesh/frontend/views.py
NejcZupec: :) I was really genius when I was writing this. Simplified.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: Constant strings should use `'`.

NejcZupec: Fixed.

piplmesh/frontend/views.py
mitar: You always have been a genius. :-)

mitar:

piplmesh/frontend/static/piplmesh/js/home.js
mitar: You should configure jQuery (globally, in `common.js`) to use [traditional type of serialization](http://api.jquery.com/jQuery.param/).

piplmesh/frontend/static/piplmesh/css/style.css
mitar: I am asking why. So this looks suspicious: or you can change the icon itself and store it like that, or something else is happening. So why exactly 3%? Not 10px? Or something?

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: So, this is the key: $.ajaxSettings.traditional = true;?

mitar: I do something like:

```
    $.ajaxSetup({
        'error': function (jqXHR, textStatus, errorThrown) {
            window.console.error(jqXHR, this, textStatus, errorThrown);
        },
        'timeout': 5000,
        'traditional': true
    });
```

mitar: I am not sure if you are allowed to override `ajaxSettings` directly.

NejcZupec: Ok, but this is still not good.

NejcZupec: I sent array like: [[panel_dummy],[panel_horoscope, panel-map], [panel-onlineusers]]
Than I read data: order = request.POST.getlist('order')
and I got this: [
            "panel-dummy",
            "panel-horoscope,panel-map",
            "panel-onlineusers"
              ]

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Off course not :) Removed.

piplmesh/frontend/static/piplmesh/css/style.css
NejcZupec: Fixed to 240px.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: You don't want to send 2D arrays. But you can unzip them and then zip them again.

jQuery:

```
{
    'name': ['a', 'b', 'c'],
    'column': [1, 2, 3],
    'order': [1, 2, 3]
}
```

POST:

```
name=a&name=b&name=c&column=1&column=2&column=3&order=1&order=2&order=3
```

Django:

```
zip(request.POST.getlist('name'), request.POST.getlist('column'),
request.POST.getlist('order'))
```

BTW, you should be sending panel names, without `panel-` prefix, this is something you are using in DOM, this should not go to the server.

mitar:

So, do you want me to pull this in or will you try to remove JSON and change the model? How much in a hurry are you?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -82,12 +194,28 @@ function showLastPosts(offset) {
...
 
     $('.panel .header').click(function (event) {
+        var visible = $(this).next().is(':visible');
         $(this).next('.content').slideToggle('fast');
+
+        var panel_id = $(this).parent().attr('id');
+        var collapsed =  visible;
+
+        $.ajax({
+            type: "POST",
```
mitar: Constant string.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -82,12 +194,28 @@ function showLastPosts(offset) {
...
+            url: URLS['panels_collapse'],
+            data: {'panel_id': panel_id, 'collapsed': collapsed}
```
mitar: Wrap.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,6 +1,118 @@
...
+        data: {
+            'order': JSON.stringify(order),
+            'number_of_columns': howManyColumns()
+        }
+    });
+}
+
+function orderPanels() {
+    $.ajax({
+        type: 'GET',
```
mitar: Quote keys as constant strings. `type` -> `'type'`.

mitar:
So, add TODOs and open tickets for things you will not do and link them here.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
NejcZupec: Fixed.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,6 +1,119 @@
...
+            'number_of_columns': howManyColumns()
+        }
+    });
+}
+
+function orderPanels() {
+    $.ajax({
+        'type': 'GET',
+        'url': URLS['panels_order'],
+        'data': {'numberOfColumns': howManyColumns()},
```
mitar: Wrap.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,6 +1,119 @@
...
+        'success': function (data) {
+            if (data['panels'].length == 0) {
```
mitar: `data['panels']` -> `data.panels`, and you can fix this everywhere, so indexing by constant string should be changed to data.index.

mitar:
You can check 9a689d6e1143a8061a79e305d0fa926caf55eb97 to see how some things could be cleaned.

PR #255, opened 2012-07-12
Sending pull request for Google weather layer

Gitek1:
1) Added google weather layer on existing google map.
2) Added checkbox for enable/disable this weather layer on map.

Signed-off-by: Gitek1 denis.zupanc@gmail.com


Gitek1:
Yes, you are right. I will fix all of that (change code formatting, add button on the map and delete one of the google maps api library). Tnx!

mitar:
> Yes, you are right. I will fix all of that (change code formatting, add button on the map and delete one of the google maps api library). Tnx!

I am not sure to what you are answering here?

piplmesh/panels/map/templates/panel/map/panel.html
```diff
@@ -19,6 +20,27 @@
             /* ]]> */
         </script>
     {% endaddtoblock %}
+    {% addtoblock "js" %}
+        <script type="text/javascript">
```
mitar, line 24: Move static JavaScript code to external file.

mitar:
So in fact this should not be modifying map panel. So this is not the right approach. Map panel should provide just a map. And then other panels should provide additional content. Which users will be able to turn on or off. So if user has a weather panel, then weather should be also visible on the map.

mitar:
Read discussion in #225 on this topic. I will probably prepare some code for this, because it seems to me that you will be stuck otherwise.

mitar:
You get few points extra for Signed-off-by. :-)

mitar:
Update from main repository.

mitar:
Closing in favor of #310.

PR #256, opened 2012-07-12
added test for get_closest_node, added get_closest_node, renamed RandomNodeBackend to NodeBackened

borismarn:
Added a few things:
      1. I think that class name RandomNodesBackend is not properly named... Not when you have methods get_source_node, get_closest_node & get node in there. Named it NodesBackend. I consider naming false even if all methods would return random node (and they dont and they didnt) .

```
  2. Because we still are dependendt on getting some random node i added function get random_node. For now get_sorce_node calls get random node and is documented that it should be fixed

  3. changed method get_closest_node so it returns closest node by iterating through all nodes and looking for closest. 

  4. Added test_get_closest_node that randomly checks 10 nodes if closest has been found
```

My first commit and will work on code even more.


piplmesh/account/templates/user/login.html
```diff
@@ -7,43 +7,52 @@
```
mitar, line 6: What is this? Why those files are changed?

piplmesh/account/templates/user/form/login-aut-form.html
```diff
@@ -0,0 +1,41 @@
```
mitar, line 0: What is this? Why those files are changed?

piplmesh/account/templates/user/registration.html
```diff
@@ -10,10 +10,10 @@
 {% block title %}{% trans "Registration" %}{% endblock %}
```
mitar, line 10: What is this? Why those files are changed?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -54,37 +54,61 @@ form input[type="submit"], form input[type="reset"] {
```
mitar, line 53: What is this? Why those files are changed?

mitar:
This pull request is completely something else than it is written it description. It has some logos for authentication providers. And nothing about node backends. I am closing this pull request. Please create correct one.

mitar:
And a comment to pull request description:

> I think that class name RandomNodesBackend is not properly named... Not when you have methods get_source_node, get_closest_node & get node in there

You seem to have problems understanding what is API and what is backend. API is interface which is exposed to the rest of the project. This interface defines some functions (`get_node` and `get_all_nodes). From the perspective of the project, this is all there is. Then, there are backends, which have their own internal API which they have to define for main API to be able to work. Those are methods`get_source_node`,`get_closest_node`,`get_node`, and`get_all_nodes`. And backend should define those methods. And hide internal implementation.

For testing purposes I have defined `RandomNodesBackend`, which randomly returns one of predefined nodes. The name is correct, it is backend which randomly returns a node.

In production, we will not use this backend. But backend which will read data from nodewatcher, our node database. But for local development such random backend is good enough. There is no use in improving it. Everything there is fake, static data, how it gets node, how it gets closest node. And this is on purpose. Because we do not need more complex stuff for local development.

mitar:
If you want, you can define a function which really returns a closest node for given location. And test for that. But the rest should be left as it is. It is not wrong, it is by design like that.

borismarn:
Well.. 

I wanted to learn some python and I saw some code I thought somebody should've repaired already... then I had some fun and sent a pull request. I talked to you on Skype afterwards and realised i shouldn't have messed with this code.

Ok i'll be glad to implement proper function that returns closest node.

Issue #244, opened 2012-06-12
Add contact form to the portal


mitar:
Probably somewhere in the footer should be a link to the contact form where users could contact administrators.

There should also be a notice that for bug reports and other issues they should use GitHub and link to the form for new GitHub issues. Of course, somewhere it could be also mentioned that it is developed as open source project and voluntarily, so they are also invited to help. (Could be through a link to about page, #243.)

PR #258, opened 2012-07-19
contact form

dcrystalj:
implemented issue #244 
https://github.com/wlanslovenija/PiplMesh/issues/244

### DIFF
--- a/piplmesh/frontend/forms.py
+++ b/piplmesh/frontend/forms.py
@@ -6,6 +6,7 @@ class ContactForm(forms.Form):
     """
     Class with contact form.
     """  
+    
     subject = forms.CharField(label=_("Subject"))
     email = forms.EmailField(label=_("E-mail"))
     message = forms.CharField(widget=widgets.Textarea())
--- a/piplmesh/frontend/forms.py
+++ b/piplmesh/frontend/forms.py
@@ -1,0 +1,11 @@
+from django import forms
+from django.forms import widgets
+from django.utils.translation import ugettext_lazy as _
+
+class ContactForm(forms.Form):
+    """
+    Class with contact form.
+    """  
+    subject = forms.CharField(label=_("Subject"))
+    email = forms.EmailField(label=_("E-mail"))
+    message = forms.CharField(widget=widgets.Textarea())
--- a/piplmesh/frontend/static/piplmesh/css/style.css
+++ b/piplmesh/frontend/static/piplmesh/css/style.css
@@ -591,12 +591,15 @@ input.gsc-search-button {
     margin: auto;
     margin-top: 50px;
 }
+
 .contact_form h1 {
     margin-left: 280px;
 }
+
 .contact_form  textarea{
     width: 300px;
 }
+
 .text p {
     margin-left: 100px;
 }
--- a/piplmesh/frontend/static/piplmesh/css/style.css
+++ b/piplmesh/frontend/static/piplmesh/css/style.css
@@ -585,3 +585,18 @@ input.gsc-search-button {
     background-color: #68AE2C;
     cursor: pointer;
 }
+/* ------------------ CONTACT ---------------------------- */
+.contact_form {
+    width: 960px;
+    margin: auto;
+    margin-top: 50px;
+}
+.contact_form h1 {
+    margin-left: 280px;
+}
+.contact_form  textarea{
+    width: 300px;
+}
+.text p {
+    margin-left: 100px;
+}
--- a/piplmesh/frontend/static/piplmesh/css/style.css
+++ b/piplmesh/frontend/static/piplmesh/css/style.css
@@ -596,7 +596,7 @@ input.gsc-search-button {
     margin-left: 280px;
 }
 
-.contact_form  textarea{
+.contact_form textarea {
     width: 300px;
 }
 
--- a/piplmesh/frontend/templates/contact.html
+++ b/piplmesh/frontend/templates/contact.html
@@ -13,9 +13,7 @@
         {% endwith %}
     </div>
     <div class="text">
-        <p>Feel free to email us at {{ mailo }}</p>
-        <p>If you have bug report or any other issue please use  <a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a>.<br />
-        This site is developed as open source project and all work is done voluntarily, so you are very welcome to help us make it even better. Check for more at <a href="{% url home %}about">{% trans "about us" %}</a>
-        </p>
+        <p>{% blocktrans %}If you have bug report or any other issue please use {% endblocktrans %}<a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a></p>
+        <p>{% blocktrans %}This site is developed as open source project and all work is done voluntarily, so you are very welcome to help us make it even better. Check for more at {% endblocktrans %}<a href="{% url about %}">{% trans "about us" %}</a></p>
     </div>
 {% endblock %}
--- a/piplmesh/frontend/templates/contact.html
+++ b/piplmesh/frontend/templates/contact.html
@@ -13,9 +13,8 @@
         {% endwith %}
     </div>
     <div class="text">
-        {% url about as about %}
-        {% blocktrans %}
-        <p>If you have bug report or any other issue please use <a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a></p>
-        <p>This site is developed as open source project and all work is done voluntarily, so you are very welcome to help us make it even better. Check for more at <a href="{{ about }}">about us</a></p>{% endblocktrans %}
+        {% url about as about_url %}
+        <p>{% blocktrans %}If you have bug report or any other issue please use <a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a>{% endblocktrans %}</p>
+        <p>{% blocktrans %}This site is developed as open source project and all work is done voluntarily, so you are very welcome to help us make it even better. Check for more at <a href="{{ about_url }}">about us</a>{% endblocktrans %}</p>
     </div>
 {% endblock %}
--- a/piplmesh/frontend/templates/contact.html
+++ b/piplmesh/frontend/templates/contact.html
@@ -1,0 +1,21 @@
+{% extends "plain.html" %}
+
+{% load i18n staticfiles %}
+
+{% block title %}{% trans "Contact" %}{% endblock %}
+
+{% block content %}
+    <div class="contact_form">
+        <h1>Contact us</h1>
+        {% with form_submit=_("Contact") %}
+            {% url contact as form_action %}
+            {% include "form/form.html" %}
+        {% endwith %}
+    </div>
+    <div class="text">
+        <p>Feel free to email us at {{ mailo }}</p>
+        <p>If you have bug report or any other issue please use  <a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a>.<br />
+        This site is developed as open source project and all work is done voluntarily, so you are very welcome to help us make it even better. Check for more at <a href="{% url home %}about">{% trans "about us" %}</a>
+        </p>
+    </div>
+{% endblock %}
--- a/piplmesh/frontend/templates/contact.html
+++ b/piplmesh/frontend/templates/contact.html
@@ -14,7 +14,7 @@
     </div>
     <div class="text">
         {% url about as about_url %}
-        <p>{% blocktrans %}If you have bug report or any other issue please use <a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a>{% endblocktrans %}</p>
-        <p>{% blocktrans %}This site is developed as open source project and all work is done voluntarily, so you are very welcome to help us make it even better. Check for more at <a href="{{ about_url }}">about us</a>{% endblocktrans %}</p>
+        <p>{% blocktrans %}If you have bug report or any other issue please use <a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a>.{% endblocktrans %}</p>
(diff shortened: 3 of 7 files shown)


piplmesh/frontend/forms.py
```diff
@@ -0,0 +1,11 @@
+from django import forms
+from django.forms import widgets
+from django.utils.translation import ugettext_lazy as _
+
+class ContactForm(forms.Form):
+    """
+    Class with contact form.
+    """  
+    subject = forms.CharField(label=_("Subject"))
```
mitar, line 10: Empty line after `"""`.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -585,3 +585,18 @@ input.gsc-search-button {
     background-color: #68AE2C;
     cursor: pointer;
 }
+/* ------------------ CONTACT ---------------------------- */
```
mitar, line 588: Style CSS similar to the rest of the file (empty lines around style blocks, spaces before `{`).

mitar:
Please merge from main repository changes there.

piplmesh/frontend/templates/footer.html
```diff
@@ -1,9 +1,9 @@
-{% load i18n %}
+{% load i18n staticfiles %}
```
mitar: Why adding `staticfiles` here?

piplmesh/frontend/templates/footer.html
```diff
@@ -1,9 +1,9 @@
...
 
 <div id="footer">
     <hr class="hidden" />
     <div id="footer_nav">
-        <a href="http://wlan-si.net/">wlan slovenija</a> | <a href="https://nodes.wlan-si.net/">{% trans "node database" %}</a> | <a href="http://dev.wlan-si.net/">{% trans "development" %}</a> | <a href="http://grow.wlan-si.net/">{% trans "growing" %}</a>
+        <a href="http://wlan-si.net/">wlan slovenija</a> | <a href="https://nodes.wlan-si.net/">{% trans "node database" %}</a> | <a href="http://dev.wlan-si.net/">{% trans "development" %}</a> | <a href="http://grow.wlan-si.net/">{% trans "growing" %}</a> | <a href="{% url home %}contact">{% trans "contact" %}</a>
     </div>
```
mitar, line 10: No, this is not correct way to of building URL to contact. Please read `url` template tag documentation. So don't stick manually home and contact together. This defeats the purpose of using the tag.

mitar, line 10: Create another `<div>` with list of footer links, currently only one link is there, contact (later on there will be also about and so on).

piplmesh/frontend/views.py
```diff
@@ -22,6 +28,33 @@ class OutsideView(generic_views.TemplateView):
...
+        try:
+            send_mail(subject, message, from_email, (settings.DEFAULT_FROM_EMAIL,), fail_silently=False)
+        except:
+            messages.error(self.request, _("Sorry, something went wrong here.")) 
+            return super(ContactView, self).form_valid(form)
+        messages.success(self.request, _("Thank you. Your message has been successfuly sent."))
+        return super(ContactView, self).form_valid(form)
+    
+    def form_invalid(self, form):
+        return super(ContactView, self).form_invalid(form)
```
mitar: You don't need to do this. If you just call parent's class method, then you simply don't define anything in this class. And only parent's class method will be called.

piplmesh/frontend/views.py
```diff
@@ -22,6 +28,33 @@ class OutsideView(generic_views.TemplateView):
...
+class ContactView(edit_views.FormView):
+    """
+    This view checks if all contact data are valid and then send mail.
+    
+    User is redirected to home page.
+    """
+    
+    template_name = 'contact.html'
+    # TODO: Redirect users to the page they initially came from
+    success_url = urlresolvers.reverse_lazy('home')
```
mitar: No, in this case redirect back to the contact form page.

piplmesh/frontend/views.py
```diff
@@ -1,7 +1,12 @@
 from django import http, template
 from django.conf import settings
+from django.contrib import messages
+from django.core import urlresolvers
 from django.core.files import storage
+from django.core.mail import send_mail, BadHeaderError
```

piplmesh/frontend/views.py
```diff
@@ -1,7 +1,12 @@
...
+from django.utils.translation import ugettext_lazy as _
 from django.views import generic as generic_views
+from django.views.generic import edit as edit_views
```
mitar: I think all views in `django.views.generic.edit` are accessible also through `django.views.generic`. Try it.

piplmesh/frontend/views.py
```diff
@@ -22,6 +28,33 @@ class OutsideView(generic_views.TemplateView):
...
+    """
+    
+    template_name = 'contact.html'
+    # TODO: Redirect users to the page they initially came from
+    success_url = urlresolvers.reverse_lazy('home')
+    form_class = forms.ContactForm
+
+    def form_valid(self, form):
+        subject = form.cleaned_data['subject'],
+        from_email = form.cleaned_data['email'],
```
mitar: Match variable name with dict key name (or form field name).

piplmesh/frontend/views.py
```diff
@@ -22,6 +28,33 @@ class OutsideView(generic_views.TemplateView):
...
+        message = form.cleaned_data['message'],
+        try:
+            send_mail(subject, message, from_email, (settings.DEFAULT_FROM_EMAIL,), fail_silently=False)
+        except:
```
mitar: Never ever catch all exceptions. Only catch those you expect. Only catch `smtplib.SMTPException` here.

piplmesh/frontend/views.py
```diff
@@ -22,6 +28,33 @@ class OutsideView(generic_views.TemplateView):
...
+    # TODO: Redirect users to the page they initially came from
+    success_url = urlresolvers.reverse_lazy('home')
+    form_class = forms.ContactForm
+
+    def form_valid(self, form):
+        subject = form.cleaned_data['subject'],
+        from_email = form.cleaned_data['email'],
+        message = form.cleaned_data['message'],
+        try:
+            send_mail(subject, message, from_email, (settings.DEFAULT_FROM_EMAIL,), fail_silently=False)
```
mitar: You can rather use `mail_managers` here. You don't have to set `fail_silently` to False, that is already default.

piplmesh/frontend/views.py
```diff
@@ -22,6 +28,33 @@ class OutsideView(generic_views.TemplateView):
...
+        except:
+            messages.error(self.request, _("Sorry, something went wrong here.")) 
+            return super(ContactView, self).form_valid(form)
+        messages.success(self.request, _("Thank you. Your message has been successfuly sent."))
```
mitar: You can move this immediately after `mail_managers` above. If `mail_managers` will raise exception, it will not be run. So you can simplify code here and simply call `super(ContactView, self).form_valid(form)` at the end.

piplmesh/frontend/templates/contact.html
```diff
@@ -0,0 +1,21 @@
...
+{% block content %}
+    <div class="contact_form">
+        <h1>Contact us</h1>
+        {% with form_submit=_("Contact") %}
+            {% url contact as form_action %}
+            {% include "form/form.html" %}
+        {% endwith %}
+    </div>
+    <div class="text">
+        <p>Feel free to email us at {{ mailo }}</p>
```
mitar: Why this line? `mailo`? I would just remove this line.

piplmesh/frontend/templates/contact.html
```diff
@@ -0,0 +1,21 @@
...
+        <p>If you have bug report or any other issue please use  <a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a>.<br />
```
mitar: You should put this into `blocktrans` template tags so that it can be translated. (Just leave anchor in the translation text.)

There is one double space just before `<a`.

piplmesh/frontend/templates/contact.html
```diff
@@ -0,0 +1,21 @@
...
+        This site is developed as open source project and all work is done voluntarily, so you are very welcome to help us make it even better. Check for more at <a href="{% url home %}about">{% trans "about us" %}</a>
```
mitar: You should create this URL better. This is not the correct way. (You will need to have a view for about page defined, which is not yet defined, but you can prepare an empty page for that.)

mitar:
Otherwise it looks good.

piplmesh/frontend/templates/contact.html
```diff
@@ -0,0 +1,19 @@
+{% extends "plain.html" %}
+
+{% load i18n staticfiles %}
```
mitar: No need for `staticfiles`?

piplmesh/frontend/templates/contact.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% block title %}{% trans "Contact" %}{% endblock %}
+
+{% block content %}
+    <div class="contact_form">
+        <h1>Contact us</h1>
+        {% with form_submit=_("Contact") %}
```
mitar: Is this really good label on the form? Maybe "Send"?

piplmesh/frontend/templates/contact.html
```diff
@@ -0,0 +1,19 @@
...
+            {% url contact as form_action %}
+            {% include "form/form.html" %}
+        {% endwith %}
+    </div>
+    <div class="text">
+        <p>{% blocktrans %}If you have bug report or any other issue please use {% endblocktrans %}<a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a></p>
+        <p>{% blocktrans %}This site is developed as open source project and all work is done voluntarily, so you are very welcome to help us make it even better. Check for more at {% endblocktrans %}<a href="{% url about %}">{% trans "about us" %}</a></p>
```
mitar: No, put also links inside of `blocktrans`.

mitar: Of course you will have to learn how to pass URLs from `url` into the blocktrans context.

piplmesh/frontend/views.py
```diff
@@ -1,6 +1,10 @@
 from django import http, template
 from django.conf import settings
+from django.contrib import messages
+from django.core import mail
+from django.core import urlresolvers
```
mitar: You can combine those two into one line.

piplmesh/frontend/views.py
```diff
@@ -22,6 +29,31 @@ class OutsideView(generic_views.TemplateView):
...
+    template_name = 'contact.html'
+    success_url = urlresolvers.reverse_lazy('contact')
+    form_class = forms.ContactForm
+
+    def form_valid(self, form):
+        subject = form.cleaned_data['subject'],
+        email = form.cleaned_data['email'],
+        message = form.cleaned_data['message'],
+        try:
+            mail.mail_managers(subject, message, email, (settings.DEFAULT_FROM_EMAIL,))
```
mitar: No need for `DEFAULT_FROM_EMAIL` part there, this argument is not needed when using `mail_managers`.

piplmesh/frontend/views.py
```diff
@@ -9,6 +13,9 @@
 
 from piplmesh.account import models as account_models
 from piplmesh.api import models as api_models, resources
+from piplmesh.frontend import forms
+
+import smtplib
```
mitar: Move this import to the top. It is system one.

mitar:
Otherwise it looks good now.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -585,3 +585,21 @@ input.gsc-search-button {
...
+    width: 960px;
+    margin: auto;
+    margin-top: 50px;
+}
+
+.contact_form h1 {
+    margin-left: 280px;
+}
+
+.contact_form  textarea{
```
mitar: Two spaces. One space missing before `{`.

piplmesh/frontend/templates/contact.html
```diff
@@ -0,0 +1,21 @@
...
+    <div class="contact_form">
+        <h1>Contact us</h1>
+        {% with form_submit=_("Send") %}
+            {% url contact as form_action %}
+            {% include "form/form.html" %}
+        {% endwith %}
+    </div>
+    <div class="text">
+        {% url about as about %}
+        {% blocktrans %}
```
mitar: No. <p> should stay outside the translated part. So put each paragraph inside of blocktrans, but links can be inside, because it could be language-specific where in the text the link should be. Don't have newlines in blocktrans.

piplmesh/frontend/templates/contact.html
```diff
@@ -0,0 +1,21 @@
...
+{% block content %}
+    <div class="contact_form">
+        <h1>Contact us</h1>
+        {% with form_submit=_("Send") %}
+            {% url contact as form_action %}
+            {% include "form/form.html" %}
+        {% endwith %}
+    </div>
+    <div class="text">
+        {% url about as about %}
```
mitar: `as about_url`. But yes, like this.

mitar:
Please write a comment when you update the pull request, because I don't get notification for that automatically.

dcrystalj:
updated

piplmesh/frontend/templates/contact.html
```diff
@@ -0,0 +1,20 @@
...
+    <div class="contact_form">
+        <h1>Contact us</h1>
+        {% with form_submit=_("Send") %}
+            {% url contact as form_action %}
+            {% include "form/form.html" %}
+        {% endwith %}
+    </div>
+    <div class="text">
+        {% url about as about_url %}
+        <p>{% blocktrans %}If you have bug report or any other issue please use <a href="https://github.com/wlanslovenija/PiplMesh/issues?state=open">GitHub</a>{% endblocktrans %}</p>
```
mitar: Dots at the end of sentences would be nice ...

dcrystalj:
oh...

dcrystalj:
anyway don't u have this http://shrani.si/f/13/AJ/3aj7sgqS/capture.png at the end?

mitar:
Yes. This means I get notifications on all comments. So everything you write. But if you just add a new commit to the pull request, I don't get anything. It is how GitHub works.

PR #261, opened 2012-08-01
Issue#254 BrowserID

cardawid:
Still working on it.. just that everyone see's im working on this..  Any help welcomed :)

### DIFF
--- a/piplmesh/account/templates/user/login.html
+++ b/piplmesh/account/templates/user/login.html
@@ -40,6 +40,16 @@
                             <p><button type="submit" class="google_login_button"><span class="google_button_text">{% trans "Login with Google" %}</span></button></p>
                         </form>
                     </li>
+                    <li>
+                        <a id="browserid" href="#">Sign In</a>
+                        <form method="post" action="{% url "browserid_verify" %}>
+                            {% csrf_token %}
+                            {{ browserid_form.as_p }}
+                        </form>
+                    </li>
+                    {% if user.is_authenticated %}
+                        Hello {{ user.email }}.
+                    {% endif %}
                     <li>{% comment %}Place for OpenID button.{% endcomment %}</li>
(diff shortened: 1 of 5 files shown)


mitar:
This will not work. django-browserid uses SQL models.

cardawid:
I was trying to do something like this:
https://developer.mozilla.org/en/BrowserID/Quick_Setup#Load_the_BrowserID_library

So is this the right way to implement browser id? If its not any documentation of feedback how and where to start would be useful...

mitar:
I am saying that for server-side you cannot use django-browserid library but will have to implement your own views.

mitar:
(You could also find a generic Python library or use just a subset of django-browserid. But django-browserid will not work simply out of the box, what is what you have been trying to do, according to current code.

cardawid:
Well i was thinking about creating own view... same as they are made for other login options... So with own view that will work ?

mitar:
You know, the idea is that you understand what you are doing. So you should understand how BrowserID works. Then you should make a plan what is all necessary to add to PiplMesh for it to be supported. Then you have to find which things you can reuse and which things you have to write yourself. None of this is just putting things together. If it would be, then you would not need to have a school for it. So, read read and read. Read documentation, read manuals and read existing code. And then create a solution.

So yes, maybe it will work with your views, maybe it will not. Also, it is not necessary that it will be less work with trying to adapt existing code to your needs. Sometimes it is easier to write from scratch. Sometimes it is not. Sometimes it is easier to find some other tool.

And the only way to learn this, to get feeling for this, is to try. To try one, try another, try something else. And then see what works best. And that's the point of all this project. Nobody in the job will tell you how to do it. They will tell you what to do. And then you will have to research existing solutions, read about them, read their code, combine them.

So, practice. Try one approach. And see how far it takes you.

mitar:
Done with #268.

PR #268, opened 2012-08-21
BrowserID integration.

Naltharial:
Added BrowserID sign in integration to the header and the login page. BrowserID package for Django using Mongo is at https://github.com/Naltharial/django-mongo-browserid. The links in the headers are getting a bit crowded (to say nothing of the login page), but I feel like the whole point of BrowserID is to put it on the front page.


mitar:
Ulalala.

One question, why do you think the BrowserID should be on the front page? And other external authentication mechanisms no?

piplmesh/frontend/templates/header.html
```diff
@@ -43,7 +44,12 @@
                     </li>
                 {% else %}
                     <li><a href="{% url "login" %}?{{ REDIRECT_FIELD_NAME }}={{ next_url }}">{% trans "Login" %}</a></li>
-                    <li class="last"><a href="{% url "registration" %}">{% trans "Register" %}</a></li>
+                    <li><a href="{% url "registration" %}">{% trans "Register" %}</a></li>
+                    <li class="last"><a id="browserid" class="browserid-signin" title="{% trans "Sign in with BrowserID" %}" aria-haspopup="true" target="_blank" href="#">{% trans "BrowserID" %}</a></li>
```
mitar: Is not "Persona" the now official public name for the project? BrowserID being just technology? So probably is should be "Login with Persona"?

mitar:
OK. Checked how it looks. And no. We will not have this as a special case on our front page. Then we can do the same with everyone else.

Which is maybe not a bad idea. We could do that when you hover over login, you get a pop-down menu with all choices. :-)

But this can be sugar on top for later (I will open a ticket). But for now make it one among all our external authentication mechanisms.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -172,6 +172,17 @@ form input[type="submit"], form input[type="reset"] {
     font-size: 10px;
 }
 
+a.browserid-signin:link, a.browserid-signin:visited {
+    background: url("../images/btn-browserid.png") no-repeat scroll -1px 50% #555555;
+    border: 1px outset rgba(50, 50, 50, 0.5);
+    border-radius: 3px 3px 3px 3px;
+    color: #FFFFFF !important;
```
mitar: Why `!important`? You should not be needing to use this?

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -172,6 +172,17 @@ form input[type="submit"], form input[type="reset"] {
...
+    padding: 1px 10px 1px 28px;
+}
+a.browserid-signin:hover, a.browserid-signin:focus, a.browserid-link:active {
```
mitar: Empty line before this line.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -2,4 +2,8 @@ $(document).ready(function () {
     $('.panel .header').click(function (event) {
         $(this).next('.content').slideToggle('fast');
     });
+    
+    $('button.auth_button').click(function (event) {
+    	navigator.id.logout();
```
mitar: Indentation (tabs)?

piplmesh/settings.py
```diff
@@ -145,6 +145,7 @@
     'django.contrib.messages.context_processors.messages',
     'sekizai.context_processors.sekizai',
     'piplmesh.frontend.context_processors.global_vars',
+    'django_mongo_browserid.context_processors.browserid_form',
```
mitar: What does this do?

piplmesh/urls.py
```diff
@@ -74,6 +74,9 @@
 
     # Internals
     url(r'^' + PUSH_SERVER_URL, include('pushserver.urls')),
+    
+    #BrowserID
```
mitar: Space after `#`. Add it a bit higher, where other authentication URLs are.

piplmesh/settings.py
Naltharial: It makes the appropriate `browserid_form` variable available in `RequestContext`. Not sure if we need it at the moment, but it might ease handling of `RequestContext` in the future, if we need to extend it with anything.

mitar: Aha. Yes, this is reasonable if you have button on every page. But because we will move this to login page, we do not need this anymore.

mitar:
Hm. Why I didn't get this notification. Sorry, will check it now.

piplmesh/frontend/templates/base.html
```diff
@@ -15,6 +14,7 @@
             <script src="{% url "django.views.i18n.javascript_catalog" %}" type="text/javascript"></script>
         {% endblock %}
         {% render_block "js" %}
+        {% block extrahead %}{% endblock %}
```
mitar: No, don't move extrahead. This is just for other meta options and stuff.

Use [sekizai](http://django-sekizai.readthedocs.org/en/latest/usage.html) for adding additional JavaScript.

piplmesh/settings.py
```diff
@@ -145,6 +145,7 @@
     'django.contrib.messages.context_processors.messages',
     'sekizai.context_processors.sekizai',
     'piplmesh.frontend.context_processors.global_vars',
+    'django_browserid.context_processors.browserid_form',
```
mitar, line 147: Move this one level up. Before our global vars. So that we can override in global vars if we will want.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -2,4 +2,8 @@ $(document).ready(function () {
     $('.panel .header').click(function (event) {
         $(this).next('.content').slideToggle('fast');
     });
+    
+    $('button.auth_button').click(function (event) {
+        navigator.id.logout();
```
mitar: Logout? Why logout?

piplmesh/account/templates/user/login.html
```diff
@@ -40,6 +45,13 @@
                             <p><button type="submit" class="google_login_button"><span class="google_button_text">{% trans "Login with Google" %}</span></button></p>
                         </form>
                     </li>
+                    <li>
+                        <form method="POST" action="{% url "browserid_verify" %}">
```
mitar: We write lower-case post.

piplmesh/account/templates/user/login.html
```diff
@@ -40,6 +45,13 @@
...
+                            {% csrf_token %}
+                            {{ browserid_form.as_p }}
```
mitar: Wrap this into one div to be more XHTML compliant.

mitar: In fact. You could move this form to external file and include it here. And in that external file you could use sekizai to add JavaScript files. So anywhere somebody would include that files, libraries would automatically be added to the output.

piplmesh/account/templates/user/login.html
```diff
@@ -1,9 +1,14 @@
 {% extends "plain.html" %}
 
-{% load future i18n %}
+{% load future i18n staticfiles %}
 
 {% block title %}{% trans "Login" %}{% endblock %}
 
+{% block extrahead %}
+	<script type="text/javascript" src="{% static "browserid/browserid.js" %}"></script>
```
mitar: Indentation. BTW.

piplmesh/account/templates/user/login.html
```diff
@@ -40,6 +45,13 @@
                             <p><button type="submit" class="google_login_button"><span class="google_button_text">{% trans "Login with Google" %}</span></button></p>
                         </form>
                     </li>
+                    <li>
+                        <form method="POST" action="{% url "browserid_verify" %}">
+                            {% csrf_token %}
+                            {{ browserid_form.as_p }}
+                           <p><a id="browserid" class="browserid-signin" title="{% trans "Sign in with Persona" %}" aria-haspopup="true" target="_blank" href="#">{% trans "Sign in with Persona" %}</a></p>
```
mitar: Don't use `href="#"` because it could change hash/anchor content which we could be using for something else. `target` is probably also not needed.

piplmesh/account/backends.py
```diff
@@ -6,12 +6,17 @@
 from mongoengine import queryset
 from mongoengine.django import auth
 
+from django_browserid.auth import default_username_algo, BrowserIDBackend as OfficialBIDB
```
mitar: Import only modules. Then you will not need such names as `OfficialBIDB`.

piplmesh/account/backends.py
```diff
@@ -260,6 +265,34 @@ def authenticate(self, foursquare_access_token, request):
...
+        return self.user_class.objects.filter(email=email)
+
+    def create_user(self, email):
+        username = default_username_algo(email)
+        return self.user_class.objects.create_user(username, email)
+    
+    def get_user(self, user_id):
+        return MongoEngineBackend.get_user(self, user_id)
+    
+    def authenticate(self, assertion=None, audience=None):
```
mitar: No no no. You should not try to copy logic of their implementation. We are not creating new users here. We already have a user created. A lazy user. We just link that lazy user to user's BrowserID. Setting e-mail address at the login. See how we are doing this in other backends.

piplmesh/account/backends.py
```diff
@@ -260,6 +265,34 @@ def authenticate(self, foursquare_access_token, request):
 
         return user
 
+class BrowserIDBackend(OfficialBIDB, MongoEngineBackend):
+    def filter_users_by_email(self, email):
+        return self.user_class.objects.filter(email=email)
+
+    def create_user(self, email):
+        username = default_username_algo(email)
+        return self.user_class.objects.create_user(username, email)
```
mitar: Furthermore, this so fails for me with `'QuerySet' object has no attribute 'create_user'`. :-) MongoDB model is different to Django original one. There is no `create_user` in its queryset.

mitar:

piplmesh/frontend/static/piplmesh/js/home.js
Naltharial: It's the logout button, we have to trigger BrowserID's (navigator object) logout method to tell it that the user is logging out. Otherwise, the user is logged in again on page refresh.

Naltharial:
Well, had to implement a custom view to send `request` across to get the lazy `User` object. Now tested login with a clean database, so a new user should work as well.

Switched BID frontend to a separate `account/user/browserid.html` template.

piplmesh/frontend/templates/base.html
```diff
@@ -5,7 +5,7 @@
         <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
         <link href="{% static "piplmesh/images/favicon.ico" %}" rel="shortcut icon" type="image/x-icon" />
         <link href="{% static "piplmesh/images/favicon.ico" %}" rel="icon" type="image/x-icon" />
-        {% block extrahead %}{% endblock %}
+         {% block extrahead %}{% endblock %}
```
mitar: You added some unnecessary space here?

piplmesh/frontend/static/piplmesh/js/home.js
mitar: And `button.auth_button` matches only the logout button?

piplmesh/account/views.py
```diff
@@ -219,6 +223,71 @@ def get(self, request, *args, **kwargs):
             # TODO: Message user that they have not been logged in because they cancelled the foursquare app
             # TODO: Use information provided from foursquare as to why the login was not successful
             return super(FoursquareCallbackView, self).get(request, *args, **kwargs)
+        
+class BrowserIDLoginView(edit_views.BaseFormView):
```
mitar: Why, oh why. Why you copy-paste code and change bit here and bit there? Why you don't embrace the beauty of object oriented programming and simply just extend django-browserid's `Verify` view in a way that request is passed?

Always when you start copying some chunk of the code from somewhere else to modify just one line you should stop and think what are you doing wrong.

piplmesh/account/templates/user/login.html
```diff
@@ -1,6 +1,6 @@
 {% extends "plain.html" %}
 
-{% load future i18n %}
+{% load future i18n staticfiles %}
```
mitar: Why `staticfiles`?

piplmesh/account/templates/user/login.html
```diff
@@ -41,6 +41,9 @@
                         <p><button type="submit" class="google_login_button"><span class="google_button_text">{% trans "Login with Google" %}</span></button></p>
                     </form>
                 </li>
+                <li>
+                    {% include "user/browserid.html" %}
```
mitar, line 45: Where is this file?

piplmesh/account/templates/user/login.html
```diff
@@ -41,6 +41,9 @@
...
+                </li>
 				{% comment %}<li>Place for OpenID button.</li>{% endcomment %}
```
mitar: Fix indentation here.

piplmesh/account/backends.py
```diff
@@ -6,12 +6,16 @@
 from mongoengine import queryset
 from mongoengine.django import auth
 
+from django_browserid import auth as bid_auth, base as bid_base
```
mitar: Spell it in full. Not only `bid`.

piplmesh/account/views.py
Naltharial: Hmm, I think it comes from the previous versions of the view, which were more significantly changed because the backend model was different. I've since changed the model, so I didn't have to change the view so much, but apparently didn't remember to inherit the functions that are now the same.

You're right, of course, and it's been fixed.

piplmesh/account/templates/user/login.html
Naltharial, line 45: Next to login.html, in `/piplmesh/account/templates/user/`.

Naltharial, line 45: Alternatively, I forgot to commit it. Well, _now_ it's next to `login.html`. :)

piplmesh/account/views.py
```diff
@@ -219,6 +222,29 @@ def get(self, request, *args, **kwargs):
...
+        """Handles the return post request from the browserID form and puts
+        interesting variables into the class. If everything checks out, then
+        we call handle_user to decide how to handle a valid user
+        """
+        self.assertion = form.cleaned_data['assertion']
+        self.audience = django_browserid.get_audience(self.request)
+        self.user = auth.authenticate(
+                assertion=self.assertion,
+                audience=self.audience,
+                request=self.request)
```
mitar: Please style it in the same way as we have it the rest of the file. Indentation and `)` at the end, for example.

piplmesh/account/templates/user/browserid.html
```diff
@@ -0,0 +1,16 @@
+{% load future i18n staticfiles sekizai_tags %}
+
+{% addtoblock "js" %}
+	<script type="text/javascript" src="{% static "browserid/browserid.js" %}"></script>
```
mitar: Indent with spaces, not tabs.

mitar: Add its script tag separately to the "js" block. Sekizai makes sure that block are not duplicated and it can do that only if there are of small chunks.

mitar: Furthermore, have the order of args in the script tag the same as in base.html file (first src, then type).

piplmesh/account/templates/user/browserid.html
```diff
@@ -0,0 +1,16 @@
...
+	<script type="text/javascript" src="https://browserid.org/include.js"></script>
+{% endaddtoblock %}
+
+{% block browserid %}
```
mitar: Ehm. No block here. You are not doing a parent template you are extending, but just a smaller piece of the template, you are including.

piplmesh/account/templates/user/browserid.html
```diff
@@ -0,0 +1,16 @@
...
+	<div id="browserid">
+		<form method="post" action="{% url "browserid_verify" %}">
+			{% csrf_token %}
+			{{ browserid_form.as_p }}
+			<p><button type="submit" id="browserid" class="browserid-signin" title="{% trans "Sign in with Persona" %}" aria-haspopup="true">{% trans "Sign in with Persona" %}</button></p>
```
mitar: That is not good. You have two different elements with the same ID (`browserid`).

piplmesh/account/backends.py
```diff
@@ -260,6 +264,47 @@ def authenticate(self, foursquare_access_token, request):
...
+            user.email = email
+            # BrowserID takes care of email confirmation
+            user.email_confirmed = True
+        
+        if user.lazyuser_username:
+            # Best username guess we can get from BrowserID
+            user.username = email.rsplit('@', 1)[0]
+            user.lazyuser_username = False
+        
+        user.browserid_issuer = result['issuer']
```
mitar: What all is available in the result?

piplmesh/account/backends.py
```diff
@@ -6,12 +6,16 @@
...
 
+from django_browserid import auth as bid_auth, base as bid_base
+
 import tweepy
 
 from piplmesh.account import models
 
 LAZYUSER_USERNAME_TEMPLATE = 'guest-%s'
 
+log = logging.getLogger(__name__)
```
mitar: You are not using that anymore?

piplmesh/account/backends.py
mitar: TODO.

piplmesh/account/backends.py
```diff
@@ -1,4 +1,4 @@
-import json, urllib
+import json, urllib, logging
```
mitar: No need for this anymore.

piplmesh/account/backends.py
```diff
@@ -260,6 +264,47 @@ def authenticate(self, foursquare_access_token, request):
...
+class BrowserIDBackend(bid_auth.BrowserIDBackend, MongoEngineBackend):
+    """
+    Persona authentication.
+
+    Persona profile data fields are:
+        email: email user uses for Persona
+    """
+    
+    def get_user(self, user_id):
+        return MongoEngineBackend.get_user(self, user_id)
```
mitar: What? What is this? Why not `super(BrowserIDBackend, self).get_user(user_id)`? What way of calling of parent implementation is this? This disables inheritance further one (what if somebody inherit the BrowserIDBackend class?

piplmesh/account/views.py
mitar: Please order arguments in the same order as they are defined in `authenticate`.

piplmesh/account/backends.py
```diff
@@ -260,6 +264,47 @@ def authenticate(self, foursquare_access_token, request):
...
+    
+    def authenticate(self, request=None, assertion=None, audience=None):
```
mitar: Change arguments name to `browserid_assertion`, `browserid_audience` and `request`, in this order.

piplmesh/account/views.py
```diff
@@ -219,6 +222,29 @@ def get(self, request, *args, **kwargs):
             # TODO: Message user that they have not been logged in because they cancelled the foursquare app
             # TODO: Use information provided from foursquare as to why the login was not successful
             return super(FoursquareCallbackView, self).get(request, *args, **kwargs)
+        
+class BrowserIDLoginView(browserid_views.Verify):
```
mitar: Why not `BrowserIDVerifyView`?

piplmesh/account/backends.py
```diff
@@ -260,6 +264,47 @@ def authenticate(self, foursquare_access_token, request):
...
+        result = bid_base.verify(assertion, audience)
+        if not result:
+            return None
+
+        email = result['email']
+
+        try:
+            user = self.user_class.objects.get(email=email)
+        except self.user_class.DoesNotExist:
+            # We reload to make sure user object is recent
```
mitar, line 297: Please add TODOs we have elsewhere.

piplmesh/account/backends.py
```diff
@@ -260,6 +264,47 @@ def authenticate(self, foursquare_access_token, request):
...
+            request.user.reload()
+            user = request.user
+
```
mitar, line 301: Add a TODO with a comment to check for multiple user objects returned. We have to do that check in all backends and decide what to do. But you do not have to decide that now. Just add a TODO.

piplmesh/account/backends.py
mitar: I would store this before those `if` sentences.

piplmesh/account/backends.py
```diff
@@ -260,6 +264,47 @@ def authenticate(self, foursquare_access_token, request):
...
+        except self.user_class.DoesNotExist:
+            # We reload to make sure user object is recent
+            request.user.reload()
+            user = request.user
+        
+        if user.email is None:
+            user.email = email
+            # BrowserID takes care of email confirmation
+            user.email_confirmed = True
+
```
mitar, line 303: No need for this empty line.

piplmesh/account/backends.py
Naltharial: Almost nothing.

```
{
    u'audience': u'https://mysite.com:443',
    u'email': u'myemail@example.com',
    u'issuer': u'browserid.org',
    u'status': u'okay',
    u'expires': 1311377222765
}
```

mitar: But still, store all of this into the database. Like we are doing in other backends. For example, expires could be helpful some day.

Naltharial: Should I store `expires` as a `DateTimeField` and `status` as a custom field, limited to `okay` and `failure`, or use `StringField` for either? Because we did define custom field for gender, but it's almost too specific to define one just for BrowserID success value?

mitar: No, just store it in a same way as we store values for other backends.

piplmesh/account/backends.py
mitar: What if you just change the order of classes you inherit from? :-)))

mitar: And on StackOverflow it is written:

> If you want to do sane multiple inheritance, you need everyone to be calling super

And read comments under the second answer.

Naltharial: Well, both the official `BrowserIDBackend` and our `MongoEngineBackend` contain `get_user()`, so `super()`, I think, would call only the official implementation because of the Method Resolution Order (MRO) of Python's multiple inheritance.

I've changed the order to be

```
class BrowserIDBackend(MongoEngineBackend, browserid_auth.BrowserIDBackend):
```

since that seems to have no such ambiguities.

Naltharial: Heh, you respond fast, before I can change myself finding the error. :)
Yes, this seems to work. I admit I'm still wrapping my head around Python's multiple inheritance, heh.

mitar: Yes. Just make sure that we don't override any method of `BrowserIDBackend` in our `MongoEngineBackend` by accident.

mitar: > Heh, you respond fast

Sorry. I think that you have already spend some time on the subject before commenting here and that you are commenting here only when you are stuck.

Naltharial:
Ok, this should be the changes up to now commited. Still works, new user or previously authenticated, now with a properly OOP-ed backend.

piplmesh/account/backends.py
```diff
@@ -1,4 +1,4 @@
-import json, urllib
+import json, urllib, datetime
```
mitar: `datetime` is not needed?

piplmesh/account/backends.py
```diff
@@ -260,6 +262,49 @@ def authenticate(self, foursquare_access_token, request):
...
+class BrowserIDBackend(MongoEngineBackend, browserid_auth.BrowserIDBackend):
+    """
+    Persona authentication.
+
+    Persona profile data fields are:
+        email: email user uses for Persona
+    """
+    
+    def get_user(self, user_id):
+        return super(BrowserIDBackend, self).get_user(user_id)
```
mitar: This is not needed. If you don't define a method, then first one found in parent classes will be used. Exactly the same as you are doing here.

piplmesh/account/models.py
```diff
@@ -81,6 +81,8 @@ class User(auth.User):
 
     email_confirmed = mongoengine.BooleanField(default=False)
     email_confirmation_token = mongoengine.EmbeddedDocumentField(EmailConfirmationToken)
+    
+    browserid_data = mongoengine.DictField()
```
mitar: Can you move it up, after other authentication backends data?

mitar: Name it `browserid_profile_data`.

piplmesh/account/models.py
```diff
@@ -98,7 +100,12 @@ def is_anonymous(self):
 
     def is_authenticated(self):
         # TODO: Check if *_data fields are really false if not linked with third-party authentication
-        return self.has_usable_password() or self.facebook_profile_data or self.twitter_profile_data or self.google_profile_data or self.foursquare_profile_data
+        return self.has_usable_password() or \
+                self.facebook_profile_data or \
```
mitar: Indent only once, not twice.

piplmesh/account/templates/user/browserid.html
```diff
@@ -0,0 +1,16 @@
+{% load future i18n staticfiles sekizai_tags %}
+
+{% addtoblock "js" %}
+    <script src="{% static "browserid/browserid.js" %}" type="text/javascript"></script>
```
mitar: No newline around the tag. Check how we are doing it elsewhere.

mitar: (And no spaces.)

Because duplicate elimination will not work if there is any space or newline difference.

piplmesh/account/templates/user/browserid.html
```diff
@@ -0,0 +1,16 @@
...
+{% endaddtoblock %}
+{% addtoblock "js" %}
+    <script src="https://browserid.org/include.js" type="text/javascript"></script>
+{% endaddtoblock %}
+
+<div id="browserid_form">
+    <form method="post" action="{% url "browserid_verify" %}">
+        {% csrf_token %}
+        {{ browserid_form.as_p }}
```
mitar: Please check if form is XHTML valid.

mitar: (You can use [HTML validator](http://users.skynet.be/mgueury/mozilla/) for this.)

piplmesh/account/backends.py
```diff
@@ -260,6 +262,49 @@ def authenticate(self, foursquare_access_token, request):
...
+    
+    def authenticate(self, browserid_assertion=None, browserid_audience=None, request=None):
+        result = browserid_base.verify(browserid_assertion, browserid_audience)
+        if not result:
+            return None
+
+        email = result['email']
+
+        try:
+            user = self.user_class.objects.get(email=email)
```
mitar: No, this is not OK. You should match on the email in `user.browserid_profile_data`. It is not necessary that user has the same e-mail used for our portal than she is using for BrowserID. Check how other backends match.

piplmesh/account/backends.py
```diff
@@ -260,6 +262,49 @@ def authenticate(self, foursquare_access_token, request):
...
+            # TODO: Based on user preference, we might create a new user here, not just link with existing, if existing user is lazy user
+            # We reload to make sure user object is recent
+            request.user.reload()
+            user = request.user
+            # TODO: Is it OK to override BrowserID link if it already exist with some other BrowserID user?
+        
+        user.browserid_data = result
+        
+        if user.email is None:
+            user.email = email
```
mitar: `user.email = email or None`

Not really possible otherwise, but still, it is important that it is always `None` and not empty string.

piplmesh/account/backends.py
```diff
@@ -260,6 +262,49 @@ def authenticate(self, foursquare_access_token, request):
...
+            # BrowserID takes care of email confirmation
+            user.email_confirmed = True
+        if user.lazyuser_username:
+            # Best username guess we can get from BrowserID
+            user.username = email.rsplit('@', 1)[0]
```
mitar, line 306: That is in fact a good idea. Can you do the same for Foursquare and Google backends?

And move this `if` part to the beginning (before e-mail). So that all are always in the same order (and the order of values in the model definition).

mitar:
You do know that your login button is not of the same size as others? Maybe you don't have the latest version. It does not matter. Probably somebody else will fix this. Just to let you know.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: One more thing. This should not be only in `home.js`, but on all pages where logout is displayed.

mitar: It also does not work. I am getting:

```
TypeError: navigator.id is undefined

navigator.id.logout();
```

piplmesh/account/templates/user/browserid.html
Naltharial: It's not because of `<input>` inside of `<form>` as `browserid_form` puts it, and `<input>` cannot be a child of `<form>`, because you need a block-level element. However, `browserid.js` runs:

```
var $e = $('#id_assertion');
$e.parent().submit();
```

so it expects the `<form>` to be the direct parent of `<input>`. I've worked around it by wrapping it in `<div>` and attaching a submit listener to it, so when `browserid.js` submits it, it automatically submits the form. Is there a better way, without modifying `browserid.js`?

Naltharial:
I've changed BrowserID button's CSS to be more like the rest of the login buttons.

piplmesh/account/backends.py
```diff
@@ -246,6 +252,10 @@ def authenticate(self, foursquare_access_token, request):
         user.foursquare_access_token = foursquare_access_token
         user.foursquare_profile_data = foursquare_profile_data
 
+        if user.lazyuser_username:
+            # Best username guess we can get from Foursquare
+            user.username = foursquare_profile_data.get('contact', {}).get('email').rsplit('@', 1)[0]
```
mitar: Not good. What it there is not e-mail information? You should add to `if` above test if we really get some username value out of it.

```
if user.lazyuser_username and foursquare_profile_data.get('contact', {}).get('email', '').rsplit('@', 1)[0]:
```

Of course you could store this into some variable before if. Like `username_guess`.

piplmesh/account/templates/user/browserid.html
```diff
@@ -0,0 +1,19 @@
+{% load future i18n staticfiles sekizai_tags %}
+
+{% addtoblock "js" %}<script src="{% static "browserid/browserid.js" %}" type="text/javascript"></script>{% endaddtoblock %}
+{% addtoblock "js" %}<script src="https://browserid.org/include.js" type="text/javascript"></script>{% endaddtoblock %}
+{% addtoblock "js" %}<script type="text/javascript">
+    $(document).ready(function () {
```
mitar: Why put this into template and not into `browserid.js`?

mitar: It is ugly but good way to solve it. I don't have any better suggestion.

Naltharial: `browserid.js` is generated automatically by `django_browserid`, so it's 3rd party.

mitar: OK. I see. Then you will have to create your own external file.

Naltharial:
Ok, I've added BrowserID's javascript to `base.html` so we can call `navigator.id.logout()` from anywhere.

Naltharial:
Changed class name in template, updated CSS to new name. Doesn't look like it appears anywhere else.

PR #270, opened 2012-08-23
Issue#269 Login drop-down menu

igypigi:
Added login drop-down menu

PR #271, opened 2012-08-23
Added basic hover overlay for login options.

Naltharial:
Added a very basic overlay to the Login link in the header. Connected to the BrowserID changes, but sufficiently distinct that they should probably be evaluated independantly.

Mostly just added jQuery handlers to display an overlay when hovering over Login and a bisic template in account/ajax/login.html, because we don't want the whole site to display, just the content.


mitar:
I am not sure why is Ajax needed here? Why do you have to load login options through it?

Naltharial:
Well, we don't entirely need it just for this overlay, but then you would have to add ajax/login.html template into the base template and hide the div. I think that would reduce "cleanness" of templating. Maybe not as much with just this one, but if any other overlays are added later, you would need to manually include every overlay template and hide all the `<div>`s. So I think Ajax is the cleanest version to be able to add overlays later or change existing ones without having to change base frontend files.

mitar:
> but then you would have to add ajax/login.html template into the base template and hide the div.

I think this is completely normal and usual way to do it. Exactly because of this you split templates into multiple files, so that you can include the same file at different places. And then you wrap that into div, so that you can style them differently, if needed.

> but if any other overlays are added later, you would need to manually include every overlay template and hide all the divs. 

So? Then you just define one big div which includes all hidden layers.

Not for our case, when site is very dynamic, but this is in general better for browsers with bad JavaScript support or no CSS. It still displays the menu. With Ajax, you would lose that content. As I said, this is not really important in our case, as everything is dynamic. But just as a general argument why is nothing wrong with thi.

Ajax version has a delay and unnecessary complexity. Templates were made exactly for this type of situations, where you want to use same output on multiple locations. To deduplicate HTML you would have to do in, for example, static HTML files.

So no, Ajax is completely unneeded here. It does not solve the problem you are describing. Because other approaches exist exactly for this use case.

mitar:

Naltharial:
Well, since Ajax was inappropriate, I implemented it with templates. Since the popups might sometimes need a more complex logic, I've implemented a custom template tag to allow including views as popups. Since external context is pushed through, we can use `sekizai` to add to base blocks from popups.
All that is needed in base template is:

```
<% popup <name of view> %>
```

and the application creates the view, renders it and returns the content.

It might be a bit complex for the task, but I wanted to experiment with some of the more powerful Django tools so that we can implement almost anything as popups if needed.

mitar:
This was done much more simpler with #270. This is an interesting approach, but also quite broken. Will comment more.

mitar:
In general, you are mixing flows. First we have views, which prepare context, and then template rendering. And you in template rendering again go to views. This is not a clean design. What you can do is tools provided in templates to break them in parts (inheritance or inclusion). What was done in #270.

So in general mixing control logic with template logic is bad. It is much harder to debug. So we should not have magical template tags, for example. It is possible to do everything in template tags, but then we have PHP and not Django. :-)

piplmesh/frontend/static/piplmesh/js/common.js
```diff
@@ -1,4 +1,22 @@
 $(document).ready(function () {
+	$("#login_item").hover(function(e){
+    	$("#overlay .login").show();
+		$("#overlay").fadeIn(150);
+		
+		// Close on the X icon; .live() monitors changes to the DOM
+		$("#overlay .close").live('click', function(){
```
mitar, line 7: `.on` is currently recommended.

piplmesh/popups/templatetags/popup.py
```diff
@@ -0,0 +1,29 @@
+from django import template
+
+from piplmesh.popups import views
+
+register = template.Library()
+
+class PopupNode(template.base.Node):
```
mitar, line 7: You could use [simple_tag](https://docs.djangoproject.com/en/dev/howto/custom-template-tags/#simple-tags) decorator. For this template tag there is no need to do custom `Node` class.

piplmesh/popups/templatetags/popup.py
```diff
@@ -0,0 +1,29 @@
...
+@register.tag('popup')
+def do_popup(parser, token):
+    """
+    Finds the view we need to include in the popup and passes it to the PopupNode.
+    """
+    
+    bits = token.split_contents()[1:]
+    
+    if hasattr(views, bits[0]):
+        return PopupNode(getattr(views, bits[0]))
```
mitar, line 27: You hardcode here views. If you would really want to do this, it would be much better to require whole Python path to the view where you would then here import the module and use that view.

piplmesh/popups/views.py
```diff
@@ -0,0 +1,29 @@
+from django import template
+from django.contrib.auth import forms as auth_forms
+from django.views.generic import edit as edit_views
+
+class ExternalContextMixin():
```
mitar, line 5: Really interesting way to put context back into the view. Really crazy ideas. :-) (In a positive way. Sometimes it is interesting to see such solutions because I would never imagine them because they are so out of the "right way". But I can see that this could maybe be of use someday. But for current problem this is really a hammer to simple problem of including template.)

piplmesh/urls.py
```diff
@@ -40,6 +40,8 @@
     url(r'^register/$', account_views.RegistrationView.as_view(), name='registration'),
     url(r'^login/$', 'django.contrib.auth.views.login', {'template_name': 'user/login.html'}, name='login'),
     url(r'^logout/$', account_views.logout, name='logout'),
+    
+    url(r'^ajax/login/$', 'django.contrib.auth.views.login', {'template_name': 'ajax/login.html'}, name='ajax/login'),
```
mitar, line 44: This should be cleaned.

piplmesh/popups/templatetags/popup.py
```diff
@@ -0,0 +1,29 @@
...
+
+register = template.Library()
+
+class PopupNode(template.base.Node):
+    def __init__(self, popup):
+        self.popup = popup
+
+    def render(self, context):
+        view_function = self.popup.as_view(context=context)
+        template_response = view_function(context['request'])
```
mitar, line 13: See. This is the main problem of your approach. View functions are wrapped in middleware. You don't run them here. So there is some pre and post processing possible. What if author of the view function counts on that?

mitar:
The problem is that it works also only with class-based views.

mitar:
Anyway, an interesting approach. And some ingenious solutions. You made this yourself? Because this is some hard stuff. If you did it yourself, I will give you some points for it.

But sorry, this will not go in.

Naltharial:
Heh, it's ok, I knew it was something of an experiment to begin with. At least I learned quite a bit about views in the process.

Yes, I did it myself. I thought I was stuck at one point
http://stackoverflow.com/questions/12262654/django-view-not-updating-context-properly/
but then I figured out that me injecting views in the middle of template processing was exactly the problem. So I injected the context into it as well, and it worked. Well, for this current view at least, unfortunately a lot of django views have differing implementations of `get_context_data()`.

PR #272, opened 2012-08-26
Site layout

borismarn:
Hi

This is my attempt and first step to create a two column layout for the site, with the #main div having a fluid width. Is this something that would fit in the sites vision?


mitar:
Sorry. There is no way I will pull those 9 commits in! 8 of those are pure garbage. Please do only clean pull requests.

And about the pull request itself. Have you checked [layout draft here](http://dev.wlan-si.net/wiki/PiplMesh)? We will have wall on the left (main content for now) and panels on the right.

But those changes to layout are done already by two other pull requests (#199 and #252). So I am closing this.

Issue #273, opened 2012-08-28
Improve User account/profile design


borismarn:
I hope I haven't overlook but I haven't seen any issues/pull requests regarding user account/profile page design. I think there is some room for css/html improvements in account.html, user.html... Like replacing class="registration" on divs that are not about registration, changing menu div to menu ul, changing the style of #menu_list (so that it wouldn't need position:absolute), titles (h1) and few other things.

I've made a branch with few changes in this direction, still not finished though.
https://github.com/BorisMarn/PiplMesh/commit/30387c9d5acb8be24a7c4d0d3d08b1ff4a4d052e

Would this be something worth doing?

mitar:
Yes. That could be improved. Especially that menu could be moved out of the template to a separate file. Or maybe even a new base template file could be made which would have account menu at the side, and then all account-related pages could extend that one.

PR #274, opened 2012-08-30
Improve User account/profile design

borismarn:
This is my take on more coherent user account/profile design (Issue #273). I'm still not sure about the buttons since I don't know how the final version of the main page will look (and the rest for that matter), but they are easy to change.

I used jquery for more dynamic/better alignment of form elements. But maybe it would be better to align everything to the left and just accept the gap between the labels and input.

I've also moved the user account/profile menu to its own template as suggested.

PS: I don't know why the pull request includes the "revert to current upstream" commit. Anybody knows an easy fix?


mitar:
Hm. I really don't like the idea of you having that first commit here. You should really work on clean repository.

piplmesh/account/templates/user/account.html
```diff
@@ -1,23 +1,26 @@
 {% extends "plain.html" %}
 
-{% load future i18n %}
+{% load future i18n staticfiles %}
 
 {% block title %}{% trans "Account" %}{% endblock %}
 
+{% block js %}
+    {{ block.super }}
+    <script src="{% static "piplmesh/js/forms.js" %}" type="text/javascript"></script>
```
mitar, line 9: Hm. You could just move this to some `common.js` which would be included from all templates (in `base.html`).

piplmesh/account/templatetags/account.py
```diff
@@ -10,3 +10,10 @@ def user_image(context, user=None):
     return {
         'user_image_url': user.get_image_url(),
     }
+
+@register.simple_tag
+def active(request, pattern):
```
mitar: No, use [startswith](http://django-missing.readthedocs.org/en/latest/templatetags.html#missing.templatetags.string_tags.startswith) tag. No need for complex regex.

piplmesh/account/templates/user/account.html
mitar, line 9: Or maybe you can leave it for now and we will optimize this later.

piplmesh/frontend/static/piplmesh/js/forms.js
```diff
@@ -0,0 +1,9 @@
+$(document).ready(function () {  
+    var max = 0;  
+    $("label").each(function(){
```
mitar: Static strings should be with `'`. Put spaces around `()` in anonymous functions. Please specify all arguments in JavaScript functions which function can take, so that later on is visible what is available.

piplmesh/frontend/static/piplmesh/js/forms.js
```diff
@@ -0,0 +1,9 @@
+$(document).ready(function () {  
+    var max = 0;  
+    $("label").each(function(){  
+        if ($(this).width() > max)  
+            max = $(this).width();     
+    });  
+    $("label").width(max);  
+    $("form .buttons, .errorlist, #password_change, .contact_comment p, .login h1, .registration h1, .contact_form h1").css({ "margin-left": max });
```
mitar: That is too hard-coded. Please define a class, something like "automatic-width" which you then match and configure.

piplmesh/account/templatetags/account.py
mitar: But good idea to do it with a tag.

piplmesh/account/templates/user/menu.html
```diff
@@ -0,0 +1,5 @@
+{% load future i18n account %}
+<ul id="menu_list">
+    <li class="{% active request "/user/" %}"><a href="{% url "profile" username=user.username %}">{% trans "Profile" %}</a></li>
```
mitar: Don't hardcode `/user/`. You can store output of `url` into a variable and then check if current request path_info starts with that, if it does, you add CSS class.

borismarn:
I removed the unwanted commits, but I had to rewrite the commit history a bit.

I've also tried to optimize the active item matching, fixed the enclosure on static strings and added automatic-width class as suggested.

mitar:
Can you write on our EtherPad how you cleaned the pull request?

piplmesh/account/templates/user/email_confirmation_process_token.html
```diff
@@ -5,14 +5,12 @@
...
-    <div id="menu_list">
-        <p><a href="{% url "profile" username=user.username %}">{% trans "Profile" %}</a></p>
-        <p id="current_item"><a href="{% url "account" %}">{% trans "Account" %}</a></p>
-    </div>
-    <p>
-        {% blocktrans with email=user.email %}You are just one step away from confirming your e-mail address. If for some reason the input box below is empty, paste the token in manually from the confirmation e-mail you have received.{% endblocktrans %}
-    </p>
-    <div class="registration" id="email_confirmation_process_token">
+    {% include "user/menu.html" %}
+    <div class="account" id="email_confirmation_process_token">
```
mitar: Decide if you have first `class` and then `id` attribute or opposite. But not mix. ;-)

piplmesh/account/templates/user/account.html
```diff
@@ -1,23 +1,26 @@
...
         {% with form_submit=_("Submit") %}
             {% include "form/form.html" %}
         {% endwith %}
-    </div>
-    <div id="password_change">
-        <p><a href="{% url "password_change" %}">{% trans "Change your password" %}</a></p>
-        {% if not user.email_confirmed and user.email %}
-            <p><a href="{% url "email_confirmation_send_token" %}">{% trans "Please confirm your e-mail address" %}</a></p>
-        {% endif %}
+        <div id="password_change" class="automatic-width">
```
mitar: Why `automatic-width` here?

piplmesh/account/templates/user/login.html
```diff
@@ -1,12 +1,17 @@
...
 
+{% block js %}
+    {{ block.super }}
+    <script src="{% static "piplmesh/js/forms.js" %}" type="text/javascript"></script>
+{% endblock %}
+
 {% block content %}
     <div class="login">
-        <h1>{% trans "Login" %}</h1>
+        <h1 class="automatic-width">{% trans "Login" %}</h1>
```
mitar: Why `automatic-width` here?

piplmesh/account/templates/user/menu.html
```diff
@@ -0,0 +1,5 @@
+{% load future i18n account %}
+<ul id="menu_list">
+    <li class="{% active request "profile" %}"><a href="{% url "profile" username=user.username %}">{% trans "Profile" %}</a></li>
```
mitar: Does not tags have automatic access to template context? You don't need to pass request here?

piplmesh/account/templates/user/registration.html
```diff
@@ -2,16 +2,17 @@
...
     <script src="{% static "piplmesh/js/registration.js" %}" type="text/javascript"></script>
+    <script src="{% static "piplmesh/js/forms.js" %}" type="text/javascript"></script>
 {% endblock %}
 
-{% block title %}{% trans "Registration" %}{% endblock %}
-
 {% block content %}
     <div class="registration">
-        <h1>{% trans "Registration" %}</h1>
+        <h1 class="automatic-width">{% trans "Registration" %}</h1>
```
mitar: Why `automatic-width` here?

piplmesh/account/templatetags/account.py
```diff
@@ -10,3 +10,12 @@ def user_image(context, user=None):
     return {
         'user_image_url': user.get_image_url(),
     }
+
+@register.simple_tag
```
mitar: Do `@register.simple_tag(takes_context=True)`.

piplmesh/frontend/templates/contact.html
```diff
@@ -1,19 +1,24 @@
...
 
+{% block js %}
+    {{ block.super }}
+    <script src="{% static "piplmesh/js/forms.js" %}" type="text/javascript"></script>
+{% endblock %}
+
 {% block content %}
     <div class="contact_form">
-        <h1>{% trans "Contact" %}</h1>
+        <h1 class="automatic-width">{% trans "Contact" %}</h1>
```
mitar: I really don't understand why are you resizing titles? I thought that we said that we will use class for form fields?

piplmesh/frontend/static/piplmesh/js/forms.js
```diff
@@ -0,0 +1,9 @@
+$(document).ready(function () {
+    var max = 0;
+    $('label').each(function () {
+        if ($(this).width() > max)
+            max = $(this).width();
+    });
+    $('label').width(max);
+    $('.automatic-width').css({ 'margin-left': max });
+});
```
mitar, line 9: I think you are not using this class as intended. ;-)

The idea is that you don't resize ALL labels, but only those which are inside of `automatic-width`-marked fields. So that you can move all this code to `common.js`, not that you have to manually search places where to include it.

piplmesh/account/templatetags/account.py
```diff
@@ -10,3 +10,12 @@ def user_image(context, user=None):
...
+def active(request, url_name):
+    from django.core.urlresolvers import resolve
```
mitar: Ehm, you are doing it in opposite direction. Wouldn't it be getter if you would simply find URL from `url_name` and then check if this is the prefix of current URL? If it is, you return something, if not, not.

piplmesh/account/templatetags/account.py
```diff
@@ -10,3 +10,12 @@ def user_image(context, user=None):
...
         'user_image_url': user.get_image_url(),
     }
+
+@register.simple_tag
+def active(request, url_name):
+    from django.core.urlresolvers import resolve 
+    startswith = request.path.startswith('/'+url_name)
+    resolved_name = resolve(request.path).url_name
+    if startswith or url_name == resolved_name:
+        return 'current_item'
```
mitar: See, this is the problem. This is part of the template, not code. The name of CSS class is designer's job to decide how to do. You should just return some value to signal if you should put the class or not.

I would recommend using filter, not tag. You could then do in the template:

```
{% if request.path|is_active:"account" %} class="current_item"{% endif %}
```

borismarn:
I've clean the pull request by moving master branch to master2 and user-account-design to user-account-design2 branch. That way I was able to make a clean master branch from the main repository and a new user-account-design branch from the updated master. I've cherry-picked the commits from user-account-design2 to user-account-design and then I just forced a push to github.

piplmesh/frontend/templates/contact.html
borismarn: I thought it looked better, but I guess it looks good without it too. I have removed it  and added .align_to_label_width just to few elements (buttons, errors), maybe this would be more informative what this class does.

piplmesh/account/templatetags/account.py
```diff
@@ -10,3 +10,9 @@ def user_image(context, user=None):
     return {
         'user_image_url': user.get_image_url(),
     }
+
+@register.filter(name='is_active')
```
mitar: No need for `name=` because it is the same as function name.

piplmesh/account/templates/user/menu.html
```diff
@@ -0,0 +1,7 @@
+{% load future i18n account %}
```
mitar, line 1: Empty line after loading tag.

piplmesh/frontend/templates/user/user.html
```diff
@@ -5,16 +5,12 @@
...
-        {% endif %}
-    </div>
-    <div id="picture" >        
-        {% user_image object %}
-    </div>
-    <div id="profile">
+    {% include "user/menu.html" %}
+    <div class="profile">
+        <h1>{% trans "Profile" %}</h1>
+        <div id="picture" >
```
mitar: Unnecessary space before `>`.

piplmesh/frontend/static/piplmesh/js/forms.js
```diff
@@ -0,0 +1,9 @@
+$(document).ready(function () {
+    var max = 0;
+    $('.field label.main').each(function () {
```
mitar: Write all arguments in callback function which function gets so that possible later developers will know what is available.

piplmesh/frontend/static/piplmesh/js/forms.js
```diff
@@ -0,0 +1,9 @@
...
+        if ($(this).width() > max)
+            max = $(this).width();
+    });
+    $('.field label.main').width(max);
+    $('.align_to_label_width').css({ 'margin-left': max });
```
mitar: Use two argument version instead of dict if you have only one value to set. (`.css('margin-left', max);`)

piplmesh/account/templatetags/account.py
```diff
@@ -10,3 +10,9 @@ def user_image(context, user=None):
...
+def is_active(current_path, url_path):
+    if current_path.startswith(url_path):
```
mitar: Ehm, `return current_path.startswith(url_path)` is not enough?

mitar:
Otherwise it is all good now. Please update from main repository for me to be able to pull this pull request in.

borismarn:
Updated from main repository and fixed as suggested.

piplmesh/frontend/static/piplmesh/js/forms.js
```diff
@@ -0,0 +1,9 @@
+$(document).ready(function () {
+    var max = 0;
+    $('.field label.main').each(function (index, element) {
+        if ($(this).width() > max)
+            max = $(this).width();
+    });
+    $('.field label.main').width(max);
+    $('.align_to_label_width').css('margin-left', max );
```
mitar, line 8: Unnecessary space near the end.

Issue #260, opened 2012-08-01
Display Bicikelj information


mitar:
Display information about Bicikelj locations, free spaces and available bicycles on each location. Both on map and in a panel.

lazarm:
Is it okay to fetch this information from these sites :
http://www.bicikelj.si/service/stationdetails/ljubljana/27
http://www.bicikelj.si/service/carto

The good thing is it's xml file, so information is easily fetched, but the problem could be if URL changes (it already did once from .../stationdetails/27 to .../stationdetails/ljubljana/27) ?

mitar:
Perfect! I didn't know that they publish this in XML. Is this official API? Is there any official documentation for it? Do you know what is "ticket" and "open"? I don't think URL changed (as far as I read around the Internet about the API). Anyway, don't be concerned about that. You cannot do much about it. But you should make sure that your code fails if URL changes so that we will notice. (Aha, ticket means "if it supports the option to rent by credit card".)

Is this information real-time? So the question which opens is: should we prefetch this information in background or should clients access it directly through JavaScript? Can clients access it?  I think not because XML is on other server and they don't support JSONP (or do they)?

But it is so visible that it was developed for Paris. "arrondissements". :-)))

Useful reading (searching Google for "service/stationdetails" helps):
 *https://ocfnash.wordpress.com/2009/10/31/dublin-bikes/

So, make a background task which fetches data every minute (yes, polling, great for them!). And then display this data in the panel and on the map for bike station around nodes.

Also add pushing of new free/empty station data to users. So the main information we want to display is not where are locations but what is state there. And we should update this in as real-time as we can. (You have timestamp in station details so you can know if something changed and you have to push.)

One more thing. Store all previous states of the stations in the database. Don't delete them. It could be interesting statistics out of it through some time. :-) (You can of course convert timestamp to data time data type.)

Check how horoscope is done. It is very similar what you are doing here to that. It also has XML parsing and background job. And it also stores old data.

lazarm:
Yes, it's their official API, I've found it on some apple forum site :) "open" has a binary value, 1 if the station is open and 0 if it's closed (apparently, this can happen).
My question is: how many bike station markers should be displayed on the map and should i calculate k nearest stations from longitude and latitude between the node and each station available (and also define max distance, so it wouldn't display a station in Siska if the user is using a node somewhere in Kranj).
[...]

Issue #248, opened 2012-06-13
Interactive map


mitar:
Currently, we have a panel with a map, which is inside panel and quite small. Its current use is only to display node location and location of other things from other panels. So if you read panels and you would like to know where some information is, you can check on the map.

But users also often search in opposite direction. Watching the map and explore it. And current small map is not useful for that. So we could add a functionality to current map panel, that after press on some button (or even anywhere on the map), an overlay popup is displayed into which map is resized over the rest of the content, almost filling the entire user's window. This map could be interactive, with layers of content from different panels users could disable/enable.

The tricky part is here interaction with panels, so there should be some way to load additional content (layers) panels are providing when user moves around the map. Small current map is fixed, so we just initially load content inside the view of the map. But as map would be dynamic, user could move around, so new information should be loaded and displayed.

Also other open nodes of the network should be displayed in this bigger map.

mitar:
Some ideas: http://www.tidepools.co/

mitar:
Initial work done with #267. Now we can get big-sized map overlayed over everything.

PR #275, opened 2012-08-31
Issue #260: Bicikelj panel

lazarm:
It's not entirely finished, I still need to do pushing and displaying changes, but I hope I'll get some help here :)


piplmesh/panels/bicikelj/management/commands/updatebicikelj.py
```diff
@@ -0,0 +1,21 @@
```
mitar, line 0: This is not really needed. As we will run task very often (on minute basis) there is no need for command. For horoscope, which runs only once per day, it is necessary so you can test.

piplmesh/settings.py
```diff
@@ -189,6 +190,8 @@
     'tastypie_mongoengine',
     'sekizai',
     'missing',
+    'django_mongodb_engine',
```
mitar: What is this?

piplmesh/settings.py
```diff
@@ -189,6 +190,8 @@
...
+    'djangotoolbox',
```
mitar: Why is this needed?

piplmesh/settings.py
```diff
@@ -178,6 +178,7 @@
     'piplmesh.utils',
     'piplmesh.panels',
     'piplmesh.panels.horoscope', # To load manage.py command
+    'piplmesh.panels.bicikelj',
```
mitar: You should not need this.

piplmesh/settings.py
```diff
@@ -219,6 +222,7 @@
 
 CHECK_ONLINE_USERS_INTERVAL = 10 # seconds
 CHECK_FOR_NEW_HOROSCOPE = 6 # am every day
+CHECK_BICIKELJ_STANDS_INFO = 60
```
mitar: Rename to `POLL_BICIKELJ_INTERVAL`.

mitar: Add units into a comment at the end.

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,14 @@
+import mongoengine
+
+NUMBER_OF_STATIONS = 32
```
mitar: O no no. You do not need to hardcode that. You have everything in `carto` URL.

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,14 @@
...
+
+class BicikeljStation(mongoengine.Document):
+    name = mongoengine.StringField(required=True)
+    address = mongoengine.StringField(required=True)
+    number = mongoengine.IntField(required=True)
+    position = mongoengine.GeoPointField(required=True)
+    available = mongoengine.IntField()
+    free = mongoengine.IntField()
+    open = mongoengine.IntField(choices=(0,1))
```
mitar: Ehm, boolean is not OK? :-)

mitar:
BTW, #248 is not Bicikelj panel.

piplmesh/panels/bicikelj/models.py
mitar: Move `open` after `position`.

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,14 @@
+import mongoengine
+
+NUMBER_OF_STATIONS = 32
+
+class BicikeljStation(mongoengine.Document):
+    name = mongoengine.StringField(required=True)
+    address = mongoengine.StringField(required=True)
+    number = mongoengine.IntField(required=True)
+    position = mongoengine.GeoPointField(required=True)
```
mitar: Rename to `location`.

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,14 @@
+import mongoengine
+
+NUMBER_OF_STATIONS = 32
+
+class BicikeljStation(mongoengine.Document):
+    name = mongoengine.StringField(required=True)
+    address = mongoengine.StringField(required=True)
+    number = mongoengine.IntField(required=True)
```
mitar: Rename to `id`.

mitar: Move to very beginning.

piplmesh/panels/bicikelj/models.py
mitar: Store also total. It does not matter that data is not normalized. It is easier to search like that.

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,14 @@
...
+    position = mongoengine.GeoPointField(required=True)
+    available = mongoengine.IntField()
+    free = mongoengine.IntField()
+    open = mongoengine.IntField(choices=(0,1))
+    ticket = mongoengine.IntField(choices=(0,1))
```
mitar: Ignore this one.

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,14 @@
...
+    timestamp = mongoengine.IntField()
```
mitar: Use datetime field.

piplmesh/panels/bicikelj/panel.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import absolute_import
+from django.utils.translation import ugettext_lazy as _
```
mitar, line 3: Empty line between each group of imports.

piplmesh/panels/bicikelj/panel.py
```diff
@@ -0,0 +1,35 @@
...
+class BicikeljPanel(panels.BasePanel):
+    def get_context(self, context):
+
+        context.update({
+            'header': _("Bicikelj stations"),
+        })
+
+        try:
+            node = self.request.node
+        except :
```
mitar: This is not possible. Furthermore, never catch ALL exceptions, only those you expect.

piplmesh/panels/bicikelj/panel.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import absolute_import
+from django.utils.translation import ugettext_lazy as _
+from piplmesh import panels
+from . import stations
+
+class BicikeljPanel(panels.BasePanel):
+    def get_context(self, context):
+
```
mitar: Unnecessary empty line.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,45 @@
...
+BICIKELJ_STANDS_URL = 'http://www.bicikelj.si/service/carto'
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/'
+ZOOM_LEVEL = 15
+MAP_WIDTH = 450
+C = 0.703119412486786 # constant at 60 degrees
+
+def max_distance(zoom, width):
+    return C*width/math.pow(2,zoom)
+
+def get_nearest_stations(lat, lng):
```
mitar: `get_nearest_stations` is not really a good name. Nearest is really only the one (it would be really strange coincidence to have more than one nearest). You are not search really for the nearest, but all in some close proximity, range.

piplmesh/panels/bicikelj/panel.py
```diff
@@ -0,0 +1,35 @@
...
+        stands = stations.get_nearest_stations(node.latitude, node.longitude)
+
+        if not stands:
+            context.update({
+                'error_data': True,
+            })
+            return context
+
+        context.update({
+            'stations': stands,
```
mitar: Why once `stations` and then `stands`?

mitar: You have problems with conflict with imported module? :-)

mitar: So this function could really be very simple:

```
context.update({
    'header': _("Bicikelj stations"),
    'stations': stations.get_stations(self.request.node.latitude, self.request.node.longitude),
})
```

mitar: And then you can simply in the template check if there is any data in `stations`. Which is not really necessary an error, simply maybe there are no stations around.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,35 @@
...
+
+{% block content %}
+    {% addtoblock "css" %}<link href="{% static "piplmesh/panel/bicikelj/panel.css" %}" rel="stylesheet" type="text/css" media="screen" />{% endaddtoblock %}
+    <div id="bicikelj">
+        {% if error_node %}
+        <span class="bicikelj_error">{% trans "Unknown location." %}</span>
+        {% elif error_data %}
+        <span class="bicikelj_error">{% trans "No Bicikelj stands nearby" %}</span>
+        {% else %}
+           {% for st in stations %}
```
mitar: Write variable names out, fully.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,35 @@
...
+                <div id="'st+str(st.number)'">
```
mitar: What is this???

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,35 @@
...
+                <dl>
+                    <dt>
+                        <span class="station_name">{{ st.name }}</span>
+                    </dt>
+                    <dd>
+                        <ul>
+                            <li><span class="station_availability">{% trans "Available bikes: " %}<span></span></span></li>
```
mitar: `<span></span></span>`???

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,35 @@
...
+                            <li><span class="station_free">{% trans "Free places: " %}{{ st.free }}</span></li>
+                            <li><span class="station_ticket">{% trans "Credit card: " %}{% if st.ticket == 1 %}{% trans "yes" %}{% else %}{% trans "no" %}{% endif %}</span></li>
```
mitar: Is there any with ticket==1 in Ljubljana?

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,35 @@
...
+           {% for st in stations %}
+                <div id="'st+str(st.number)'">
+                <dl>
+                    <dt>
+                        <span class="station_name">{{ st.name }}</span>
+                    </dt>
+                    <dd>
+                        <ul>
+                            <li><span class="station_availability">{% trans "Available bikes: " %}<span></span></span></li>
+                            <li><span class="station_free">{% trans "Free places: " %}{{ st.free }}</span></li>
```
mitar: Hm, wouldn't it be better to have string "23 free places"? Can you format string like that?

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,31 @@
+from celery import task
+from pushserver.utils import updates
+
+from piplmesh.frontend import views
```
mitar, line 5: Group imports properly.

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+from . import models, stations
+
+@task.task
+def update_station_info():
+    station_info = stations.fetch_data()
+    for n in range(1,models.NUMBER_OF_STATIONS+1):
```
mitar: Why? Why you cannot simply iterate over anything returned by `fetch_data`?

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+        if not models.BicikeljStation.objects(number=n, timestamp=station_info[n-1]['timestamp']):
```
mitar: There is a racing condition here. You should not check and then save. Use upsert.

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+            models.BicikeljStation(
+                                           name=station_info[n-1]['name'],
```
mitar: Indentation.

And there are easier ways to pass this data if keys in the dict are the same as keys in the document model. Which they should be.

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+                                           timestamp=station_info[n-1]['timestamp'],
+                                           available = station_info[n-1]['available'],
+                                           free = station_info[n-1]['free'],
+                                           open = station_info[n-1]['open'],
+                                           ticket = station_info[n-1]['ticket'],
+            ).save()
+            updates.send_update(
+                views.HOME_CHANNEL_ID,
+                # TODO
+                )
```
mitar: Indentation.

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+
+
```
mitar: Unnecessary empty lines.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,45 @@
...
+            break
+    stands = [models.BicikeljStation.objects(number=num).order_by('-timestamp').first() for num in ids]
+    return stands
+
+def fetch_data():
+    stands_data = urllib2.urlopen(BICIKELJ_STANDS_URL).read()
+    stands_xml_data = minidom.parseString(stands_data)
+    stands_nodes = stands_xml_data.getElementsByTagName('marker')
+    info_data = [urllib2.urlopen(BICIKELJ_INFO_URL+str(number)).read() for number in range(1,models.NUMBER_OF_STATIONS+1)]
+    info_xml_data = [minidom.parseString(info_data[i]) for i in range(models.NUMBER_OF_STATIONS)]
```
mitar: Do it without using `NUMBER_OF_STATIONS`.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,45 @@
...
+        except:
+            break
+    stands = [models.BicikeljStation.objects(number=num).order_by('-timestamp').first() for num in ids]
+    return stands
+
+def fetch_data():
+    stands_data = urllib2.urlopen(BICIKELJ_STANDS_URL).read()
+    stands_xml_data = minidom.parseString(stands_data)
+    stands_nodes = stands_xml_data.getElementsByTagName('marker')
+    info_data = [urllib2.urlopen(BICIKELJ_INFO_URL+str(number)).read() for number in range(1,models.NUMBER_OF_STATIONS+1)]
```
mitar: Use more spaces, for example, around `,`.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,45 @@
...
+    while 1:
+        try:
+            ids.append(models.BicikeljStation.objects(position__near= [lat, lng], position__within_box=[(lat-abs(maxd),lng-abs(maxd)),(lat+abs(maxd),lng+abs(maxd))], number__nin=ids).first().number)
+        except:
+            break
+    stands = [models.BicikeljStation.objects(number=num).order_by('-timestamp').first() for num in ids]
+    return stands
+
+def fetch_data():
+    stands_data = urllib2.urlopen(BICIKELJ_STANDS_URL).read()
```
mitar: Decide if you name things `stations` or `stands`.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,45 @@
+import math, urllib2
+from xml.dom import minidom
+
+from . import models
+
+BICIKELJ_STANDS_URL = 'http://www.bicikelj.si/service/carto'
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/'
+ZOOM_LEVEL = 15
+MAP_WIDTH = 450
+C = 0.703119412486786 # constant at 60 degrees
```
mitar: I think I said that you should not do it like that.

piplmesh/panels/bicikelj/panel.py
lazarm: Does node also have a value if a user comes to our page when not connected to one of the wireless nodes?

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
lazarm: So is this a right way, displaying info about each node in a for loop?

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
lazarm: Nothing yet :-) I don't know anything about jquery, but I was reading about accessing html elements and their values. And I don't know any other way how to  access a particular element in a for loop (for example free spots of the third station) than adding id value for each div element in a loop (but then I also read having a number as id is possible only for HTML5, but is not recommended). So I'm still without any ideas.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
lazarm: No, I'll remove it from a model.

piplmesh/panels/bicikelj/tasks.py
lazarm: I know. My idea was to find a particular station with the newest timestamp, and if it fails, it would save it and push changes to the server. So pushing happens only if timestamp is changed. If I use upsert, it will update if it finds the object (which is pointless since timestamp is old and so is the data about free spots etc) or save it. So I would have to know  which action happened, update or save. to know if pushing needs to be done. Is there a way to retrieve this info?

piplmesh/panels/bicikelj/stations.py
lazarm: You don't mind any of them?

piplmesh/panels/bicikelj/stations.py
lazarm: I know, I just wanted to check if formula si correct and only stations within a map are chosen. But still, I don't quite understand how to do it. Should I calculate NW and SE points based on node and zoom level inside a panel.js in a map panel, and then use:

> var bounds = new google.maps.LatLngBounds();
> bounds.extend(NW);
> bounds.extend(SE);

And how should I access these two points to query the stations with "within_box" ?

mitar:

piplmesh/panels/bicikelj/panel.py
mitar: Yes. You can assume that `request.node` is always populated with something. How it is done you do not need to worry. :-)

(Currently, we allow only access inside the network.)

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
mitar: Sure. Just write `station` instead of `st`. Use indentation correctly. Order them in some meaningful way (maybe by distance from the node).

piplmesh/panels/bicikelj/stations.py
mitar: I don't know. What they are using in their API? Use the same.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
mitar: You could do something like that, yes. But this syntax is not correct for Django templates.

Or, you could use `.data()` function from jQuery to attach arbitrary data to DOM elements. Like id. And then search for that. (We are using that in #199.

piplmesh/panels/bicikelj/tasks.py
mitar: OK. You are right. In this context it does not matter if we override the previous value, because we will override with same content. So yes, we can check if value exists and if not, we store it.

Please rename `number` to `id`. And don't have number of stations hard-coded.

piplmesh/panels/bicikelj/stations.py
mitar: You store width and height (bounds) hard-coded as global variables in Python. You use that for "within_box". And pass them to JavaScript. And use them to check if they are inside of bounds when map is displayed.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
lazarm: They are already ordered in get_nearest_stations (yes, I'll rename that function).  __near=[lat,lng]  orderes all stations by distance from the [lat,lng] and within_box removes those who fall out of map.

piplmesh/panels/bicikelj/management/commands/updatebicikelj.py
mitar, line 0: TODO.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
mitar: Great!

mitar:
So, waiting for you ... :-)

piplmesh/panels/bicikelj/stations.py
lazarm: I don't understand your last sentence. Where and what should I check?

piplmesh/panels/bicikelj/models.py
lazarm: How do I query with id? models.BicikeljStation.objects(id=station['id'],..)  keeps returning ValidationError?

mitar: Do you have `id` field? Currently you have it named `number`. You should rename it to `id`. The idea is that `number` name is badly named.

piplmesh/panels/bicikelj/stations.py
mitar: That values you have computed are still in the bounds of the displayed map.

piplmesh/panels/bicikelj/models.py
lazarm: I renamed it in my code and then checked if everything works fine, but it returns ValidationError, when doing a query mentioned above in a task. If I don't have it named "id", it works fine. Is something wrong with defining a field with "id = mongoengine.IntField(required=True)" ?

lazarm:
Great, what a mess... just great! New pull request?

mitar:
Yes, better a new one. Or clean this one. You can remove the commit you don't like from your repository and it will also remove here.

piplmesh/panels/bicikelj/models.py
mitar: Yes, `id` field is special. Maybe name it then `stand_id` (or `station_id`), I am not sure what you decided to use.

PR #286, opened 2012-09-08
Issue #260: Bicikelj panel (2nd attempt)

lazarm:
#260

### DIFF
--- a/piplmesh/panels/bicikelj/models.py
+++ b/piplmesh/panels/bicikelj/models.py
@@ -1,0 +1,12 @@
+import mongoengine
+
+class BicikeljStation(mongoengine.Document):
+    id = mongoengine.IntField(required=True)
+    name = mongoengine.StringField(required=True)
+    address = mongoengine.StringField(required=True)
+    location = mongoengine.GeoPointField(required=True)
+    open = mongoengine.IntField()
+    available = mongoengine.IntField()
+    free = mongoengine.IntField()
+    total = mongoengine.IntField()
+    timestamp = mongoengine.DateTimeField()
--- a/piplmesh/panels/bicikelj/models.py
+++ b/piplmesh/panels/bicikelj/models.py
@@ -2,6 +2,8 @@ import mongoengine
 
 class BicikeljStation(mongoengine.Document):
     station_id = mongoengine.IntField(required=True, unique_with='timestamp')
+    timestamp = mongoengine.DateTimeField(required=True, unique_with='station_id')
+    fetch_time = mongoengine.DateTimeField(required=True)
     name = mongoengine.StringField(required=True)
     address = mongoengine.StringField(required=True)
     location = mongoengine.GeoPointField(required=True)
@@ -9,7 +11,6 @@ class BicikeljStation(mongoengine.Document):
     available = mongoengine.IntField(required=True)
     free = mongoengine.IntField(required=True)
     total = mongoengine.IntField(required=True)
-    timestamp = mongoengine.DateTimeField(required=True, unique_with='station_id')
 
     meta = {
         'indexes': ['station_id', 'timestamp'],
--- a/piplmesh/panels/bicikelj/models.py
+++ b/piplmesh/panels/bicikelj/models.py
@@ -1,12 +1,16 @@
 import mongoengine
 
 class BicikeljStation(mongoengine.Document):
-    id = mongoengine.IntField(required=True)
+    station_id = mongoengine.IntField(required=True, unique_with='timestamp')
     name = mongoengine.StringField(required=True)
     address = mongoengine.StringField(required=True)
     location = mongoengine.GeoPointField(required=True)
-    open = mongoengine.IntField()
-    available = mongoengine.IntField()
-    free = mongoengine.IntField()
-    total = mongoengine.IntField()
-    timestamp = mongoengine.DateTimeField()
+    open = mongoengine.BooleanField(required=True)
+    available = mongoengine.IntField(required=True)
+    free = mongoengine.IntField(required=True)
+    total = mongoengine.IntField(required=True)
+    timestamp = mongoengine.DateTimeField(required=True, unique_with='station_id')
+
+    meta = {
+        'indexes': ['station_id', 'timestamp'],
+    }
--- a/piplmesh/panels/bicikelj/panel.py
+++ b/piplmesh/panels/bicikelj/panel.py
@@ -10,7 +10,7 @@ class BicikeljPanel(panels.BasePanel):
     def get_context(self, context):
         context.update({
             'header': _("Bicikelj stations"),
-            'stations': stations.get_stations_nearby(self.request.node.latitude, self.request.node.longitude)
+            'stations': stations.get_stations_nearby(self.request.node.latitude, self.request.node.longitude),
         })
         return context
 
--- a/piplmesh/panels/bicikelj/panel.py
+++ b/piplmesh/panels/bicikelj/panel.py
@@ -1,0 +1,17 @@
+from __future__ import absolute_import
+
+from django.utils.translation import ugettext_lazy as _
+
+from piplmesh import panels
+
+from . import stations
+
+class BicikeljPanel(panels.BasePanel):
+    def get_context(self, context):
+        context.update({
+            'header': _("Bicikelj stations"),
+            'stations': stations.get_stations_nearby(self.request.node.latitude, self.request.node.longitude)
+        })
+        return context
+
+panels.panels_pool.register(BicikeljPanel)
--- a/piplmesh/panels/bicikelj/static/piplmesh/panel/bicikelj/panel.js
+++ b/piplmesh/panels/bicikelj/static/piplmesh/panel/bicikelj/panel.js
@@ -1,0 +1,1 @@
+/* TODO: attach DOM elements to stations and deal with pushed updates
--- a/piplmesh/panels/bicikelj/static/piplmesh/panel/bicikelj/panel.js
+++ b/piplmesh/panels/bicikelj/static/piplmesh/panel/bicikelj/panel.js
@@ -1,1 +1,1 @@
-/* TODO: attach DOM elements to stations and deal with pushed updates
+/* TODO: update displayed data from pushed updates */
--- a/piplmesh/panels/bicikelj/stations.py
+++ b/piplmesh/panels/bicikelj/stations.py
@@ -1,46 +1,62 @@
-import datetime, urllib2
-from xml.dom import minidom
+import urllib2
+from datetime import date, datetime, timedelta
+from lxml import etree, objectify
+from pytz import UTC
 
+from piplmesh import settings
 from . import models
 
 BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
 BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/%d'
+
+""" A comment explaining how to calculate map bounds, so we can search for stations visible on map only
+
+BOUNDS_LATITUDE and BOUNDS_LONGITUDE are the distance between a node and top location of the map
+and the distance between a node and leftmost location of the map, respectively
+BOUNDS_LATITUDE = map_heigth*0.703119412486786/2^map_zoom
+BOUNDS_LONGITUDE = map_width*0.703119412486786/2^map_zoom
+
 """
-BOUNDS_LATITUDE = map_heigth*0.703119412486786/2^(map_zoom+1)
-BOUNDS_LONGITUDE = map_width*0.703119412486786/2^(map_zoom+1)
-"""
-BICIKELJ_BOUNDS_LATITUDE = 0.0027895
-BICIKELJ_BOUNDS_LONGITUDE = 0.0027895
+
+BICIKELJ_BOUNDS_LATITUDE = 0.00557895
+BICIKELJ_BOUNDS_LONGITUDE = 0.00557895
 
 def get_stations_nearby(latitude, longitude):
-    ids = []
-    while 1:
-        try:
-            ids.append(models.BicikeljStation.objects(location__near= (latitude, longitude), location__within_box=((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)), id__nin=ids).first().id)
-        except:
-            break
-    stations = [models.BicikeljStation.objects(id=n).order_by('-timestamp').first() for n in ids]
-    return stations
+    stations_nearby_all = models.BicikeljStation.objects(
+                location__near=(latitude, longitude),
+                location__within_box=((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)),
+                fetch_time__gt=datetime.now() - timedelta(seconds=12*settings.POLL_BICIKELJ_INTERVAL),
+    )
+    n = stations_nearby_all[0].station_id
+    station_idx = []
+    for j,station in enumerate(stations_nearby_all):
+        if station.station_id != n:
+            station_idx.append(j-1)
+            n = station.station_id
+    station_idx.append(len(stations_nearby_all)-1)
+    stations_nearby = [stations_nearby_all[i] for i in station_idx]
+    for station in stations_nearby:
+        station.old_data = station.fetch_time < UTC.localize(datetime.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL))
+        station.very_old_data = station.fetch_time < UTC.localize(datetime.now() - timedelta(seconds=10*settings.POLL_BICIKELJ_INTERVAL))
+    return stations_nearby
 
 def fetch_data():
-    stations_data = urllib2.urlopen(BICIKELJ_STATIONS_URL).read()
-    stations_xml_data = minidom.parseString(stations_data)
-    stations_nodes = stations_xml_data.getElementsByTagName('marker')
-    info_data = [urllib2.urlopen(BICIKELJ_INFO_URL % number).read() for number in range(1,len(stations_nodes)+1)]
-    info_xml_data = [minidom.parseString(info_data[i]) for i in range(len(stations_nodes))]
-    data = [
-        [{
-            'station_id': int(stations_nodes[number].attributes['number'].value),
-            'timestamp': datetime.datetime.fromtimestamp(int(info_xml_data[number].getElementsByTagName('updated')[0].firstChild.nodeValue)),
-        },
-        {
-            'name': stations_nodes[number].attributes['name'].value,
-            'address': stations_nodes[number].attributes['address'].value,
-            'location': [float(stations_nodes[number].attributes['lat'].value), float(stations_nodes[number-1].attributes['lng'].value)],
-            'available': int(info_xml_data[number].getElementsByTagName('available')[0].firstChild.nodeValue),
-            'free': int(info_xml_data[number].getElementsByTagName('free')[0].firstChild.nodeValue),
-            'total': int(info_xml_data[number].getElementsByTagName('total')[0].firstChild.nodeValue),
-            'open': int(info_xml_data[number].getElementsByTagName('open')[0].firstChild.nodeValue),
-        }]
-        for number in range(len(stations_nodes))]
+    stations_tree = objectify.fromstring(urllib2.urlopen(BICIKELJ_STATIONS_URL).read())
+    data = []
+    for node in stations_tree.markers.marker:
+        info_data = objectify.fromstring(urllib2.urlopen(BICIKELJ_INFO_URL % int(node.attrib['number'])).read())
+        data.append(
+            {
+                'station_id': int(node.attrib['number']),
+                'timestamp': datetime.fromtimestamp(int(info_data.updated)),
+                'name': node.attrib['name'],
+                'address': node.attrib['address'],
+                'location': (float(node.attrib['lat']), float(node.attrib['lng'])),
+                'available': int(info_data.available),
+                'free': int(info_data.free),
+                'total': int(info_data.total),
+                'open': bool(info_data.open),
+                'fetch_time': datetime.now(),
+            }
+        )
     return data
--- a/piplmesh/panels/bicikelj/stations.py
+++ b/piplmesh/panels/bicikelj/stations.py
@@ -6,19 +6,12 @@ from . import models
 BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
 BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/'
 BICIKELJ_BOUNDS = 0.00965
-ZOOM_LEVEL = 15
-MAP_WIDTH = 450
-C = 0.703119412486786 # constant at 60 degrees
-
-def max_distance(zoom, width):
-    return C*width/math.pow(2,zoom)
 
 def get_stations_nearby(lat, lng):
-    maxd = max_distance(ZOOM_LEVEL,MAP_WIDTH)
     ids = []
     while 1:
         try:
-            ids.append(models.BicikeljStation.objects(location__near= [lat, lng], location__within_box=[(lat-abs(maxd),lng-abs(maxd)),(lat+abs(maxd),lng+abs(maxd))], id__nin=ids).first().id)
+            ids.append(models.BicikeljStation.objects(location__near= [lat, lng], location__within_box=[(lat-BICIKELJ_BOUNDS,lng-BICIKELJ_BOUNDS),(lat+BICIKELJ_BOUNDS,lng+BICIKELJ_BOUNDS)], id__nin=ids).first().id)
         except:
             break
     stations = [models.BicikeljStation.objects(id=n).order_by('-timestamp').first() for n in ids]
@@ -28,7 +21,6 @@ def fetch_data():
     stations_data = urllib2.urlopen(BICIKELJ_STATIONS_URL).read()
     stations_xml_data = minidom.parseString(stations_data)
     stations_nodes = stations_xml_data.getElementsByTagName('marker')
-    print stations_nodes
     info_data = [urllib2.urlopen(BICIKELJ_INFO_URL+str(number)).read() for number in range(1,len(stations_nodes)+1)]
     info_xml_data = [minidom.parseString(info_data[i]) for i in range(len(stations_nodes))]
     data = [
--- a/piplmesh/panels/bicikelj/stations.py
+++ b/piplmesh/panels/bicikelj/stations.py
@@ -1,17 +1,22 @@
-import math, urllib2, datetime
+import datetime, urllib2
 from xml.dom import minidom
 
 from . import models
 
 BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
-BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/'
-BICIKELJ_BOUNDS = 0.00965
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/%d'
+"""
+BOUNDS_LATITUDE = map_heigth*0.703119412486786/2^(map_zoom+1)
+BOUNDS_LONGITUDE = map_width*0.703119412486786/2^(map_zoom+1)
+"""
+BICIKELJ_BOUNDS_LATITUDE = 0.0027895
+BICIKELJ_BOUNDS_LONGITUDE = 0.0027895
 
-def get_stations_nearby(lat, lng):
+def get_stations_nearby(latitude, longitude):
     ids = []
     while 1:
         try:
-            ids.append(models.BicikeljStation.objects(location__near= [lat, lng], location__within_box=[(lat-BICIKELJ_BOUNDS,lng-BICIKELJ_BOUNDS),(lat+BICIKELJ_BOUNDS,lng+BICIKELJ_BOUNDS)], id__nin=ids).first().id)
+            ids.append(models.BicikeljStation.objects(location__near= (latitude, longitude), location__within_box=((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)), id__nin=ids).first().id)
         except:
             break
     stations = [models.BicikeljStation.objects(id=n).order_by('-timestamp').first() for n in ids]
@@ -21,21 +26,21 @@ def fetch_data():
     stations_data = urllib2.urlopen(BICIKELJ_STATIONS_URL).read()
     stations_xml_data = minidom.parseString(stations_data)
     stations_nodes = stations_xml_data.getElementsByTagName('marker')
-    info_data = [urllib2.urlopen(BICIKELJ_INFO_URL+str(number)).read() for number in range(1,len(stations_nodes)+1)]
+    info_data = [urllib2.urlopen(BICIKELJ_INFO_URL % number).read() for number in range(1,len(stations_nodes)+1)]
     info_xml_data = [minidom.parseString(info_data[i]) for i in range(len(stations_nodes))]
     data = [
+        [{
+            'station_id': int(stations_nodes[number].attributes['number'].value),
+            'timestamp': datetime.datetime.fromtimestamp(int(info_xml_data[number].getElementsByTagName('updated')[0].firstChild.nodeValue)),
+        },
         {
-        'name': stations_nodes[number].attributes['name'].value,
-        'address': stations_nodes[number].attributes['address'].value,
-        'id': int(stations_nodes[number].attributes['number'].value),
-        'location': [float(stations_nodes[number].attributes['lat'].value), float(stations_nodes[number-1].attributes['lng'].value)],
-        'available': int(info_xml_data[number].getElementsByTagName('available')[0].firstChild.nodeValue),
-        'free': int(info_xml_data[number].getElementsByTagName('free')[0].firstChild.nodeValue),
-        'total': int(info_xml_data[number].getElementsByTagName('total')[0].firstChild.nodeValue),
-        'open': int(info_xml_data[number].getElementsByTagName('open')[0].firstChild.nodeValue),
-        'timestamp': datetime.datetime.fromtimestamp(int(info_xml_data[number].getElementsByTagName('updated')[0].firstChild.nodeValue)),
-        }
+            'name': stations_nodes[number].attributes['name'].value,
+            'address': stations_nodes[number].attributes['address'].value,
+            'location': [float(stations_nodes[number].attributes['lat'].value), float(stations_nodes[number-1].attributes['lng'].value)],
+            'available': int(info_xml_data[number].getElementsByTagName('available')[0].firstChild.nodeValue),
+            'free': int(info_xml_data[number].getElementsByTagName('free')[0].firstChild.nodeValue),
+            'total': int(info_xml_data[number].getElementsByTagName('total')[0].firstChild.nodeValue),
+            'open': int(info_xml_data[number].getElementsByTagName('open')[0].firstChild.nodeValue),
+        }]
         for number in range(len(stations_nodes))]
-    for d in data:
-        print d['timestamp']
     return data
--- a/piplmesh/panels/bicikelj/stations.py
+++ b/piplmesh/panels/bicikelj/stations.py
@@ -25,7 +25,7 @@ def get_stations_nearby(latitude, longitude):
     stations_nearby_all = models.BicikeljStation.objects(
                 location__near=(latitude, longitude),
                 location__within_box=((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)),
-                fetch_time__gt=datetime.now() - timedelta(seconds=12*settings.POLL_BICIKELJ_INTERVAL),
+                fetch_time__gt=datetime.now() - timedelta(seconds=10*settings.POLL_BICIKELJ_INTERVAL),
     )
     n = stations_nearby_all[0].station_id
     station_idx = []
@@ -37,7 +37,7 @@ def get_stations_nearby(latitude, longitude):
     stations_nearby = [stations_nearby_all[i] for i in station_idx]
     for station in stations_nearby:
         station.old_data = station.fetch_time < UTC.localize(datetime.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL))
-        station.very_old_data = station.fetch_time < UTC.localize(datetime.now() - timedelta(seconds=10*settings.POLL_BICIKELJ_INTERVAL))
+    # TODO: check if some stations nearby have data older than 10 minutes. Then display just their names, without current information
     return stations_nearby
 
 def fetch_data():
--- a/piplmesh/panels/bicikelj/stations.py
+++ b/piplmesh/panels/bicikelj/stations.py
@@ -1,52 +1,51 @@
 import urllib2
-from datetime import date, datetime, timedelta
-from lxml import etree, objectify
-from pytz import UTC
+from datetime import datetime, timedelta
+from django.utils import timezone
+
+from lxml import objectify
 
 from piplmesh import settings
 from . import models
 
 BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
 BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/%d'
+STALE_DATA_TIME = 10
 
-""" A comment explaining how to calculate map bounds, so we can search for stations visible on map only
-
-BOUNDS_LATITUDE and BOUNDS_LONGITUDE are the distance between a node and top location of the map
-and the distance between a node and leftmost location of the map, respectively
-BOUNDS_LATITUDE = map_heigth*0.703119412486786/2^map_zoom
-BOUNDS_LONGITUDE = map_width*0.703119412486786/2^map_zoom
-
-"""
+# A comment explaining how to calculate map bounds, so we can search for stations visible on map only
+#
+# BOUNDS_LATITUDE and BOUNDS_LONGITUDE are the distance between a node and top location of the map
+# and the distance between a node and leftmost location of the map, respectively
+# BOUNDS_LATITUDE = map_height*0.703119412486786/2^map_zoom
+# BOUNDS_LONGITUDE = map_width*0.703119412486786/2^map_zoom
 
 BICIKELJ_BOUNDS_LATITUDE = 0.00557895
 BICIKELJ_BOUNDS_LONGITUDE = 0.00557895
 
 def get_stations_nearby(latitude, longitude):
     stations_nearby_all = models.BicikeljStation.objects(
-                location__near=(latitude, longitude),
-                location__within_box=((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)),
-                fetch_time__gt=datetime.now() - timedelta(seconds=10*settings.POLL_BICIKELJ_INTERVAL),
+        location__near = (latitude, longitude),
+        location__within_box = ((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)),
+        fetch_time__gt = timezone.now() - timedelta(seconds=STALE_DATA_TIME*settings.POLL_BICIKELJ_INTERVAL),
     )
-    n = stations_nearby_all[0].station_id
-    station_idx = []
-    for j,station in enumerate(stations_nearby_all):
-        if station.station_id != n:
-            station_idx.append(j-1)
-            n = station.station_id
-    station_idx.append(len(stations_nearby_all)-1)
-    stations_nearby = [stations_nearby_all[i] for i in station_idx]
-    for station in stations_nearby:
-        station.old_data = station.fetch_time < UTC.localize(datetime.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL))
+    if not stations_nearby_all:
+        return
+    newest_station = stations_nearby_all[0]
+    for station in stations_nearby_all[1:]:
+        if station.station_id != newest_station.station_id:
+            newest_station.old_data = newest_station.fetch_time < timezone.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL)
+            yield newest_station
+            newest_station = station
+        elif station.fetch_time > newest_station.fetch_time:
+            newest_station = station
+    newest_station.old_data = newest_station.fetch_time < timezone.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL)
+    yield newest_station
     # TODO: check if some stations nearby have data older than 10 minutes. Then display just their names, without current information
-    return stations_nearby
 
 def fetch_data():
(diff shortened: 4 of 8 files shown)


piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,12 @@
+import mongoengine
+
+class BicikeljStation(mongoengine.Document):
+    id = mongoengine.IntField(required=True)
```
mitar: `id` will not work. Use something else.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,28 @@
+{% extends "panel/panel.html" %}
+
+{% load i18n staticfiles sekizai_tags %}
+
+{% block content %}
+    {% addtoblock "css" %}<link href="{% static "piplmesh/panel/bicikelj/panel.css" %}" rel="stylesheet" type="text/css" media="screen" />{% endaddtoblock %}
+    <div id="bicikelj">
+        {% if not stations %}
+        <span class="bicikelj_error">{% trans "No Bicikelj stations nearby" %}</span>
```
mitar: Indent.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,28 @@
...
+        {% else %}
+           {% for station in stations %}
```
mitar: You can use [for-empty](https://docs.djangoproject.com/en/dev/ref/templates/builtins/#for-empty) construct instead.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,28 @@
...
+                <dl>
+                    <dt><span class="station_name">{{ station.name }}</span></dt>
+                    <dd>
+                        <ul>
+                            <li><span class="station_availability"><b>{{ station.available }}</b>{% trans " available bikes" %}</span></li>
```
mitar: This is not the right way to do translations which depends on the count. See [here](https://docs.djangoproject.com/en/dev/topics/i18n/translation/#blocktrans-template-tag).

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,27 @@
+from celery import task
+
+from pushserver.utils import updates
+
+from piplmesh.frontend import views
+from . import models, stations
+
+@task.task
+def update_station_info():
+    station_info = stations.fetch_data()
```
mitar: `station_info` -> `stations_data`

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+    for station in station_info:
+        if not models.BicikeljStation.objects(id=station['id'], timestamp=station['timestamp']):
```
mitar: You can use [get_or_create](http://mongoengine-odm.readthedocs.org/en/latest/apireference.html#mongoengine.queryset.QuerySet.get_or_create) instead.

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+            models.BicikeljStation(
+                                   name=station['name'],
+                                   address=station['address'],
+                                   id=station['id'],
+                                   location=station['location'],
+                                   timestamp=station['timestamp'],
+                                   available = station['available'],
```
mitar: Don't use spaces around `=`.

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+from piplmesh.frontend import views
+from . import models, stations
+
+@task.task
+def update_station_info():
+    station_info = stations.fetch_data()
+    for station in station_info:
+        if not models.BicikeljStation.objects(id=station['id'], timestamp=station['timestamp']):
+            models.BicikeljStation(
+                                   name=station['name'],
```
mitar: Indent only once, not to the `(`.

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+                                   timestamp=station['timestamp'],
+                                   available = station['available'],
+                                   free = station['free'],
+                                   total = station['total'],
+                                   open = station['open'],
+            ).save()
+            updates.send_update(
+                views.HOME_CHANNEL_ID,
+                # TODO
+                )
```
mitar: Yes, TODO. :-), But indent one less this last `)`.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,41 @@
+import math, urllib2, datetime
+from xml.dom import minidom
```
mitar: Order alphabetically.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,41 @@
...
+
+from . import models
+
+BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/'
+BICIKELJ_BOUNDS = 0.00965
+
+def get_stations_nearby(lat, lng):
```
mitar: Be verbose, write variables in full, `latitude` not `lat`.

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,12 @@
...
+    name = mongoengine.StringField(required=True)
+    address = mongoengine.StringField(required=True)
+    location = mongoengine.GeoPointField(required=True)
+    open = mongoengine.IntField()
```
mitar: Is not this a boolean value?

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,12 @@
+import mongoengine
+
+class BicikeljStation(mongoengine.Document):
```
mitar, line 3: Aren't all fields required?

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,41 @@
+import math, urllib2, datetime
+from xml.dom import minidom
+
+from . import models
+
+BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/'
```
mitar: This URL does not work for me.

piplmesh/panels/bicikelj/models.py
mitar: (Like `station_id`.)

mitar: You can also configure, that `station_id` and `timestamp` should be unique together.

mitar: Also, add index on `station_id` and `timestamp`.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,41 @@
...
+BICIKELJ_BOUNDS = 0.00965
+
+def get_stations_nearby(lat, lng):
+    ids = []
+    while 1:
+        try:
+            ids.append(models.BicikeljStation.objects(location__near= [lat, lng], location__within_box=[(lat-BICIKELJ_BOUNDS,lng-BICIKELJ_BOUNDS),(lat+BICIKELJ_BOUNDS,lng+BICIKELJ_BOUNDS)], id__nin=ids).first().id)
+        except:
+            break
+    stations = [models.BicikeljStation.objects(id=n).order_by('-timestamp').first() for n in ids]
```
mitar: I don't understand what are you doing here, why you first manually do queries to the database to get IDs and then again?

Doing queries in a loop is very bad. Queries to the database cost! And the idea of query is that you do one query, exactly the one which gives you data you want, not that you do that in your code, but that you leave to database to do this.

So what is wrong with simply:

```
models.BicikeljStation.objects(location__near=(lat, lng), location__within_box=((lat-BICIKELJ_BOUNDS, lng-BICIKELJ_BOUNDS),(lat+BICIKELJ_BOUNDS,lng+BICIKELJ_BOUNDS)))
```

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,41 @@
...
+
+BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/'
+BICIKELJ_BOUNDS = 0.00965
+
+def get_stations_nearby(lat, lng):
+    ids = []
+    while 1:
+        try:
+            ids.append(models.BicikeljStation.objects(location__near= [lat, lng], location__within_box=[(lat-BICIKELJ_BOUNDS,lng-BICIKELJ_BOUNDS),(lat+BICIKELJ_BOUNDS,lng+BICIKELJ_BOUNDS)], id__nin=ids).first().id)
```
mitar: Use tuples for constant length lists, not lists.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,41 @@
+import math, urllib2, datetime
+from xml.dom import minidom
+
+from . import models
+
+BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/'
+BICIKELJ_BOUNDS = 0.00965
```
mitar: Bounds are in both directions of the same size?

piplmesh/panels/bicikelj/stations.py
lazarm: You have to insert a number at the end of URL to get info about specific station. It works for me?

mitar: Aaa. Then you should make it a template:

```
BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/%d'
```

So it is clear that it is a template.

piplmesh/panels/bicikelj/stations.py
lazarm: Well, if I'm correct, we have a lot of station models in our database with the same id, but different timestamps. So if we use the query you mentioned above, we would get all stations nearby, but not just one object for each one of them. We would get objects with specific station_id, but with different timestamps. So I first wanted to find all stations nearby, and then retrieve the most updated info about them with ordering by timestamp. I know it's time consuming, I just didn't think of anything better.

mitar: This is why you need another field which will tell you when was this value fetched. Then you limit by that field (you don't want stale data) and order by it.

mitar: So their timestamps is nice to have for statistics purposes, but for us, we would like to know if our data is stale and which data is the most recent.

piplmesh/panels/bicikelj/stations.py
lazarm: Yes, because our map has the same width and height (actually, this number is computed from width = 450px, but I see now that our map has width=260px, was it changed lately?). Anyway, should I write in comments what is a formula based on map width and zomm level?

mitar: I am not aware of width being changed. Anyway, prepare bounds in a way that it supports maps of non-square sizes. And yes, write a comment to tell how this value is computer if there will ever be a need to change it.

piplmesh/panels/bicikelj/models.py
mitar, line 3: You should define also a field where you will store time when this data was fetched/updated. By that you can see if data is current. If I understand correctly, timestamp just tells you when the information about the station has changed. Not how recent the data is. So we need to keep track of how recent the information is ourselves. And if by any reason (our background job crashes, their website crashes) data is not recent enough, we should first display a warning if it just little stale data, or otherwise simply don't display it.

lazarm, line 3: So I should define a datetime field, which would get a value of current time when information would be fetched? But all stations wouldn't have the same value for this field, because there would be differences in miliseconds?
And how would the query in get_stations_nearby look like? We have to order twice, once for distance (to preserve order in distance from node) and once for this new field. We could first order by new field, retrieve just first 32 stations, and then order these by distance. But as some stations would be updated a few times in some period, and others never, we can't just take first 32 stations. I still can't see how the query could be simplified with this field.

lazarm:
Update.

piplmesh/panels/bicikelj/models.py
mitar, line 3: >  But as some stations would be updated a few times in some period, and others never, we can't just take first 32 stations.

I don't understand this. Every time you will fetch data for a station you will update this field. Even if it has the same value as before. The idea of this field is to tell that data there is not stale, not obsolete. So if there is no change in the station, if everything is still as it was, and you check this, this is new information, now you know that even now the state of station is like that, so you update the field.

Anyway, you probably don't need to order by this field, you just limit by it (for example, query only documents from last 2 minutes) and order by distance. Then in Python you check if you have any duplicates and use only the newer one. And of course, you can do that also on the server side, what would be better but you probably don't have time for that.

One further improvement would be to fetch for last 10 minutes, and then check what is the most recent there, and if is older then 1 minute you display warning about stale data.

And one more improvement would be that if if there is data older than 10 minutes, you don't display status of the station, but just that there is a station.

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,16 @@
...
+    station_id = mongoengine.IntField(required=True, unique_with='timestamp')
+    name = mongoengine.StringField(required=True)
+    address = mongoengine.StringField(required=True)
+    location = mongoengine.GeoPointField(required=True)
+    open = mongoengine.BooleanField(required=True)
+    available = mongoengine.IntField(required=True)
+    free = mongoengine.IntField(required=True)
+    total = mongoengine.IntField(required=True)
+    timestamp = mongoengine.DateTimeField(required=True, unique_with='station_id')
```
mitar: Move timestamp just bellow the `station_id`.

piplmesh/panels/bicikelj/models.py
```diff
@@ -0,0 +1,16 @@
...
+
+    meta = {
+        'indexes': ['station_id', 'timestamp'],
```
mitar, line 16: Use tuples for constant lists.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,46 @@
+import datetime, urllib2
+from xml.dom import minidom
+
+from . import models
+
+BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/%d'
+"""
```
mitar: Please format it as a comment, with some text explaining that it is a comment and what it does.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,46 @@
...
+BOUNDS_LONGITUDE = map_width*0.703119412486786/2^(map_zoom+1)
+"""
+BICIKELJ_BOUNDS_LATITUDE = 0.0027895
+BICIKELJ_BOUNDS_LONGITUDE = 0.0027895
+
+def get_stations_nearby(latitude, longitude):
+    ids = []
+    while 1:
+        try:
+            ids.append(models.BicikeljStation.objects(location__near= (latitude, longitude), location__within_box=((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)), id__nin=ids).first().id)
```
mitar: TODO.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,26 @@
...
+                <dt><span class="station_name">{{ station.name }}</span></dt>
+                <dd>
+                    <ul>
+                        <li><span class="station_availability">{% blocktrans %}<b>{{ station.available }}</b> available bikes{% endblocktrans %}</span></li>
+                        <li><span class="station_free">{% blocktrans %}<b>{{ station.free }}</b> free places %}{% endblocktrans %}</span></li>
+                    </ul>
+                </dd>
+            </dl>
+        {% empty %}
+            <span class="bicikelj_error">{% trans "No Bicikelj stations nearby" %}</span>
```
mitar, line 22: This is not true. One is that we don't have recent data bout them, the other is that there are not stations nearby.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,26 @@
...
+
+{% block content %}
+    {% addtoblock "css" %}<link href="{% static "piplmesh/panel/bicikelj/panel.css" %}" rel="stylesheet" type="text/css" media="screen" />{% endaddtoblock %}
+    <div id="bicikelj">
+        {% for station in stations %}
+            <dl>
+                <dt><span class="station_name">{{ station.name }}</span></dt>
+                <dd>
+                    <ul>
+                        <li><span class="station_availability">{% blocktrans %}<b>{{ station.available }}</b> available bikes{% endblocktrans %}</span></li>
```
mitar: This way of translation does not work, I believe.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,26 @@
...
+                        <li><span class="station_free">{% blocktrans %}<b>{{ station.free }}</b> free places %}{% endblocktrans %}</span></li>
+                    </ul>
+                </dd>
+            </dl>
+        {% empty %}
+            <span class="bicikelj_error">{% trans "No Bicikelj stations nearby" %}</span>
+        {% endfor %}
+    </div>
+    <script type="text/javascript">
+        /* TODO
```
mitar: O no. You will do an external file for JavaScript.

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,26 @@
...
+    </script>
+
```
mitar: Unnecessary empty line.

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,21 @@
...
+
+from piplmesh.frontend import views
+from . import models, stations
+
+@task.task
+def update_station_info():
+    stations_data = stations.fetch_data()
+    for station in stations_data:
+        station_object, created = models.BicikeljStation.objects.get_or_create(
+            station_id=station[0]['station_id'],
```
mitar: OK, so this is strange. Splitting result from fetch into two parts would be interesting approach if it would make you not need to access elements manually here. It would be interesting because it would allow split of logic between fetch_data and this task definition. That this task definition would not need to know about which fields are there.

But how you defined it, this is not done so. So, you can do two things.

Simplify everything and access filed directly for query:

```
models.BicikeljStation.objects.get_or_create(station_id=station['station_id'], timestamp=station['timestamp], default=station)
```

Or live data split, but then do something like (untested):

```
models.BicikeljStation.objects.get_or_create(default=station[1], **station[0])
```

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,21 @@
+from celery import task
+
+from pushserver.utils import updates
+
+from piplmesh.frontend import views
+from . import models, stations
+
+@task.task
+def update_station_info():
+    stations_data = stations.fetch_data()
```
mitar: Why have this variable? Why not just for loop over the result of the function call?

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,21 @@
...
+    for station in stations_data:
+        station_object, created = models.BicikeljStation.objects.get_or_create(
+            station_id=station[0]['station_id'],
+            timestamp=station[0]['timestamp'],
+            defaults = station[1]
```
mitar: Be precise with spaces around `=`. This is function call, no spaces. And comma at the end.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,46 @@
...
+    stations = [models.BicikeljStation.objects(id=n).order_by('-timestamp').first() for n in ids]
+    return stations
+
+def fetch_data():
+    stations_data = urllib2.urlopen(BICIKELJ_STATIONS_URL).read()
+    stations_xml_data = minidom.parseString(stations_data)
+    stations_nodes = stations_xml_data.getElementsByTagName('marker')
+    info_data = [urllib2.urlopen(BICIKELJ_INFO_URL % number).read() for number in range(1,len(stations_nodes)+1)]
+    info_xml_data = [minidom.parseString(info_data[i]) for i in range(len(stations_nodes))]
+    data = [
```
mitar: Why are you generating list here? It could just be an iterator.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,46 @@
...
+        [{
```
mitar: Tuples, not lists. And please indent better.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            ids.append(models.BicikeljStation.objects(location__near= (latitude, longitude), location__within_box=((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)), id__nin=ids).first().id)
+        except:
+            break
+    stations = [models.BicikeljStation.objects(id=n).order_by('-timestamp').first() for n in ids]
+    return stations
+
+def fetch_data():
+    stations_data = urllib2.urlopen(BICIKELJ_STATIONS_URL).read()
+    stations_xml_data = minidom.parseString(stations_data)
```
mitar: minidom can parse directly from file descriptor, no need for first converting it to string and then parsing it.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,46 @@
...
+    stations_nodes = stations_xml_data.getElementsByTagName('marker')
+    info_data = [urllib2.urlopen(BICIKELJ_INFO_URL % number).read() for number in range(1,len(stations_nodes)+1)]
```
mitar: No, iterate over `number` values. There might me some number missing.

mitar: So `number` values from fetched data.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,46 @@
...
+        [{
+            'station_id': int(stations_nodes[number].attributes['number'].value),
+            'timestamp': datetime.datetime.fromtimestamp(int(info_xml_data[number].getElementsByTagName('updated')[0].firstChild.nodeValue)),
+        },
+        {
+            'name': stations_nodes[number].attributes['name'].value,
+            'address': stations_nodes[number].attributes['address'].value,
+            'location': [float(stations_nodes[number].attributes['lat'].value), float(stations_nodes[number-1].attributes['lng'].value)],
+            'available': int(info_xml_data[number].getElementsByTagName('available')[0].firstChild.nodeValue),
+            'free': int(info_xml_data[number].getElementsByTagName('free')[0].firstChild.nodeValue),
```
mitar: You should maybe use xpath or something. This is ugly.

piplmesh/panels/bicikelj/models.py
lazarm, line 16: If I use a tuple, I get an error:

TypeError: Error when calling the metaclass bases
    can only concatenate tuple (not "list") to tuple

mitar, line 16: Hm, OK. Grr, why they use lists ...

piplmesh/panels/bicikelj/tasks.py
lazarm: So in the first case, it doesn't matter if station_id and timestamp appear twice in get_or_create? I thought this is not possible, that's why i split it into two parts.

mitar: I don't know. Try it. ;-) But I don't know why it would matter, if the value is  the same. It could be strange if the value would not be the same.

piplmesh/panels/bicikelj/stations.py
lazarm: You mean a missing attribute number or a missing value for attribute number? If I iterate over number values stored in a list, and converted to int, conversion could fail in a case when value is missing, because it would try to convert empty string.

mitar: ???

I am saying, that it is not necessary that all stations in range(1, len(stations_nodes)+1) exist in Ljubljana. You should iterate over stations_nodes to see which stations exist.

piplmesh/panels/bicikelj/tasks.py
lazarm: OK, it doesn't complain :-) In what format field open has to be stored, since it's boolean? If I use int, I get validation error.

mitar: `True` or `False`? :-) Try simply `bool()` on the value.

piplmesh/panels/bicikelj/models.py
lazarm, line 3: So we will have to compare current datetime with fetched time in a django template to group these situations properly. How to   import datetime in a django template?

mitar, line 3: Why in Django template? You do this in view. In Python code.

piplmesh/panels/bicikelj/stations.py
lazarm: So is this way ok:

>  for node in station_nodes:
> info_data = urllib2.urlopen(BICIKELJ_INFO_URL % int(node.attributes['number'].value))
> data.append ({ fields... })

piplmesh/panels/bicikelj/models.py
lazarm, line 3: Inside get_context function? So I check for every station nearby, if it's fetched time is older than 1 or 10 minutes, save it's station_id into two seperate lists (one for 1 minute error, and one for 10 minutes error), and then use these lists in panel.html  to properly show info?

piplmesh/panels/bicikelj/stations.py
mitar: This is good.

piplmesh/panels/bicikelj/models.py
mitar, line 3: No, you can have only one list given to template. In Python you do a query to the database and remove duplicates (retain only the newest one). Then you can just add additional attribute to Python object with a check for stale date you want to do:

```
station.old_data = station.time (or how you will name the field, see how we named similar fields elsewhere) < now - 2 * (interval from settings, 60 seconds currently)
```

Then in template, display this data, and based on the flag, add a warning sign about stale data.

lazarm, line 3: Thanks.
So I'll limit the query to fetch for last 12 minutes, and then add two flags, one for data older than 2 minutes and one for data older than 10 minutes, ok?

lazarm:
Update

piplmesh/panels/bicikelj/models.py
mitar, line 3: Why to flags? You have three possibilities:
- show station and current data
- show station and old data
- show just station (so that we know, that there is a station)

So what exactly do you get by 12 minutes and 10 minutes limit? What means if something is between 10 and 12 minutes? If you think of trying to  get locations with this approach, it will not work, because if last update was older than 12 minutes, you will not get anything. So why 12 minutes?

You can have a cut at 10 minutes for old 1.) and 2.) above. If you want to support 3.), you have to do it through aggregation in database.

lazarm, line 3: Okay, I left a cut at 10 minutes, and added TODO for data older than 10 minutes.

lazarm:
update

piplmesh/panels/bicikelj/static/piplmesh/panel/bicikelj/panel.js
```diff
@@ -0,0 +1 @@
+/* TODO: attach DOM elements to stations and deal with pushed updates
```
mitar: This comment is not closed at the end with `*/`.

mitar: What do you mean by "attach DOM elements to stations"?

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
+import urllib2
+from datetime import date, datetime, timedelta
+from lxml import etree, objectify
```
mitar: lxml is additional library, not built-in. Move to separate part.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+from pytz import UTC
```
mitar: You should use `django.utils.timezone.now()`. You probably want that?

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+
+from piplmesh import settings
+from . import models
+
+BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/%d'
+
+""" A comment explaining how to calculate map bounds, so we can search for stations visible on map only
```
mitar: Don't do docsting, a comment please. With `#` at the beginning.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+            n = station.station_id
+    station_idx.append(len(stations_nearby_all)-1)
+    stations_nearby = [stations_nearby_all[i] for i in station_idx]
+    for station in stations_nearby:
+        station.old_data = station.fetch_time < UTC.localize(datetime.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL))
+    # TODO: check if some stations nearby have data older than 10 minutes. Then display just their names, without current information
+    return stations_nearby
+
+def fetch_data():
+    stations_tree = objectify.fromstring(urllib2.urlopen(BICIKELJ_STATIONS_URL).read())
```
mitar: Why not:

```
objectify.parse(urllib2.urlopen(BICIKELJ_STATIONS_URL))
```

You don't convert to string and back in this way?

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+    data = []
+    for node in stations_tree.markers.marker:
+        info_data = objectify.fromstring(urllib2.urlopen(BICIKELJ_INFO_URL % int(node.attrib['number'])).read())
```
mitar: Same.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+                'station_id': int(node.attrib['number']),
+                'timestamp': datetime.fromtimestamp(int(info_data.updated)),
+                'name': node.attrib['name'],
+                'address': node.attrib['address'],
+                'location': (float(node.attrib['lat']), float(node.attrib['lng'])),
+                'available': int(info_data.available),
+                'free': int(info_data.free),
+                'total': int(info_data.total),
+                'open': bool(info_data.open),
+                'fetch_time': datetime.now(),
```
mitar: `...timezone.now()`.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+    station_idx.append(len(stations_nearby_all)-1)
+    stations_nearby = [stations_nearby_all[i] for i in station_idx]
+    for station in stations_nearby:
+        station.old_data = station.fetch_time < UTC.localize(datetime.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL))
+    # TODO: check if some stations nearby have data older than 10 minutes. Then display just their names, without current information
+    return stations_nearby
+
+def fetch_data():
+    stations_tree = objectify.fromstring(urllib2.urlopen(BICIKELJ_STATIONS_URL).read())
+    data = []
```
mitar: Why are you doing list? Why not simply iterator and use `yield` for each of that records?

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+
+"""
+
+BICIKELJ_BOUNDS_LATITUDE = 0.00557895
+BICIKELJ_BOUNDS_LONGITUDE = 0.00557895
+
+def get_stations_nearby(latitude, longitude):
+    stations_nearby_all = models.BicikeljStation.objects(
+                location__near=(latitude, longitude),
+                location__within_box=((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)),
```
mitar: Spaces around all operators.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+                fetch_time__gt=datetime.now() - timedelta(seconds=10*settings.POLL_BICIKELJ_INTERVAL),
```
mitar: Move `10` to a top-level constant.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+BOUNDS_LONGITUDE = map_width*0.703119412486786/2^map_zoom
+
+"""
+
+BICIKELJ_BOUNDS_LATITUDE = 0.00557895
+BICIKELJ_BOUNDS_LONGITUDE = 0.00557895
+
+def get_stations_nearby(latitude, longitude):
+    stations_nearby_all = models.BicikeljStation.objects(
+                location__near=(latitude, longitude),
```
mitar: Indentation.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+                location__within_box=((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)),
+                fetch_time__gt=datetime.now() - timedelta(seconds=10*settings.POLL_BICIKELJ_INTERVAL),
+    )
+    n = stations_nearby_all[0].station_id
```
mitar: What if there is no elements in `stations_nearby_all`?

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,62 @@
...
+    station_idx = []
+    for j,station in enumerate(stations_nearby_all):
```
mitar: This code is strange. ;-)

You could do all this in one pass. You know that similar stations are close together. So you can use this property and simply compare it with previous. And if new one is newer, you use new one, otherwise you keep as best found match the previous one. It is like searching for max number in an array. Just that you do that for subruns of numbers.

So:

```
if not stations_nearby_all:
    return
newest_station = stations_nearby_all[0]
for station in stations_nearby_all[1:]:
    if station.station_id != newest.station_id:
        set_possible_old_flag(newest_station)
        yield newest_station
        newest_station = station
    elif station is newer than newest station:
        newest_station = station
set_possible_old_flag(newest_station)
yield newest_station
```

mitar: `set_possible_old_flag` is not the best name, but so that you know what you have to do there.

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,16 @@
...
+
+@task.task
+def update_station_info():
+    for station in stations.fetch_data():
+        station_object, created = models.BicikeljStation.objects.get_or_create(
+            station_id=station['station_id'],
+            timestamp=station['timestamp'],
+            defaults=station,
+        )
+        # TODO: if created, send an update to server
```
mitar: You should add that if it is not created, then you should update the timestamp!

mitar: Ehm, not timestamp, but our fetch time.

mitar:
OK, the rest looks good now.

piplmesh/panels/bicikelj/stations.py
lazarm: I don't know, but apparently the way to access elements is then also changed, in a line 

> for node in stations_tree.markers.marker

I get 'lxml.etree._ElementTree' object has no attribute 'markers'. Can it stay like it is? :P

piplmesh/panels/bicikelj/static/piplmesh/panel/bicikelj/panel.js
lazarm: I remember you saying we should attach additional data to stations. What should I write then?

piplmesh/panels/bicikelj/stations.py
lazarm: I never got to iterators in python. I can see now what I was missing all this time :)

lazarm:
I fixed everything but TODO in javascript and fromstring(..).read()

piplmesh/panels/bicikelj/static/piplmesh/panel/bicikelj/panel.js
mitar: You mean we should attach stations to DOM elements displaying stations? Like we are attaching posts in the wall to DOM elements displaying posts?

But this is implementation detail. You don't put this into TODO (but you can open a ticket, with summary what all has yet to be done after this pull request).

In the TODO we should just write that we should update displayed data from pushed updates.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,59 @@
+import urllib2
+from datetime import datetime, timedelta
+from django.utils import timezone
+
+from lxml import objectify
```
mitar, line 5: Mnja, order is not OK.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,59 @@
...
+# BOUNDS_LATITUDE = map_height*0.703119412486786/2^map_zoom
+# BOUNDS_LONGITUDE = map_width*0.703119412486786/2^map_zoom
+
+BICIKELJ_BOUNDS_LATITUDE = 0.00557895
+BICIKELJ_BOUNDS_LONGITUDE = 0.00557895
+
+def get_stations_nearby(latitude, longitude):
+    stations_nearby_all = models.BicikeljStation.objects(
+        location__near = (latitude, longitude),
+        location__within_box = ((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)),
```
mitar: Add spaces around operators.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,59 @@
...
+from django.utils import timezone
+
+from lxml import objectify
+
+from piplmesh import settings
+from . import models
+
+BICIKELJ_STATIONS_URL = 'http://www.bicikelj.si/service/carto'
+BICIKELJ_INFO_URL = 'http://www.bicikelj.si/service/stationdetails/ljubljana/%d'
+STALE_DATA_TIME = 10
```
mitar: Which unit is this? Comment!

piplmesh/panels/bicikelj/templates/panel/bicikelj/panel.html
```diff
@@ -0,0 +1,25 @@
...
+        {% for station in stations %}
+            <dl>
+                <dt><span class="station_name">{{ station.name }}</span></dt>
+                <dd>
+                    <ul>
+                        {% if station.old_data %}
+                            <span class="bicikelj_error">{% trans "WARNING: stale data" %}</span>
+                        {% endif %}
+                        <li><span class="station_availability">{% blocktrans with available=station.available %}<b>{{ available }}</b> available bikes{% endblocktrans %}</span></li>
+                        <li><span class="station_free">{% blocktrans with free=station.free %}<b>{{ free }}</b> free places{% endblocktrans %}</span></li>
```
mitar: This is better, closer to the correct one, but does not change translation based on the number of stands available ("1 prosto mesto", "5 prostih mest").

piplmesh/panels/bicikelj/tasks.py
mitar: TODO.

mitar: If you will not be updating fetch time here in this pull request, then it does not work properly! Because then all stations will display stale data, even if it is not stale, but just there were not changes in the station. You have to fix this here, cannot be a TODO.

piplmesh/panels/bicikelj/stations.py
mitar: So, spaces around `-` and commas in tuples, but not around `=` as this function arguments.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,59 @@
...
+        return
+    newest_station = stations_nearby_all[0]
+    for station in stations_nearby_all[1:]:
+        if station.station_id != newest_station.station_id:
+            newest_station.old_data = newest_station.fetch_time < timezone.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL)
+            yield newest_station
+            newest_station = station
+        elif station.fetch_time > newest_station.fetch_time:
+            newest_station = station
+    newest_station.old_data = newest_station.fetch_time < timezone.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL)
```
mitar: Spaces around `*`.

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,59 @@
...
+        location__near = (latitude, longitude),
+        location__within_box = ((latitude-BICIKELJ_BOUNDS_LATITUDE,longitude-BICIKELJ_BOUNDS_LONGITUDE),(latitude+BICIKELJ_BOUNDS_LATITUDE,longitude+BICIKELJ_BOUNDS_LONGITUDE)),
+        fetch_time__gt = timezone.now() - timedelta(seconds=STALE_DATA_TIME*settings.POLL_BICIKELJ_INTERVAL),
+    )
+    if not stations_nearby_all:
+        return
+    newest_station = stations_nearby_all[0]
+    for station in stations_nearby_all[1:]:
+        if station.station_id != newest_station.station_id:
+            newest_station.old_data = newest_station.fetch_time < timezone.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL)
```
mitar: Spaces. Please, go over all code and check such things yourself.

piplmesh/panels/bicikelj/panel.py
```diff
@@ -0,0 +1,17 @@
...
+
+from piplmesh import panels
+
+from . import stations
+
+class BicikeljPanel(panels.BasePanel):
+    def get_context(self, context):
+        context.update({
+            'header': _("Bicikelj stations"),
+            'stations': stations.get_stations_nearby(self.request.node.latitude, self.request.node.longitude)
```
mitar: Comma at the end.

mitar: (To allow easier adding new stuff.)

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,59 @@
...
+            yield newest_station
+            newest_station = station
+        elif station.fetch_time > newest_station.fetch_time:
+            newest_station = station
+    newest_station.old_data = newest_station.fetch_time < timezone.now() - timedelta(seconds=2*settings.POLL_BICIKELJ_INTERVAL)
+    yield newest_station
+    # TODO: check if some stations nearby have data older than 10 minutes. Then display just their names, without current information
+
+def fetch_data():
+    stations_tree = objectify.fromstring(urllib2.urlopen(BICIKELJ_STATIONS_URL).read())
```
mitar: Why you name result of `objectify` once `tree` and other time `data`? It should be the same.

piplmesh/panels/bicikelj/tasks.py
mitar: As you are not really interested in resulted object, you could simply do an upsert here? Check how this is done in horoscope. So it is the same trick, you tell for what you query and what are values you insert or update, and database inserts new or updates existing one.

piplmesh/panels/bicikelj/stations.py
mitar: You have not read documentation about `parse()`. Please compare `fromstring` and `parse` use in the [example here](http://lxml.de/objectify.html#creating-objectify-trees).

mitar:
Have you really tried this? Because for me it does not work as task is not loaded by Celery. ;-) There is a bug I fixed in main repository and you can pull in. But I am not sure how you tested all this then.

So those comments for style you can ignore, if you want. But you cannot ignore translation, unit comment, upsert, and remove `fromstring`.

lazarm:
Of course, I wouldn't push changes before I will check if runserver and celery don't complain anything. Celery loads my task without problems. Hmmm...? :-)

lazarm:
update

lazarm:
ping :-)

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,22 @@
...
+            set__address=station['address'],
+            set__location=station['location'],
+            set__available=station['available'],
+            set__free=station['free'],
+            set__total=station['total'],
+            set__open=station['open'],
+            set__fetch_time=station['fetch_time'],
+            upsert=True)
+
+        # TODO: use get_or_create to know the result of the query. If created, send an update to server, otherwise just update fetch time
```
mitar, line 22: Not really. You can check the flag from upsert and see if new entry was created. And if it was, you have already data available to push (`station` object).

piplmesh/panels/bicikelj/tasks.py
```diff
@@ -0,0 +1,22 @@
...
+from pushserver.utils import updates
+
+from piplmesh.frontend import views
+from . import models, stations
+
+@task.task
+def update_station_info():
+    for station in stations.fetch_data():
+        models.BicikeljStation.objects(station_id=station['station_id'], timestamp=station['timestamp']).update(
+            set__name=station['name'],
```
mitar, line 12: Is is really necessary to use `set` here? Something like:

```
models.BicikeljStation.objects(station_id=station['station_id'], timestamp=station['timestamp']).update(upsert=True, **station)
```

does not work?

mitar:
Except for some minor changes, it looks good!

piplmesh/panels/bicikelj/stations.py
```diff
@@ -0,0 +1,59 @@
+import urllib2
+from datetime import datetime, timedelta
+from django.utils import timezone
+
+from lxml import objectify
+
+from piplmesh import settings
```
mitar, line 7: Oh, this is completely wrong. I just noticed. You should never access settings directly, but through `django.conf.settings`.

piplmesh/panels/bicikelj/tasks.py
mitar, line 12: It seems it need to have `set`. :-)

piplmesh/panels/bicikelj/tasks.py
lazarm, line 22: I thought of that before I changed it to get_or_create. But how do you check the flag, in the same way as with get_or_create (with two variables, like "station, created")?

mitar, line 22: Update returns some number. I think it is the number of objects inserted or something like this.

mitar, line 22: Anyway. It seems code is far from good. It is just adding and adding new elements, because timestamp changes even if there is no change. And even if there is a change, it happens that it does not change.

mitar:
Check the changes I did.

piplmesh/panels/bicikelj/tasks.py
lazarm, line 22: I wasn't paying attention to this, because you assumed timestamp works as we thought it does, and I just thought you're probably right. I don't know what is its purpose then.

mitar, line 22: Yes. But I noticed this just by running the code for few minutes. You should also notice it!

Issue #210, opened 2012-05-30
Implement weather panel


mitar:
Implement weather panel on home page for today and future days. You can also include some nice forecast picture. Find some public domain daily source in English and Slovene, or if not public, Creative Commons, of if not commons, something you use. And at the bottom of the weather panel put attribution link to the source.

dcrystalj:
I found a way to do it, if is ok to you.
Use placefinder api (http://developer.yahoo.com/geo/placefinder/) to convert long. and lat. to yahoo "woeid". Then use this "woeid" to get weather from yahoo weather api  (http://developer.yahoo.com/weather/) ??? 
They allow 50.000 requests per day, and yahoo logo needs to be displayed

mitar:
Mnja mnja. I don't really like it. I don't know. How precise is? Weather APIs are probably a lot around. It would be great to find some precise ones. Especially such which can give you different values for each city in Slovenia. Not just for Ljubljana. We want to be microlocal here! That's the idea of our portal. That you get one information in one part of Ljubljana, and another in another part. Where is this information? Does anybody have it? Can we access it?

dcrystalj:
I couldn't found anything more precise. Even data on Slovenian pages like http://www.vreme-on.net/vremenska-napoved-slika.html or http://www.arso.gov.si/vreme/napovedi%20in%20podatki/vreme_si_w.html is not precise enough to satisfy what you want. I really don't know how you would get forecast on every location if there is no weather station? 
Ivan Pepelnjak said once that all web pages buy same source of weather data to display weather. I think cities in Slovenia are to small to have different forecast on different locations in same city.

mitar:
Then ask him which source we should use. ;-)

There is no raw data from weather stations in Slovenia available online? You could write to arso.gov.si. It is probably public data and should be provided. ;-)

dcrystalj:
I cannot make forecast based on statistic, coz i'm not meteorologist.
If i get current data from weather station i can display only current weather and not for future days ?

mitar:
OK, if it is not clear, we want to give two types of information in this panel: current data and forecast. You can get current data base on microlocation, and general forecast for the area.

dcrystalj:
[...]

PR #292, opened 2012-09-15
weather panel

dcrystalj:
https://github.com/wlanslovenija/PiplMesh/issues/210


dcrystalj:
please take a look

piplmesh/panels/weather/models.py
```diff
@@ -0,0 +1,30 @@
+import mongoengine
+
+class Weather(mongoengine.Document):
+    created = mongoengine.DateTimeField()
+    latitude = mongoengine.DecimalField()
+    longitude = mongoengine.DecimalField()
+    model_name = mongoengine.StringField()
+    meta = {
```
mitar, line 39: Empty line before this.

piplmesh/panels/weather/models.py
```diff
@@ -0,0 +1,30 @@
+import mongoengine
+
+class Weather(mongoengine.Document):
+    created = mongoengine.DateTimeField()
```
mitar: Probably some fields are required? And have to be unique with others, no?

piplmesh/panels/weather/models.py
```diff
@@ -0,0 +1,30 @@
...
+    model_name = mongoengine.StringField()
+    meta = {
+        'allow_inheritance': True
+    }
+
+class Precipitation(Weather):
+    date_from = mongoengine.DateTimeField(required=True)
+    date_to = mongoengine.DateTimeField(required=True)
+    precipitation = mongoengine.DecimalField()
+    symbol = mongoengine.IntField(min_value=1, max_value=23)
```
mitar: Probably better would be to define choices map between ints and some useful strings.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,45 @@
...
+import datetime
+
+from django.utils import translation
+from django.utils.translation import ugettext_lazy as _
+
+from piplmesh import nodes, panels
+
+from . import models, weather
+
+WEATHER_OBSOLETE = 6
```
mitar: Add comment with unit.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,45 @@
...
+
+class WeatherPanel(panels.BasePanel): 
+
```
mitar: Unnecessary empty line.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,45 @@
...
+        state = models.State.objects(latitude=latitude, longitude=longitude).order_by('+at')
+        
+        if datetime.datetime.now() > state[0].created + datetime.timedelta(hours=WEATHER_OBSOLETE):
+            context.update({
+                'error_obsolete': True,
+            })
+            return context
+
+        weather_objects = []
+        for i in range(0,3):
```
mitar: Space after `,`. Use something else instead of `i`. What is `i`? Minutes? Hours? Intervals?

I am also not sure if this is the right approach. What if some entry is missing in the database? Then it will select wrong entry.

I would do multiple database quires here. For each time new, where you limit entries that they have to be for future from then on and select first one available.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,45 @@
...
+            weather_object= {
```
mitar: Space before `=`.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,45 @@
...
+                'at': state[i*10].at,
+                'temperature':  state[i*9].temperature,
+                'symbol':  precipitation[i*18].symbol
```
mitar: Double spaces. Space around `*`.

mitar: Comma at the end.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,45 @@
...
+        precipitation = models.Precipitation.objects(latitude=latitude, longitude=longitude).order_by('+date_to')
+        state = models.State.objects(latitude=latitude, longitude=longitude).order_by('+at')
+        
+        if datetime.datetime.now() > state[0].created + datetime.timedelta(hours=WEATHER_OBSOLETE):
+            context.update({
+                'error_obsolete': True,
+            })
+            return context
+
+        weather_objects = []
```
mitar: It is also probably not necessary to use lists. You could create iterator.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,45 @@
...
+class WeatherPanel(panels.BasePanel): 
+    
+    def get_context(self, context):
+
+        latitude = nodes.get_node(self.request).latitude
+        longitude = nodes.get_node(self.request).longitude      
+        precipitation = models.Precipitation.objects(latitude=latitude, longitude=longitude).order_by('+date_to')
+        state = models.State.objects(latitude=latitude, longitude=longitude).order_by('+at')
+        
+        if datetime.datetime.now() > state[0].created + datetime.timedelta(hours=WEATHER_OBSOLETE):
```
mitar: This probably fails if there is no data returned in `state`?

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+from piplmesh import nodes
+
+from . import models, weather as weather_functions
+
+@task.task
+def update_weather():
+    """
+    Task which updates weather for all nodes.
+
+    Old horoscopes are currently left in the database.
```
mitar: `horoscopes?

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+                    latitude = node.latitude, 
+                    longitude = node.longitude,
+                    model_name = weather_object.meta.model.attrib['name'],  
+                    at = datetime.strptime(product.attrib['from'],"%Y-%m-%dT%H:%M:%SZ")
+                ).update(
+                    set__created = datetime.strptime(weather_object.attrib['created'], "%Y-%m-%dT%H:%M:%SZ"),
+                    set__latitude = node.latitude,
+                    set__longitude = node.longitude,
+                    set__model_name = weather_object.meta.model.attrib['name'],
+                    set__at = datetime.strptime(product.attrib['from'],"%Y-%m-%dT%H:%M:%SZ"),
```
mitar: Constant string. Space after `,`.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+        for product in weather_object.product.iterchildren():
+            if product.attrib['from'] == product.attrib['to']:
+                models.State.objects(
+                    created = datetime.strptime(weather_object.attrib['created'], "%Y-%m-%dT%H:%M:%SZ"),
+                    latitude = node.latitude, 
+                    longitude = node.longitude,
+                    model_name = weather_object.meta.model.attrib['name'],  
+                    at = datetime.strptime(product.attrib['from'],"%Y-%m-%dT%H:%M:%SZ")
+                ).update(
+                    set__created = datetime.strptime(weather_object.attrib['created'], "%Y-%m-%dT%H:%M:%SZ"),
```
mitar: Are you sure you have to repeat those values used in `objects` query? I think they are added automatically? No?

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+                    upsert=True
+                )
+            else:
+                models.Precipitation.objects(
+                    created = datetime.strptime(weather_object.attrib['created'], "%Y-%m-%dT%H:%M:%SZ"),
+                    latitude = node.latitude, 
+                    longitude = node.longitude,
+                    model_name = weather_object.meta.model.attrib['name'],
+                    date_from = datetime.strptime(product.attrib['from'],"%Y-%m-%dT%H:%M:%SZ"),
+                    date_to = datetime.strptime(product.attrib['to'],"%Y-%m-%dT%H:%M:%SZ")
```
mitar: Constant string, comma.

piplmesh/panels/weather/templates/panel/weather/panel.html
```diff
@@ -0,0 +1,27 @@
+{% extends "panel/panel.html" %}
+
+{% load i18n static sekizai_tags %}
+
+{% block content %}
+    {% addtoblock "css" %}<link href="{% static "piplmesh/panel/weather/panel.css" %}" rel="stylesheet" type="text/css" media="screen" />{% endaddtoblock %}
+    {% if error_data or error_obsolete %}
```
mitar, line 7: Where do you define `error_data`? You might above if `state` is empty and `state[0]` does not exist. But you currently don't do it.

piplmesh/panels/weather/templates/panel/weather/panel.html
```diff
@@ -0,0 +1,27 @@
...
+                        <li>
+                            <span>{{ weather.at|date:"DATE_FORMAT" }}</span>
+                            {{ weather.temperature }}°C
+                            <img src="http://api.yr.no/weatherapi/weathericon/1.0/?symbol={{ weather.symbol }};content_type=image/png" alt="weather icon" />
+                        </li> 
+                    {% endfor %}
+                </ul>
+            </li>
+            <li>
+                <p>{% blocktrans with date=created %}Forcast created on: {{ date }}{% endblocktrans %}</p>
```
mitar: `Forcast` -> `Forecast`.

piplmesh/panels/weather/templates/panel/weather/panel.html
```diff
@@ -0,0 +1,27 @@
...
+        <span class="error">{% trans "Weather data is temporarily unavailable." %}</span>
+    {% else %}
+        <ul id="weather">
+            <li>
+                <ul>
+                    {% for weather in weather_objects %}
+                        <li>
+                            <span>{{ weather.at|date:"DATE_FORMAT" }}</span>
+                            {{ weather.temperature }}°C
+                            <img src="http://api.yr.no/weatherapi/weathericon/1.0/?symbol={{ weather.symbol }};content_type=image/png" alt="weather icon" />
```
mitar: `weather icon` is translatable string.

piplmesh/panels/weather/weather.py
```diff
@@ -0,0 +1,11 @@
+from lxml import etree, objectify
+
+source_url = 'http://api.met.no/'
+
+def fetch_data(latitude, longitude):
+    weather_url = '%sweatherapi/locationforecast/1.8/?lat=%s;lon=%s' % (source_url, latitude, longitude)
+    parser = etree.XMLParser(remove_blank_text=True)
+    lookup = objectify.ObjectifyElementClassLookup()
+    parser.setElementClassLookup(lookup)
+    weather = objectify.parse(weather_url,parser).getroot()
```
mitar: Space after comma.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+                ).update(
+                    set__created = datetime.strptime(weather_object.attrib['created'], "%Y-%m-%dT%H:%M:%SZ"),
```
mitar: No spaces around `=` because it is function call argument.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+
+from celery import task
+from datetime import datetime
+
+from piplmesh import nodes
+
+from . import models, weather as weather_functions
+
+@task.task
+def update_weather():
```
mitar: Does this task really work for you?

dcrystalj: cannot really test because celery is not working on windows

mitar: How not working?

piplmesh/panels/weather/panel.py
dcrystalj: Isn't multiple queries slower than one query? 
Or is slower to compare data in python then in database?

piplmesh/panels/weather/models.py
dcrystalj: so should i match all fields together to make them unique?
I don't really get the point here

piplmesh/panels/weather/panel.py
mitar: Maybe it is slower, but it returns the right results. :-) So, what if some entry is missing in the database? One type of prediction is not made for one timestamp?

If you can make it better, please. :-)

piplmesh/panels/weather/models.py
mitar: Yes. The idea is that database helps you catch programmatic error. So if we have some assumptions about data in the database, it is useful to try to encode this into database logic. Our assumptions are that there is only one entry for each time period for given location. No?

piplmesh/panels/weather/tasks.py
dcrystalj: run only celery

[Tasks]
[2012-09-22 11:13:56,440: WARNING/MainProcess] celery@Crystal has started.
c:\Python27\Scripts\env_piplmesh\lib\site-packages\celery\loaders\default.py:64: NotConfigured: No '
celeryconfig' module found! Please make sure it exists and is available to Python.
  "is available to Python." % (configname, )))
No handlers could be found for logger "multiprocessing"
[2012-09-22 11:13:57,460: ERROR/MainProcess] Consumer: Connection Error: [Errno 10061] No connection
 could be made because the target machine actively refused it. Trying again in 2 seconds...

run all together, this error repeats constantly:
  File "c:\Python27\Scripts\env_piplmesh\lib\site-packages\celery\utils__init__.py", line 293, in g
et_cls_by_name
    return getattr(module, cls_name)
AttributeError: 'module' object has no attribute 'DjangoLoader'
No handlers could be found for logger "multiprocessing"

mitar: Are you running this in virtualenv? Have you installed all the things from requirements.txt? Which Python version you have?

dcrystalj:
please take a look

piplmesh/panels/weather/tasks.py
dcrystalj: yes i did it in virtualenv.

here are some logs when i run install..
pip install -r requirements.txt  -->  http://pastebin.com/Tf2qdMrV
pip install --upgrade -r requirements.txt --> http://www.sendspace.com/file/ghii7i

python = 2.7.3

as i talked to others, it also doesn't work for them

mitar: Aaa. Why they and you don't tell this sooner?

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,56 @@
...
+                    set__cloudiness=product.location.cloudiness.attrib['percent'],
+                    set__fog=product.location.fog.attrib['percent'],
+                    set__low_clouds=product.location.lowClouds.attrib['percent'],
+                    set__medium_clouds=product.location.mediumClouds.attrib['percent'],
+                    set__high_clouds=product.location.highClouds.attrib['percent'],
+                    upsert=True
+                )
+            else:
+                models.Precipitation.objects(
+                    created=datetime.strptime(weather_object.attrib['created'], "%Y-%m-%dT%H:%M:%SZ"),
```
mitar: Constant string. You should use `'`.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    
+    for node in nodes.get_all_nodes():   
+        weather_object = weather_functions.fetch_data(node.latitude, node.longitude)
+        for product in weather_object.product.iterchildren():
+            DATE_FROM = datetime.strptime(product.attrib['from'], "%Y-%m-%dT%H:%M:%SZ")
+            DATE_TO = datetime.strptime(product.attrib['to'], "%Y-%m-%dT%H:%M:%SZ")
+            if DATE_FROM == DATE_TO:
+                models.State.objects(
+                    created=datetime.strptime(weather_object.attrib['created'], "%Y-%m-%dT%H:%M:%SZ"),
```
mitar: Constant string. You should use `'`.

piplmesh/panels/weather/models.py
```diff
@@ -0,0 +1,58 @@
...
+    
+    meta = {
+        'allow_inheritance': True,
+    }
+
+class Precipitation(Weather):
+    date_from = mongoengine.DateTimeField(required=True, unique_with=('created', 'latitude', 'longitude', 'date_to'))
+    date_to = mongoengine.DateTimeField(required=True, unique_with=('created', 'latitude', 'longitude', 'date_from'))
+    precipitation = mongoengine.DecimalField(required=True)
+    symbol = mongoengine.IntField(choices=SYMBOLS, required=True)
```
mitar, line 47: You have here IntField, but above you have strings as keys, not numbers. Does this work?

piplmesh/panels/weather/models.py
```diff
@@ -0,0 +1,58 @@
+import mongoengine
+from django.utils.translation import ugettext_lazy as _
+
+SYMBOLS = (
+    ('1', _("Sun")),
```
mitar: So why is here `'1'` and not `1`?

piplmesh/panels/weather/panel.py
mitar: Space after `#`.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,47 @@
...
+
+from django.utils import translation
+from django.utils.translation import ugettext_lazy as _
+
+from piplmesh import nodes, panels
+
+from . import models, weather
+
+WEATHER_OBSOLETE = 6 #hours
+FORECAST_HOUR = 14 #2:00 PM
```
mitar: What does this mean? It is unclear from the name of the variable, or from the comment.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Task which updates weather for all nodes.
+
+    Obsolete data is currently left in the database.
+    """
+    
+    for node in nodes.get_all_nodes():   
+        weather_object = weather_functions.fetch_data(node.latitude, node.longitude)
+        for product in weather_object.product.iterchildren():
+            DATE_FROM = datetime.strptime(product.attrib['from'], "%Y-%m-%dT%H:%M:%SZ")
```
mitar: This is not a constant, why use all upper case?

piplmesh/panels/weather/weather.py
```diff
@@ -0,0 +1,11 @@
+from lxml import etree, objectify
```
mitar: I am not really sure if we need completely new file just for this short function? You could move this to tasks file?

piplmesh/settings.py
```diff
@@ -186,7 +186,8 @@
     'piplmesh.utils',
     'piplmesh.panels',
     'piplmesh.panels.horoscope', # To load manage.py command
-
+    'piplmesh.panels.weather',
```
mitar: Add a comment same as above.

piplmesh/panels/weather/tasks.py
mitar: Which command do you use to start celery?

mitar: So, I think it is normal that Celery does not work if not everything from requirements.txt is installed successfully. So, you should make sure that everything is installed successfully. Have you tried searching the Google how to install those problematic libraries on Windows?

mitar: This [could help](http://stackoverflow.com/questions/4635734/trying-to-build-the-basic-python-extension-example-fails-windows).

mitar: OK, I updated requirements.txt so that a newer version of billiard is there. It is written that it has some improvements for installation on Windows. Please try it (pull from main repository and run pip install on it).

dcrystalj: omg this is just great!! 
whole site became working =)

before update, I also had to manually remove the billiard folder in python to make things right

dcrystalj: but still some problem with pymongo

http://pastebin.com/Ch8mhN6v

dcrystalj: also still problem with time :/

i get error in horoscope:
- can't compare offset-naive and offset-aware datetimes *

mitar: You can install pymongo manually:

```
pip install pymongo==2.3 --install-option='--no_ext'
```

Does this work for you?

Can you comment here which exactly command for running celery are you using? I think it is time that we add instructions on how to run everything on Windows because there are problems again and again.

> can't compare offset-naive and offset-aware datetimes

Can you provide me stacktrace or more detailed information about where does this come from?

dcrystalj: yes it works!!

my script 

```
#!/bin/bash
source env_piplmesh/Scripts/activate
path="$PiplMesh/manage.py "
$path celerybeat &
$path celeryd --loglevel=info --concurrency=1 --maxtasksperchild=10 &
$path runpushserver &
$path runserver &
```

I copied it from here https://common.tnode.com/pad/p/ipri1

---

When i set my birtday i horoscope panel cant be loaded. I think i have that python bug:
 `if datetime.datetime.now() > horoscope.date + datetime.timedelta(days=HOROSCOPE_OBSOLETE):`

mitar: So what is the error you are getting?

(I updated documentation, if you have time, you can test if instructions there work.)

mitar: So please test everything and make it work. :-) Or ask if you are unable to make it work.

mitar: If you have not deleted PiplMesh database, I recommend that you do that. We have in last month migrated to offset-aware datetimes. It might be that you have offset-naive stored in the database.

dcrystalj: `Exception Type:  TypeError
Exception Value:    
can't compare offset-naive and offset-aware datetimes
Exception Location: c:\Users\Crystal\git\abbb\piplmesh\panels\horoscope\panel.py in get_context, line 56`

well i dropped all databases and i still have this problem (also tried manually clean data/db folder

piplmesh/panels/weather/models.py
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+class Precipitation(Weather):
+    date_from = mongoengine.DateTimeField(required=True, unique_with=('created', 'latitude', 'longitude', 'date_to'))
+    date_to = mongoengine.DateTimeField(required=True, unique_with=('created', 'latitude', 'longitude', 'date_from'))
+    precipitation = mongoengine.DecimalField(required=True)
+    symbol = mongoengine.IntField(choices=SYMBOLS, required=True)
+
+class State(Weather):
+    at = mongoengine.DateTimeField(required=True, unique_with=('created', 'latitude', 'longitude'))
```
dcrystalj: I have problem here because of inheritance i guess.
it fails on fetching data when at is null, so when saving precipitation at==null and vice versa. please help

mitar: Can you please describe it in more detail?

mitar: How "it fails"?

dcrystalj: `Exception Type:  DuplicateKeyError
Exception Value:    
E11000 duplicate key error index: PiplMesh.weather.$at_1_created_1_latitude_1_longitude_1  dup key: { : null, : new Date(1348481093000), : 46.5507723961, : 15.6441316009 }`

dcrystalj: if unique is removed there's no null value in database

piplmesh/panels/weather/tasks.py
mitar: Do you still have problems with this?

dcrystalj: yes.

here is whole error log :
http://www.rubik.si/klub/TypeError.htm

ok solved

mitar: Great! :-)

piplmesh/panels/weather/static/piplmesh/panel/weather/panel.css
```diff
@@ -0,0 +1,34 @@
...
+    font-size: x-small;
+}
+
+#weather {
+    font-size: 10px;
+    text-align: left;
+}
+
+#weather ul{
+    float: left;
```
dcrystalj: why is this float not working at all?

piplmesh/panels/weather/static/piplmesh/panel/weather/panel.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+#weather span{
```
mitar: Space before `{`.

piplmesh/panels/weather/static/piplmesh/panel/weather/panel.css
```diff
@@ -0,0 +1,34 @@
+.error {
+    font-size: x-small;
+}
+
+#weather {
+    font-size: 10px;
+    text-align: left;
+}
+
+#weather ul{
```
mitar: Space before `{`.

piplmesh/panels/weather/static/piplmesh/panel/weather/panel.css
mitar: What exactly do you want to achieve?

dcrystalj: for example from this 
http://shrani.si/f/2A/6a/2HnLPAZv/capture.png
to
http://shrani.si/f/3a/su/2XJ3ChsO/capture1.png

dcrystalj: ok solved with padding: 0

mitar: Good. So I just have to leave you alone and you solve all bugs. Good. :-)

mitar:
What is current state with other errors you reported? DuplicateKeyError? inheritance? null values? I am not sure what exactly is the problem with null value and `at`?

dcrystalj:
here is error log on unique fields
http://www.rubik.si/klub/dke.htm

dcrystalj:
please take a look to last commit, also if design is ok to you...

piplmesh/panels/weather/models.py
```diff
@@ -0,0 +1,59 @@
+import mongoengine
```
mitar: Order imports. Django is first, then MongoEngine, empty line in-between.

piplmesh/panels/weather/models.py
```diff
@@ -0,0 +1,59 @@
...
+    (17, _("Light cloud ( winter darkness )")),
+    (18, _("Light rain and sun")),
+    (19, _("Snow and sun ( used for winter darkness )")),
+    (20, _("Sleet, sun and thunder")),
+    (21, _("Snow, sun and thunder")),
+    (22, _("Light rain and thunder")),
+    (23, _("Sleet thunder")),
+)
+
+#TODO unique fields
```
mitar: Describe a bit more what is needed to be done here. `:` after TODO.

mitar: OK, here is solution how to do it: https://github.com/MongoEngine/mongoengine/issues/140

You have to define that indexes should be sparse.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,56 @@
...
+WEATHER_OBSOLETE = 6 # 6 hours
+FORECAST_HOUR = 14 # 14:00 o'clock or 2:00 PM
+
+class WeatherPanel(panels.BasePanel): 
+    def get_context(self, context):
+        context.update({
+            'header': _("Weather"),
+        })
+        
+        latitude = nodes.get_node(self.request).latitude
```
mitar: Hm, `self.request.node` should contain node? You do not have to manually access that again.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+from . import models
+
+source_url = 'http://api.met.no/'
+
+def fetch_data(latitude, longitude):
+    weather_url = '%sweatherapi/locationforecast/1.8/?lat=%s;lon=%s' % (source_url, latitude, longitude)
+    parser = etree.XMLParser(remove_blank_text=True)
+    lookup = objectify.ObjectifyElementClassLookup()
+    parser.setElementClassLookup(lookup)
+    weather = objectify.parse(weather_url,parser).getroot()
```
mitar: Space after comma.

dcrystalj:
updated

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,56 @@
...
+        return context
+
+def get_weather_content(latitude, longitude):
+    date = datetime.datetime.now()
+    for interval in range(0, 3):
+        state = models.State.objects(latitude=latitude, longitude=longitude, at__lte=date).order_by('-at').first()
+        precipitation = models.Precipitation.objects(latitude=latitude, longitude=longitude, date_from__lte=date, date_to__gte=date).order_by('-date_from').first()
+        weather_object = {
+            'at': state.at,
+            'temperature':  state.temperature,
```
mitar: Double spaces.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+
+@task.task
+def update_weather():
+    """
+    Task which updates weather for all nodes.
+
+    Obsolete data is currently left in the database.
+    """
+    
+    for node in nodes.get_all_nodes():
```
mitar: This is quite tight loop. Are there any limits on how many requests per second/minute can we do to them?

dcrystalj: yes, no more than 20 requests from the api per second. (http://api.met.no/conditions_service.html)

Can I just add sleep function?

piplmesh/panels/weather/models.py
```diff
@@ -0,0 +1,60 @@
...
+    (17, _("Light cloud ( winter darkness )")),
+    (18, _("Light rain and sun")),
+    (19, _("Snow and sun ( used for winter darkness )")),
+    (20, _("Sleet, sun and thunder")),
+    (21, _("Snow, sun and thunder")),
+    (22, _("Light rain and thunder")),
+    (23, _("Sleet thunder")),
+)
+
+#TODO: Some fields have to be unique to make sure that there is only one entry for each time period for given location.
```
mitar: OK, add link to that pull request. Or just do it like it is written there.

piplmesh/panels/weather/management/commands/updateweather.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def handle(self, *args, **options):
+        """
+        Forces Weather update.
+        """
+
+        verbosity = int(options['verbosity'])
+
+        if verbosity > 1:
+            self.stdout.write('Updating weather...\n')
```
mitar: Should be `"`. (Yes, also in horoscope is wrong.)

piplmesh/panels/weather/tasks.py
mitar: Hmm, this takes really long on my computer. Hmm.

piplmesh/panels/weather/templates/panel/weather/panel.html
```diff
@@ -0,0 +1,26 @@
...
+                        <ul>
+                            <li>{{ weather.at|date:"DATE_FORMAT" }}</li>
+                            <li><img src="http://api.yr.no/weatherapi/weathericon/1.0/?symbol={{ weather.symbol }};content_type=image/png" alt="{% trans "weather icon" %}" /></li>
+                            <li>{{ weather.temperature }}°C</li>
+                        </ul>
+                    </li> 
+                {% endfor %}
+            </ul>
+            <p>{% blocktrans with date=created %}Forecast created on: {{ date }}{% endblocktrans %}</p>
+            <span>{%trans "Based on data from met.no" %}</span>
```
mitar: This could be link. Before `trans` should be space.

Otherwise I am OK with design.

piplmesh/panels/weather/tasks.py
dcrystalj: on my to. Mongodb get's really nervous =)
Maybe we shouldn't really store all data to make things faster?

mitar: I would use [rage_limit](http://docs.celeryproject.org/en/latest/userguide/tasks.html#Task.rate_limit) support in Celery.

So we could use this task to generate many new tasks. For each latitude&longitude combination one. So define another task which takes latitude and longitude and fetches and stores data and define rate limit on that.

BTW, be careful, there might be duplicate latitude and longitude combinations if multiple nodes are on the same location. You should remove such duplications to make less requests.

From this task return the list of all subtasks you created. (If you cannot return tasks directly, try just their IDs.)

Then, in mange.py you can call eager version of task, so that it is run locally, and then get back all this spawned tasks (or their IDs) and then wait on them, nicely displaying progress. :-)

mitar: `rage_limit`, nice name. :-) So, `rate_limit` is correct.

mitar: http://docs.celeryproject.org/en/latest/userguide/canvas.html#groups

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,56 @@
...
+            })
+            return context   
+            
+        context.update({
+            'weather_objects': get_weather_content(latitude, longitude),
+            'created': state.created,
+        })
+        return context
+
+def get_weather_content(latitude, longitude):
```
mitar, line 45: I am a bit skeptical about this function. It looks strange. And the query above. But OK, leave it as it is. Or maybe add TODO saying to check and possibility optimize.

mitar:
So, it looks good. Those queries are a bit strange, but we can play with that later. Maybe really add TODO. (Maybe for me, to check it again when it will be pulled in.)

We should only really solve the rate limiting problem.

piplmesh/panels/weather/tasks.py
mitar: > Maybe we shouldn't really store all data to make things faster?

No. :-)

But it is really strange. We do not have so much data to store. This is strange.

Hm, or it is. 32k entries? Where from???

dcrystalj: there are about 120 documets created for one location per one day. (first 3 days then it's reduced) so on average 60
So  60docs \* 9 days \* 160 nodes ==  a lot!

mitar: I don't understand this. So how many times they make model prediction per day? How many values each prediction has for different time spans? How man time spans are for each prediction?

dcrystalj:
updated

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+
+def fetch_data(latitude, longitude):
+    weather_url = '%sweatherapi/locationforecast/1.8/?lat=%s;lon=%s' % (source_url, latitude, longitude)
+    parser = etree.XMLParser(remove_blank_text=True)
+    lookup = objectify.ObjectifyElementClassLookup()
+    parser.setElementClassLookup(lookup)
+    weather = objectify.parse(weather_url, parser).getroot()
+    return weather
+
+@task.task(rate_limit=20)
```
mitar: No, you should limit `update_weather` tasks, not `generate_weather_tasks`. The former should not access their server too many times.

mitar: Also, add a comment after decorator explaining the unit of rate limiting, with possible URL to terms of use, so that later on somebody else can update it, if necessary.

mitar:
To make less entries stored: store only predictions for future three days. But make this a top-level Python constant (for how many days data should be fetched and stored).

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+def generate_weather_tasks():
+    """
+    Task which updates weather for all nodes.
+
+    Obsolete data is currently left in the database.
+    """ 
+    list=[]
+    jobs=[]
+    for node in nodes.get_all_nodes():
```
mitar: Easier would be to simply create a set:

```
for latitude, longitude in {(node.latitude, node.longitude) for node in nodes.get_all_nodes()}:
```

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+
+@task.task(rate_limit=20)
+def generate_weather_tasks():
+    """
+    Task which updates weather for all nodes.
+
+    Obsolete data is currently left in the database.
+    """ 
+    list=[]
+    jobs=[]
```
mitar: Spaces around `=`.

piplmesh/panels/weather/static/piplmesh/panel/weather/panel.css
```diff
@@ -0,0 +1,41 @@
...
+    width: 95%;
+}
+
+#forecast li {
+    border: none;
+    display: table-cell;
+    margin-left: 5px;
+    padding: 0;
+    vertical-align: middle;
+    width: 31%;
```
mitar: Why exactly 31%? Such test-and-find numbers are dangerous. Try to find underlying fixed value.

piplmesh/panels/weather/static/piplmesh/panel/weather/panel.css
```diff
@@ -0,0 +1,41 @@
...
+    font-size: 7px;
+}
+
+#forecast ul {
+    border: 0;
+    display: table;
+    list-style: none;
+    margin: 0 auto;
+    text-align: center;
+    width: 95%;
```
mitar: Same here.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+    weather = objectify.parse(weather_url, parser).getroot()
+    return weather
+
+@task.task(rate_limit=20)
+def generate_weather_tasks():
+    """
+    Task which updates weather for all nodes.
+
+    Obsolete data is currently left in the database.
+    """
```
mitar, line 39: Empty line after docstring.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+    for node in nodes.get_all_nodes():
+        if (node.latitude, node.longitude) not in list:
+            list.append((node.latitude, node.longitude))
+            jobs.append(update_weather.s(node.latitude, node.longitude))
+    group(jobs).apply_async()
+
+@task.task    
+def update_weather(latitude, longitude):
+    """
+    Task which updates weather for one node.
```
mitar: For one location.

piplmesh/panels/weather/tasks.py
mitar: Rename to `weather_tasks`.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Obsolete data is currently left in the database.
+    """ 
+    list=[]
+    jobs=[]
+    for node in nodes.get_all_nodes():
+        if (node.latitude, node.longitude) not in list:
+            list.append((node.latitude, node.longitude))
+            jobs.append(update_weather.s(node.latitude, node.longitude))
+    group(jobs).apply_async()
```
mitar: `apply_async` creates a temporary task just to start all of those tasks. Because we are already in the task, we could just create all tasks here.

So it is important not to wait for tasks to complete in the task (this means we simply blocked one worker from doing useful work). But we can still spawn all those tasks here.

So, `group(weather_tasks)()`.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import absolute_import
+
+from celery import group, task
```
mitar: Import only modules.

piplmesh/panels/weather/tasks.py
mitar: And return the results (that is, a set of pending tasks objects).

piplmesh/panels/weather/management/commands/updateweather.py
```diff
@@ -0,0 +1,21 @@
...
+        """
+        Forces Weather update.
+        """
+
+        verbosity = int(options['verbosity'])
+
+        if verbosity > 1:
+            self.stdout.write("Updating weather...\n")
+
+        tasks.generate_weather_tasks.delay().wait()
```
mitar: This does not really wait for all tasks to finish. If you return a set of pending tasks with return, you can do something like:

```
        results = tasks.generate_weather_tasks()
        prev_completed = None
        while not results.ready():
            completed = results.completed_count()
            if prev_completed != completed:
                self.stdout.write("Completed %d/%d.\n" % (completed, len(results)))
                prev_completed = completed
            time.sleep(1)
```

So we call `generate_weather_tasks` directly, we do not invoke it over the worker (as it just spawns subtasks, we can afford that). Because of that, we get access to the result, and then we poll status of tasks.

dcrystalj:
updated

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,83 @@
+from __future__ import absolute_import
+
+import celery
+import datetime
+
+from celery import task
```
mitar: No no. We group by packages. So, first are built-in. Then Django. Then celery. Then ours.

piplmesh/settings.py
```diff
@@ -228,8 +229,11 @@
 
 CHECK_ONLINE_USERS_INTERVAL = 10 # seconds
 CHECK_FOR_NEW_HOROSCOPE = 6 # am every day
+CHECK_FOR_NEW_WEATHER = 30 # minutes
 POLL_BICIKELJ_INTERVAL = 60 # seconds
 
+WEATHER_FORECAST = 3 # days
```
mitar: I thought that it wold be a top-level constant in the code, not setting. Because we do not really have code which then adapts the display based on that. But ... we could have it! In fact, we could just change that `range()` call to use this setting, and then we could have dynamic number of forecast, based on days.

mitar: So, change that `3` into this setting.

And think if you want to leave this as a setting or top-level (module) constant.

mitar:

mitar:
We had to change way periodic tasks are run because they were piling up. Sorry for additional work. So now we do not define periodic tasks in settings.py anymore.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,59 @@
...
+
+class WeatherPanel(panels.BasePanel): 
+    def get_context(self, context):
+        context.update({
+            'header': _("Weather"),
+        })
+        
+        latitude = self.request.node.latitude
+        longitude = self.request.node.longitude
+        #TODO: Check and possibly optimize
```
mitar: We put space before `TODO` keyword.

piplmesh/panels/weather/panel.py
```diff
@@ -0,0 +1,59 @@
...
+            })
+            return context   
+   
+        context.update({
+            'weather_objects': get_weather_content(latitude, longitude),
+            'created': state.created,
+        })
+        return context
+
+#TODO: Check and possibly optimize
```
mitar: Same.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,83 @@
...
+@task.task
+def generate_weather_tasks():
+    """
+    Task which updates weather for all nodes.
+
+    Obsolete data is currently left in the database.
+    """ 
+
+    weather_tasks = []
+    for latitude, longitude in {(node.latitude, node.longitude) for node in nodes.get_all_nodes()}:
```
mitar, line 43: A comment discussing what we are doing here could be useful, above this line. For example:

```
# Fetching data only once per possible duplicate locations
```

mitar:
Otherwise it is great now! Super!

dcrystalj:
bump

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,90 @@
...
+    Task which updates weather for all nodes.
+
+    Obsolete data is currently left in the database.
+    """ 
+    
+    weather_tasks = []
+    # Fetching data only once per possible duplicate locations
+    for latitude, longitude in {(node.latitude, node.longitude) for node in nodes.get_all_nodes()}:
+        weather_tasks.append(update_weather.s(latitude, longitude))
+    return celery.group(weather_tasks)()@task.task(rate_limit=20) # 20 tasks per second. Limitation by the api http://api.yr.no/conditions_service.html
```
mitar: This line has invalid syntax.

You can see that by tests which have also failed because of this (GitHub is showing them just bellow the pull request description).

piplmesh/panels/weather/templates/panel/weather/panel.html
```diff
@@ -0,0 +1,26 @@
...
+        <span class="error">{% trans "Weather data is temporarily unavailable." %}</span>
+    {% else %}
+        <div id="weather">
+            <ul>
+                {% for weather in weather_objects %}
+                    <li id="forecast">
+                        <ul>
+                            <li>{{ weather.at|date:"DATE_FORMAT" }}</li>
+                            <li><img src="http://api.yr.no/weatherapi/weathericon/1.0/?symbol={{ weather.symbol }};content_type=image/png" alt="{% trans "weather icon" %}" /></li>
+                            <li>{{ weather.temperature }}°C</li>
```
mitar, line 18: Add a TODO comment that we should make temperature unit user-configurable.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,90 @@
...
+from celery.task import schedules
+
+from piplmesh import nodes
+from piplmesh.utils import decorators
+
+from . import models, panel
+
+CHECK_FOR_NEW_WEATHER = 30 # minutes
+
+source_url = 'http://api.met.no/'
```
mitar: This is also a constant. So write it with all-upper-case letters.

piplmesh/panels/weather/tasks.py
```diff
@@ -0,0 +1,90 @@
+from __future__ import absolute_import
+
+import celery
```
mitar: Move this import bellow to celery group.

mitar:
Otherwise it looks good. But we should be doing tests too. Check how I made tests for horoscope. You can make them, or open a ticket for them. You have done a lot already.

dcrystalj:
updated, still some bug in test(frozen) but i think it's not a problem in code

mitar:
Merged. Tests to be done with #326.

mitar:
It seems tests are still failing, there is still bug in the code. Grr. You should really try your code before pushing!

https://travis-ci.org/#!/wlanslovenija/PiplMesh/builds/2686585

dcrystalj:
maybe you haven't seen my comment https://github.com/wlanslovenija/PiplMesh/commit/e6c61828bfd883e438662d1a75652942b9692a9b 
i have problems again to run/install requirements

Issue #293, opened 2012-09-20
Allow staff users to switch/mock location


mitar:
Allow staff users to switch/mock location easily. Maybe there could be additional all present menu entry in the top right menu (next to logout) with drop-down options for staff users. This should be displayed and should work only for staff users.

This could be done with storing special entry into user's session which would override value returned by our nodes API.

If you have other ideas for what this drop-down menu could provide, suggest, implement, or open a ticket.

mitjap:
I will put drop-down menu in header.html and make all views (that include header.html) inherit new class that can provide all nodes from piplmesh.nodes.get_all_nodes(). Changing this drop-down would switch/mock location the way it is described above.

Is this solution ok?

mitar:
> I will put drop-down menu in header.html

Good.

> make all views (that include header.html) inherit new class that can provide all nodes from piplmesh.nodes.get_all_nodes()

Ehm, what? You probably have to learn about context processors.

mitar:
There is still user feedback missing that mocking is in progress and to which node it is locked. Probably just have to provide initial data to the form.

mitar:
Check what the view does if form validation fails.

PR #297, opened 2012-09-23
Issue #293

mitjap:
Mocking location is working but is not perfect. 
-Right now everyone can mock location. How do we know if user is authorized for such action.
-Translations are still not working. Any idea what I did wrong?
#293


mitar:
> Right now everyone can mock location. How do we know if user is authorized for such action.

As I described, only staff users should be allowed to do that. Check definition of user document, I think you will find an answer there.

mitar:
> Translations are still not working. Any idea what I did wrong?

Have you recompiled them? Read our documentation how to work with translations.

piplmesh/frontend/context_processors.py
```diff
@@ -3,6 +3,7 @@
 from django.utils import translation
 
 from piplmesh import urls
+from piplmesh import nodes
```
mitar: No need for two lines. You can merge these two imports into one line. Order alphabetically.

piplmesh/frontend/middleware.py
```diff
@@ -8,6 +8,7 @@ class NodesMiddleware(object):
     def process_request(self, request):
         outside_url = urlresolvers.reverse('outside')
         request.node = nodes.get_node(request)
+        request.nodes = nodes.get_all_nodes()
```
mitar: No, we do not want that EVERY request has to load all nodes. Staff users will be a big minority and why would you load huge data for everybody?

Also, middleware is not the right place to store data you want to use in templates. Please read about context processors.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -181,6 +181,12 @@ input.gsc-search-button:hover,
     height: 50px;
 }
 
+#header #mock_location {
+    position:absolute;
+    top: 5px;
+    right:300px;
```
mitar: Space after `:`.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -181,6 +181,12 @@ input.gsc-search-button:hover,
     height: 50px;
 }
 
+#header #mock_location {
+    position:absolute;
```
mitar: Ehm, absolute? Why? It should simply be another element in the header menu (next to login, logout, and other entries).

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -181,6 +181,12 @@ input.gsc-search-button:hover,
     height: 50px;
 }
 
+#header #mock_location {
```
mitar: HTML element IDs have to be unique over all page. There is no reason to say that uniquely identified element should be child of another niquely identified element. You have already uniquely identified it in the first place. :-)

piplmesh/frontend/templates/header.html
```diff
@@ -21,6 +21,18 @@
                 {% trans "See your query results on " %}<a id="google_results_link">Google</a>.
             {% endif %}
         </div>
+        <div id="mock_location">
+            <form method="post" action="{% url "mock_location" %}" id="mock_location_form">
+                {% csrf_token %}
+                {% trans "Mock location" %}:
+                <select name="mocked_location" onchange="this.form.submit();">
```
mitar: No inline JavaScript. Use jQuery to attach event handler to elements.

piplmesh/frontend/templates/header.html
```diff
@@ -21,6 +21,18 @@
                 {% trans "See your query results on " %}<a id="google_results_link">Google</a>.
             {% endif %}
         </div>
+        <div id="mock_location">
+            <form method="post" action="{% url "mock_location" %}" id="mock_location_form">
```
mitar: Don't construct forms manually. Use Django forms for this. (Also that you learn how to use it.)

piplmesh/frontend/views.py
```diff
@@ -147,3 +148,15 @@ def panels_order(request):
         number_of_columns = request.GET['number_of_columns']
         panels = request.user.panels_order.get(number_of_columns, [])
         return http.HttpResponse(simplejson.dumps(panels), mimetype='application/json')
+
+def mock_location(request):
+    next = request.REQUEST.get('next', None)
```
mitar: Where do you set next? Why it would be configured? This is your job to set it.

mitar: And why mixing GET and POST? Why not have next URL as part of POST?

piplmesh/frontend/views.py
```diff
@@ -147,3 +148,15 @@ def panels_order(request):
...
+    if not next:
+        next = request.META.get('HTTP_REFERER', None)
```
mitar: We don't want to depend on browser features. If we want user to be redirected back where she came from, we should make sure we always have this data. And then it is not necessary to check browser headers for this.

piplmesh/frontend/views.py
```diff
@@ -147,3 +148,15 @@ def panels_order(request):
...
+    if not next:
+        next = '/'
+    response = http.HttpResponseRedirect(next)
+    if request.method == 'POST':
+        mock_location = request.POST.get('mocked_location', None)
```
mitar: You are not doing any verification here on the value POST. This is bad.

You should really be using Django forms.

piplmesh/frontend/views.py
```diff
@@ -147,3 +148,15 @@ def panels_order(request):
         number_of_columns = request.GET['number_of_columns']
         panels = request.user.panels_order.get(number_of_columns, [])
         return http.HttpResponse(simplejson.dumps(panels), mimetype='application/json')
+
+def mock_location(request):
```
mitar: Use class-based views. Not function views.

piplmesh/nodes/__init__.py
```diff
@@ -63,10 +64,22 @@ def get_node(request):
 
     node = None
     try:
-        node_id = request.session[SESSION_KEY]
-        backend_path = request.session[BACKEND_SESSION_KEY]
-        backend = load_backend(backend_path)
-        node = backend.get_node(node_id)
+        mocked_node_id = request.session[MOCKED_SESSION_KEY]
```
mitar: This code is similar to what I have in mind, but needs more improvement:
- you want this code to be fast as it is called on every request, iterating over all backends in not a good idea, this is why we store backend in the session
- so in the view above, you should mingle with session, not here, so in the view you trick the system to believe it is on that location you are mocking, and then leave this code here to blindly trust it
- so I would just override stored nodes_id and backend to the chosen one, I would only have a session value which tells that mocking is in effect so that you can provide user feedback to the staff user (showing that mocking is in effect and which node it is mocked to (the one in `request.node`))

piplmesh/nodes/__init__.py
```diff
@@ -7,6 +7,7 @@
 LONGITUDE_SESSION_KEY = '_longitude'
 
 SESSION_KEY = '_nodes_id'
+MOCKED_SESSION_KEY = '_nodes_id_mocked'
```
mitar: If you are adding new session keys, you should extend also `flush_session` function, and also check elsewhere if there is something similar to add too.

mitar:
Please update from main repository. It cannot be automatically merged.

piplmesh/frontend/forms.py
```diff
@@ -10,3 +12,8 @@ class ContactForm(forms.Form):
     subject = forms.CharField(label=_("Subject"))
     email = forms.EmailField(label=_("Your e-mail address"))
     message = forms.CharField(widget=widgets.Textarea())
+
+class LocationsForm(forms.Form):
+    locationChoices = [("%s_%s" % (backend.get_full_name(), node.id), node.name) for backend, node in nodes.get_all_nodes_with_backends()]
+    locationChoices.insert(0, ('-1', 'None'))
+    locations = forms.ChoiceField(choices=locationChoices, label=None, required=False)
```
mitjap: Initial value should be set here, but I'm not sure how to create constructor with arguments (so that request can be passed). What am I doing wrong:

```
def __init__(self, request):
    locationChoices = [("%s_%s" % (backend.get_full_name(), node.id), node.name) for backend, node in nodes.get_all_nodes_with_backends()]
    locationChoices.insert(0, ('-1', 'None'))
    init = get_initial_value_from_request
    locations = forms.ChoiceField(choices=locationChoices, initial=init, label=None, required=False)
    forms.Form.__init__(self)
```

piplmesh/frontend/context_processors.py
```diff
@@ -21,4 +20,7 @@ def global_vars(request):
         # Variables
         'logo_url': "piplmesh/images/logo-%s.png" % translation.get_language(),
         'request_get_next': request.REQUEST.get(auth.REDIRECT_FIELD_NAME),
+
+        # Forms
+        'locations_form' : forms.LocationsForm() if request.user and request.user.is_staff else None
```
mitjap: request would be passed here (see comment below)

piplmesh/frontend/static/piplmesh/js/common.js
```diff
@@ -22,6 +22,11 @@ $(document).ready(function () {
         }
     );
 
+    // Locations
+    $('#id_locations').change(function (event) {
+        $(this).closest("form").submit();
```
mitjap: Selecting an item should submit parent form. Is this the right way to do it?

mitar: Yes.

Use `'` for constant strings.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -181,6 +181,10 @@ input.gsc-search-button:hover,
     height: 50px;
 }
 
+#mock_location {
+    float: right;
```
mitar: I don't like floating here. It should be simple addition to the menu list in the header.

piplmesh/frontend/forms.py
mitar: I provided you already with an [example](https://github.com/wlanslovenija/PiplMesh/blob/master/piplmesh/account/forms.py#L60).

mitar:
Pull requests is failing tests. See above (Travis build). Please research why.

mitjap:
Tests are now working. Commit speaks for itself.

mitar:
This pull request cannot be automatically merged. Please update from main repository.

piplmesh/frontend/context_processors.py
mitar: Not really correct check. Much better check would be to check if user is authenticated.

And also add comma to the end.

```
'locations_form' : forms.LocationsForm() if getattr(request, 'user', None) and request.user.is_authenticated() and request.user.is_staff else None,
```

piplmesh/frontend/forms.py
```diff
@@ -1,6 +1,8 @@
+
```
mitar: Unnecessary empty line.

piplmesh/frontend/forms.py
```diff
@@ -1,6 +1,8 @@
...
 from django import forms
 from django.forms import widgets
 from django.utils.translation import ugettext_lazy as _
+from piplmesh import nodes
```
mitar, line 5: You should group imports. So this is another group, different than Django imports. So newline before it.

piplmesh/frontend/static/piplmesh/js/common.js
```diff
@@ -22,6 +22,11 @@ $(document).ready(function () {
         }
     );
 
+    // Locations
```
mitar: If we would be really strict, this code should be provided to users which are not staff.

Probably we have to define another JavaScript file which is included only for those users. We might later on even have more stuff there.

mitjap: (you have a typo in your comment) 

Never thought about that :) This solution would be good.

piplmesh/nodes/__init__.py
```diff
@@ -67,6 +68,10 @@ def get_node(request):
         backend_path = request.session[BACKEND_SESSION_KEY]
         backend = load_backend(backend_path)
         node = backend.get_node(node_id)
+
+        if allow_mocking and request.session[MOCKING_SESSION_KEY]:
```
mitar: You probably want here `request.session.get(MOCKING_SESSION_KEY, False)`.

piplmesh/frontend/static/piplmesh/js/common.js
mitar: Then do it. :-)

mitjap: In another ticket or this one?

mitar: What do you mean? On what? I am commenting what all you have yet to fix with this pull request.

piplmesh/nodes/__init__.py
mitjap: In case key is not present exception will happen and will not execute that statement. I'll fix it anyway.

piplmesh/frontend/templates/header.html
```diff
@@ -21,8 +21,15 @@
                 {% trans "See your query results on " %}<a id="google_results_link">Google</a>.
             {% endif %}
         </div>
+
         {% with next|default:request_get_next|default:request.get_full_path as next_url %}
             <ul class="user">
+                {% with locations_form as form %}
```
mitar: This should be displayed only to authenticated and staff users, no?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -250,13 +250,13 @@ function addComment(comment) {
     var post_url = $('.post').first().data('post').resource_uri;
 
     $.ajax({
-        type: 'POST',
+        'type': 'POST',
```
mitar, line 264: This all is completely unrelated stuff and would probably have to be another pull request. But leave it now.

piplmesh/frontend/views.py
```diff
@@ -64,6 +65,23 @@ class UserView(detail.DetailView):
     slug_field = 'username'
     slug_url_kwarg = 'username'
 
+class LocationsView(generic_views.FormView):
+    form_class = forms.LocationsForm
+    success_url = urlresolvers.reverse_lazy('home')
```
mitar, line 72: Not good. It should redirect back to initial page. This is not necessary home. You can add TODO here.

piplmesh/frontend/templates/header.html
mitjap: It does. If users does not have permission 'locations_form' will be undefined (None) and will not include that code. I don't like to duplicate same conditions. Its usually prone to errors.

piplmesh/frontend/views.py
mitar, line 72: initial page -> initiator page.

piplmesh/frontend/templates/header.html
mitar: It will still display `<li></li>`. No?

mitjap: I'll check. If it doesn't, do I leave it like it is?

mitar: So there should not be any output here. If there is any output, you should put if around. If there is none, then it is good.

piplmesh/frontend/views.py
```diff
@@ -64,6 +65,23 @@ class UserView(detail.DetailView):
...
+
+    def form_valid(self, form):
+        location = form.cleaned_data['locations']
+
+        if location == '-1':
+            nodes.flush_session(self.request)
+        else:
+            node_backend, node_id = location.rsplit('_')
+            self.request.session[nodes.SESSION_KEY] = int(node_id)
```
mitar: No, `node_id` is not necessary int. We do not define this anywhere in the API. It can be arbitrary string defined by the backend. So you should not convert it here.

piplmesh/frontend/views.py
```diff
@@ -64,6 +65,23 @@ class UserView(detail.DetailView):
...
+    form_class = forms.LocationsForm
+    success_url = urlresolvers.reverse_lazy('home')
+
+    def form_valid(self, form):
+        location = form.cleaned_data['locations']
+
+        if location == '-1':
+            nodes.flush_session(self.request)
+        else:
+            node_backend, node_id = location.rsplit('_')
```
mitar: As explained bellow, `node_id` could in theory contain `_`. I would suggest that you use `-` instead and search from left. Because Python modules and classes cannot contain `-` in their names. So `name_backend` cannot contain `-`.

mitar:
As I said, pull request cannot be merged. Please update from main repository.

mitar:
There is user feedback missing that mocking is in progress and to which node it is locked. You have to also do that. (Or open another ticket.) Probably you just have to provide initial data to the form.

piplmesh/frontend/forms.py
```diff
@@ -10,3 +12,8 @@ class ContactForm(forms.Form):
     subject = forms.CharField(label=_("Subject"))
     email = forms.EmailField(label=_("Your e-mail address"))
     message = forms.CharField(widget=widgets.Textarea())
+
+class LocationsForm(forms.Form):
+    locationChoices = [("%s_%s" % (backend.get_full_name(), node.id), node.name) for backend, node in nodes.get_all_nodes_with_backends()]
+    locationChoices.insert(0, ('-1', _("Don't mock location")))
```
mitar: This is ugly. Define outside function which returns iterator which provides all this data. So first it provides non-mocking value, then the rest.

mitar: You don't have to use `-1` value there. It could be simply an empty string. :-) Because it is strange to have strings for other values, and then number for this value. Empty string is much easier to understand: nothing should be mocked.

piplmesh/nodes/__init__.py
mitar: This if is not really correct. You probably want that if mocking is in effect (request.session[MOCKING_SESSION_KEY] is true) and `allow_mocking` is false, then this node value which was retrieved should not be allowed to be passed on! So you should set it to `None` in such case.

piplmesh/frontend/templates/header.html
mitjap: <li> tag (and some content that is hidden) is shown as you predicted. Will add a surrounding if around.

mitar:
Don't forget to update from main repository.

piplmesh/frontend/forms.py
```diff
@@ -2,6 +2,10 @@
 from django.forms import widgets
 from django.utils.translation import ugettext_lazy as _
 
+import itertools
```
mitar: Not needed.

piplmesh/frontend/forms.py
```diff
@@ -10,3 +14,12 @@ class ContactForm(forms.Form):
     subject = forms.CharField(label=_("Subject"))
     email = forms.EmailField(label=_("Your e-mail address"))
     message = forms.CharField(widget=widgets.Textarea())
+
+def locationChoices():
```
mitar: We use _ naming for functions.

piplmesh/frontend/forms.py
```diff
@@ -10,3 +14,12 @@ class ContactForm(forms.Form):
...
+    yield ('', _("Don't mock location"))
+    for backend, node in nodes.get_all_nodes_with_backends():
+        yield ("%s-%s" % (backend.get_full_name(), node.id), node.name)
```
mitar: For constant strings, we use `'`.

piplmesh/frontend/static/piplmesh/js/common_staff.js
```diff
@@ -0,0 +1,7 @@
+$(document).ready(function () {
+    // Locations
```
mitar: This comment is really unnecessary. It does not tell you anything new.

piplmesh/frontend/static/piplmesh/js/common_staff.js
```diff
@@ -0,0 +1,7 @@
...
+    $('#id_locations').change(function (event) {
+        $(this).closest('form').submit();
+    });
+
```
mitar: Unnecessary line.

piplmesh/frontend/templates/header.html
```diff
@@ -21,8 +21,17 @@
                 {% trans "See your query results on " %}<a id="google_results_link">Google</a>.
             {% endif %}
         </div>
+
         {% with next|default:request_get_next|default:request.get_full_path as next_url %}
             <ul class="user">
+                {% if locations_form %}
```
mitar: I would add here also `user.is_authenticated and user.is_staff` to the mix. Not that it is necessary, but to the reader tells easier when this thing goes through.

piplmesh/frontend/views.py
```diff
@@ -64,6 +65,25 @@ class UserView(detail.DetailView):
...
+    # TODO: redirect to initiator page
+    success_url = urlresolvers.reverse_lazy('home')
+
+    def form_valid(self, form):
+        location = form.cleaned_data['locations']
+
+        if location == '':
+            nodes.flush_session(self.request)
+        else:
+            node_backend, node_id = location.split('-')
```
mitar: You have to limit how many times it splits.

piplmesh/frontend/views.py
```diff
@@ -64,6 +65,25 @@ class UserView(detail.DetailView):
     slug_field = 'username'
     slug_url_kwarg = 'username'
 
+class LocationsView(generic_views.FormView):
+    form_class = forms.LocationsForm
+
+    # TODO: redirect to initiator page
```
mitar: Start TODO comments with upper-case letter.

piplmesh/frontend/views.py
```diff
@@ -64,6 +65,25 @@ class UserView(detail.DetailView):
     slug_field = 'username'
     slug_url_kwarg = 'username'
 
+class LocationsView(generic_views.FormView):
```
mitar, line 68: You are forgetting to validate if user is really authenticated and staff. Otherwise somebody can fake this request (especially because we are an open source project and can see how the request should look like).

piplmesh/nodes/__init__.py
```diff
@@ -65,8 +66,16 @@ def get_node(request):
     try:
         node_id = request.session[SESSION_KEY]
         backend_path = request.session[BACKEND_SESSION_KEY]
+        mocking = request.session.get(MOCKING_SESSION_KEY, False)
```
mitar: Move this to the end of this variables. It is the least important, or it is a special value.

piplmesh/nodes/__init__.py
```diff
@@ -65,8 +66,16 @@ def get_node(request):
...
         backend = load_backend(backend_path)
         node = backend.get_node(node_id)
+
```
mitar, line 72: Unnecessary empty line.

piplmesh/nodes/__init__.py
```diff
@@ -65,8 +66,16 @@ def get_node(request):
...
+
+        if allow_mocking and mocking:
+            return node
```
mitar, line 75: I would again verify here that user is authenticated and that is staff.

piplmesh/nodes/__init__.py
```diff
@@ -65,8 +66,16 @@ def get_node(request):
...
+        elif mocking:
+            node = None
```
mitar, line 78: A comment describing what are you doing here would be useful.

piplmesh/nodes/backends.py
```diff
@@ -23,12 +23,12 @@ def get_closest_node(self, request, latitude, longitude):
 
         node_id = random.randrange(len(data.nodes))
         node = copy.copy(data.nodes[node_id])
-        node.id = node_id
+        node.id = str(node_id)
```
mitar: No, there should be no need for this.

piplmesh/nodes/backends.py
```diff
@@ -37,5 +37,8 @@ def get_node(self, node_id):
     def get_all_nodes(self):
         for i, node in enumerate(data.nodes):
             node = copy.copy(node)
-            node.id = i
+            node.id = str(i)
```
mitar: No, there should be no need for this.

mitar: Caller should convert to string when necessary. ID object should provide a way to convert it to string.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -13,7 +13,7 @@ $(document).ready(function () {
         'mapTypeControlOptions': {
             'mapTypeIds': [
                 google.maps.MapTypeId.ROADMAP,
-                google.maps.MapTypeId.SATELLITE
+                google.maps.MapTypeId.SATELLITE,
```
mitar, line 18: See, because you mixed other things in your pull request, you now have to resolve conflicts with main repository. Pull from it and merge.

(You can fix such things locally, but there is no need to commit them into this pull request.)

piplmesh/frontend/context_processors.py
```diff
@@ -21,4 +20,6 @@ def global_vars(request):
         # Variables
         'logo_url': "piplmesh/images/logo-%s.png" % translation.get_language(),
         'request_get_next': request.REQUEST.get(auth.REDIRECT_FIELD_NAME),
-    }
+
+        # Forms
+        'locations_form' : forms.LocationsForm() if getattr(request, 'user', None) and request.user.is_authenticated() and request.user.is_staff else None,    }
```
mitar: `}` into new line.

piplmesh/frontend/views.py
```diff
@@ -3,10 +3,12 @@
 from django import dispatch, http, template
 from django.conf import settings
 from django.contrib import messages
+from django.contrib.auth.decorators import login_required
```
mitar: You are not using this?

piplmesh/frontend/views.py
```diff
@@ -3,10 +3,12 @@
...
 from django.core import mail, urlresolvers
 from django.core.files import storage
 from django.test import client
 from django.utils import simplejson
+from django.utils.decorators import method_decorator
```
mitar: You are not using this?

piplmesh/frontend/views.py
```diff
@@ -64,6 +67,33 @@ class UserView(detail.DetailView):
...
 
+class LocationsView(generic_views.FormView):
+    form_class = forms.LocationsForm
+
+    # TODO: Redirect to initiator page
+    success_url = urlresolvers.reverse_lazy('home')
+
+    def form_valid(self, form):
+        user = self.request.user
+
```
mitar: What is this? Some stale code. Clean!

piplmesh/frontend/views.py
```diff
@@ -64,6 +67,33 @@ class UserView(detail.DetailView):
...
+        else:
+            node_backend, node_id = location.split('-', 1)
+            self.request.session[nodes.SESSION_KEY] = node_id
+            self.request.session[nodes.BACKEND_SESSION_KEY] = node_backend
+            self.request.session[nodes.MOCKING_SESSION_KEY] = True
+
+        return super(LocationsView, self).form_valid(form)
+
+    def dispatch(self, request, *args, **kwargs):
+        staff = request.user
```
mitar: Don't rename variables. And probably you do not need to store request.user into a variable. Just access it directly.

piplmesh/frontend/views.py
```diff
@@ -64,6 +67,33 @@ class UserView(detail.DetailView):
...
+        if not (staff and staff.is_authenticated() and staff.is_staff):
```
mitar: In general, positive conditions is easier to think for humans. So:

```
if request.user and request.user.is_authenticated() and request.use.is_staff):
   return super..
raise PermissionDenied
```

piplmesh/nodes/__init__.py
```diff
@@ -67,6 +68,16 @@ def get_node(request):
         backend_path = request.session[BACKEND_SESSION_KEY]
         backend = load_backend(backend_path)
         node = backend.get_node(node_id)
+        mocking = request.session.get(MOCKING_SESSION_KEY, False)
+
+        # Return node if mocking is in progress and user is authenticated
+        if allow_mocking and mocking and request.user and request.user.is_authenticated() and request.user.is_staff:
+            return node
+        # If mocking was in progress and user isn't allowed to mock reset nodes value.
```
mitar: No `.` at the end.

mitjap: that dot looked like stain on my screen :)

piplmesh/frontend/context_processors.py
```diff
@@ -3,13 +3,12 @@
...
 
 from piplmesh import urls
-from piplmesh.frontend import tasks
+from piplmesh.frontend import forms, tasks
 
 def global_vars(request):
     """
     Adds global context variables to the context.
     """
-
```
mitar, line 12: Why removed this line?

piplmesh/frontend/forms.py
```diff
@@ -10,3 +12,12 @@ class ContactForm(forms.Form):
...
     message = forms.CharField(widget=widgets.Textarea())
+
+def location_choices():
+    yield ('', _("Don't mock location"))
+    for backend, node in nodes.get_all_nodes_with_backends():
+        yield ('%s-%s' % (backend.get_full_name(), node.id), node.name)
+
+class LocationsForm(forms.Form):
+    locations = forms.ChoiceField(choices=location_choices(), label='', required=False)
+
```
mitar: One line too much. :-)

piplmesh/frontend/views.py
```diff
@@ -4,6 +4,7 @@
 from django.conf import settings
 from django.contrib import messages
 from django.core import mail, urlresolvers
+from django.core.exceptions import PermissionDenied
```
mitar: Import only modules. So import just exceptions. And then access through that.

mitar:
Good. Otherwise looks good.

Issue #163, opened 2012-05-14
When a new wall post is made, update all clients


mitar:
When a new wall post is made, send an update with new post to all clients. Clients should get a button between input box and posts that so and so new posts are waiting. So that they can click and load them.

There could also be a checkbox, to enable autoloading of posts as they come.

mitar:
Storing the preference should still be added. So that preference between auto-loading or on-demand loading is stored into session. And next time this is the initial setting.

PR #298, opened 2012-09-23
User can prevent posts from keep updating issue #163

mitjap:
#163


mitar:
Please update from main repository. It cannot be automatically merged.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,6 +1,11 @@
 var POSTS_LIMIT = 20;
 var POSTS_DATE_UPDATE_INTERVAL = 60000; // ms
 
+// is new post automatically posted
+var CHANNEL_QUEUE_ENABLED = [];
```
mitar, line 5: This is not constant. Also, global variables should be kept to minimum.

mitar:
This is very complicated code. You could simply always add new posts to the list, only hide them. Because there are in general problems of possibility of duplicate posts. So it is much easier if you always add posts to the list, just that you do not display them if they are hidden. And when user clicks, you just display them.

mitar:
Please update/merge from main repository.

mitar:
This pull request is garbage. You will have to do it again.

mitar:
(Not the content, but how you merged.)

mitjap:
I totally agree :)

Issue #284, opened 2012-09-07
Improve handling of panels drag&drop


mitar:
Currently drag&drop of panels has quite some issues:
- Ajax requests to store their position is send many times while window resizing, it should be send only at the end
- window is currently much to high (it should collapse if columns are getting shorter because user moves a panel around)
- panels should not fluctuate around while user is resizing the window, but stay where they are and only rewrap if window becomes wide enough for more columns, but then again stay stable

One feature to discuss:
- if user moves panels around, should they also move in all other windows user might have open at the same time? in some way this would happen if she would reload those windows, as new state is stored, so doing that without reloading is plus, but on the other hand it will disallow user to have different configuration in different windows (maybe she wants one window with more funny stuff and one with more serious panels), but then this our state saving does not support anyway

PR #302, opened 2012-09-24
ajax request for ordering panels is send on after finished resizing only

dcrystalj:
#284


dcrystalj:
fixed only first issue 
- Ajax requests to store their position is send many times while window resizing, it should be send only at the end

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -216,10 +216,7 @@ $(document).ready(function () {
         });
     });
 
-    // TODO: Ajax request to store panels state is currently send many times while resizing, it should be send only at the end
-    $(window).resize(function (event) {
-        initializePanels();
-    });
+    $(window).resize( $.debounce( 250, initializePanels) );
```
mitar: Looks good. But there is unnecessary space around.

mitar: Hmm. But be careful. We probably do want that panels are repositioned automatically while window is resizing? Only we should not be doing Ajax requests to store the state until the very end.

mitar:
So this is a good idea. To use a library and to simply wrap in such a way. But you should do this only for Ajax requests.

dcrystalj:
Is not as easy as it might looks like. 
It's first time for me doing javascript =)

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -281,9 +281,10 @@ $(document).ready(function () {
         });
     });
 
-    // TODO: Ajax request to store panels state is currently send many times while resizing, it should be send only at the end
+    $(window).resize($.debounce(250, initializePanels));
     $(window).resize(function (event) {
-        initializePanels();
+        initializeEmptyColumnsForPanels();
+        orderPanelsDefault();
```
mitar, line 287: I think you also have to call `makeColumnsSortable`.

And `orderPanelsDefault` is not really the correct thing, no? It reorders thing.

mitar:
So this is not so easy. :-)

So when you are resizing, you should in fact reorder them properly when window becomes wide enough for another column. And the way you reorder that should be the same as it is stored in the database. So then you have to make an Ajax request. But just when number of columns change.

It seems it does not store anything at the end of the resizing.

So the problem is in fact just that it reads from the server too many times, not just when number of columns change. Sorry for confusion.

mitar:
Do you want to work more on this?

dcrystalj:
Not really, don't have time.

PR #303, opened 2012-09-24
Fixed some things. Started adding comments

mfrlin:
I've fixed some minor things and started adding comments. Would appreciate some feedback before I loose too much time heading in wrong direction. 

Can't figure out how to trigger addComment() when submiting a comment via generated form. Can't figure out how to trigger deletePost() via delete_link. JavaScript help desired :)


mitar:
Please update from main repository. This pull request cannot be merged automatically.

piplmesh/account/backends.py
```diff
@@ -75,9 +75,9 @@ def authenticate(self, facebook_access_token, request):
             # TODO: Do we know if all e-mail addresses given by Facebook are verified?
             # TODO: Does not Facebook support multiple e-mail addresses? Which one is given here?
             user.email = facebook_profile_data.get('email') or None
-        if user.gender is None:
-            # TODO: Does it really map so cleanly?
-            user.gender = facebook_profile_data.get('gender') or None
+        
+        user.gender = facebook_profile_data.get('gender') or None
```
mitar: Why this change?

piplmesh/account/fields.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, *args, **kwargs):
 
 class GenderField(mongoengine.StringField):
     def __init__(self, *args, **kwargs):
+        # TODO; Do we want to default gender to female? Wouldn't it be nicer to None?
```
mitar: Use `:` and not `;`.

Good TODO to ask.

mitar: But none == not required. So the question is if want this field to be required or not.

piplmesh/api/models.py
```diff
@@ -41,6 +43,8 @@ class Post(base.AuthoredDocument):
     def save(self, *args, **kwargs):
         self.updated_time = timezone.now()
         return super(Post, self).save(*args, **kwargs)
+    
+
```
mitar, line 49: What are those new lines?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -1,5 +1,6 @@
 var POSTS_LIMIT = 20;
-var POSTS_DATE_UPDATE_INTERVAL = 60000; // ms
+var POSTS_DATE_UPDATE_INTERVAL = 60000;
+// ms
```
mitar: Why newline?

piplmesh/api/models.py
```diff
@@ -32,6 +32,8 @@ class Post(base.AuthoredDocument):
 
     comments = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Comment), default=lambda: [], required=False)
     attachments = mongoengine.ListField(mongoengine.EmbeddedDocumentField(Attachment), default=lambda: [], required=False)
+    
+    is_edited = mongoengine.BooleanField(default=False)
```
mitar: What is this?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -31,7 +32,7 @@ function initializeEmptyColumnsForPanels() {
 function orderPanelsDefault() {
     var numberOfColumns = howManyColumns();
 
-    $('.panel').each(function (index, panel) {
+    $('.panel').each(function(index, panel) {
```
mitar: No, anonymous functions should have space after `function` part. [See our style example](https://github.com/wlanslovenija/PiplMesh/pull/199#commitcomment-1707793).

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -54,23 +55,22 @@ function sendOrderOfPanelsToServer() {
 
     if (numberOfColumns) {
         $.post(URLS.panels_order, {
-            'names': names,
-            'columns': columns,
-            'number_of_columns': numberOfColumns
+            'names' : names,
```
mitar: We don't have space before `:`. Why space?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -54,23 +55,22 @@ function sendOrderOfPanelsToServer() {
...
-    }, function (data, textStatus, jqXHR) {
+        'number_of_columns' : howManyColumns()
+    }, function(data, textStatus, jqXHR) {
         if (data.length == 0) {
             orderPanelsDefault();
-        }
-        else {
-            $.each(data, function (i, column) {
-                $.each(column, function (j, panel) {
+        } else {
```
mitar: We do not have `else` on the same line as `}`. Why this change?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -144,80 +148,107 @@ function Post(data) {
...
+
+        return comments;
+
+    }
+
+    function addComment() {
+        var message = $('#comment_text').val();
+        $.ajax({
+            'type' : 'POST',
+            'url' : '/api/v1/post/' + self.id + '/comments/',
```
mitar: Don't have hard-coded URLs.

mitar:
OK. Why code style changes? This is not really something you do in collaborative project where there is already established code style. So please explain and/or change back.

piplmesh/account/backends.py
mfrlin: because user gender defaults to female and doesnt take into consideration facebook gender because of the fi statement
if user.gender is None:

piplmesh/frontend/static/piplmesh/js/home.js
mfrlin: i tried the format function to get rid of the tabs in the code and it went south. ill fix those empty lines and newlines, didn't have the time yesterday.

piplmesh/api/models.py
mfrlin: If post is edited this will switch to true and show when is displayed (similar to facebook without message history)
that is if we will enable edit options. i thought its a good idea.

piplmesh/frontend/static/piplmesh/js/home.js
mfrlin: formatting did this. would be nice if i could revert to previous commit somehow?

piplmesh/frontend/static/piplmesh/js/home.js
mfrlin: formatting

piplmesh/frontend/static/piplmesh/js/home.js
mfrlin: formatting, i need to revert it to previous commit.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: Of course you can. Read git documentation/tutorials how to do it.

piplmesh/api/models.py
mitar: But it is unneeded until we have edit functionality. Why adding it now to the model?

And probably we want to keep history anyway, if we want to do it. So this could be a separate issue.

mitar:

piplmesh/account/backends.py
mitar: Yes. But the idea is that if user has already set some value, then we assume that she wants that value, even if it is different than Facebook's.

Probably the right solution here is to not make gender required and to have it assigned none value for default.

mitar:
Please update from main repository to be able to merge.

mitar:
OK. You should really learn how to merge and work with the repository.

This pull request became garbage. I am closing it. Please open new one, clean, and without those bad commits.

Also, in each pull request concentrate only on one thing. So, fixing some things should be one pull request. Comments another.

It is nothing wrong if pull request is not finished when you start it, make it to the point to which you know to do it and then ask for help. But this currently is simply unreadable so I do not even know what is wrong with it or where you have problems.

PR #304, opened 2012-09-24
horoscope now working properly

dcrystalj:
no more error:  can't compare offset-naive and offset-aware datetimes

### DIFF
--- a/piplmesh/panels/horoscope/panel.py
+++ b/piplmesh/panels/horoscope/panel.py
@@ -4,6 +4,7 @@ import datetime
 
 from django.utils import translation
 from django.utils.translation import ugettext_lazy as _
+from django.utils.timezone import utc
 
 from piplmesh import panels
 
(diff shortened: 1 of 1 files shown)


piplmesh/panels/horoscope/panel.py
```diff
@@ -53,7 +54,7 @@ def get_context(self, context):
             })
             return context
 
-        if datetime.datetime.now() > horoscope.date + datetime.timedelta(days=HOROSCOPE_OBSOLETE):
+        if datetime.datetime.utcnow().replace(tzinfo=utc) > horoscope.date + datetime.timedelta(days=HOROSCOPE_OBSOLETE):
```
mitar, line 57: Aaa. Noo. What, we had datetime.now??? How I missed this.

The solution should be:

```
from django.utils import timezone

timezone.now()
```

You do not have to complicate with replace of timezone.

mitar, line 57: And import only modules!

mitjap:
Thank you for fixing that :)

mitar:
So, does it work now also with my changes?

dcrystalj:
works like a charm =)

Issue #301, opened 2012-09-24
Support marking notifications as read


mitar:
Support marking notifications as read.

Related: #165

PR #309, opened 2012-09-27
Issue #301 - Support marking notifications as read

zupan:
Notifications can be marked as read. #301


zupan:
There is an error in Tastypie when using `PATCH` on resource where a related field has `full=True` parameter. This is known [issue](https://github.com/toastdriven/django-tastypie/issues/501) and it was fixed [here](https://github.com/toastdriven/django-tastypie/commit/e4de9377cb).

zupan:
This solution seems to be working.

mitar:
Great! I pushed change to my Tastypie fork. It is now in `requirements.txt`. Thanks for finding this out.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -231,6 +231,19 @@ function Notification(data) {
         $('#notifications_list').prepend(createDOM());
     };
 
+    self.markAsRead = function () {
+        $.ajax({
```
mitar: I would add here a test if notification is not already marked as read. So that we do not do unnecessary calls to the server.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -231,6 +231,19 @@ function Notification(data) {
         $('#notifications_list').prepend(createDOM());
     };
 
+    self.markAsRead = function () {
```
mitar: Some additional feedback would be useful to the user (like background color change of notification, similar to how Facebook has) and counter change. But this will probably come.

piplmesh/frontend/static/piplmesh/js/home.js
zupan: Fixed

piplmesh/frontend/static/piplmesh/js/home.js
zupan: Fixed

zupan:
Problem with related fields is still there. Hope to find solution in tests.

mitar:
Have you found error through tests?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -233,7 +233,7 @@ function Notification(data) {
         var format = gettext("%(author)s commented on post.");
         var author = interpolate(format, {'author': self.comment.author.username}, true);
 
-        var notification = $('<li/>').addClass('notification').data('notification', self).append(
+        var notification = $('<li/>').addClass('notification').addClass('unread_notification').data('notification', self).append(
```
mitar: Not necessary. You should add `unread_notification` class only if notification is really unread.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -412,6 +429,11 @@ $(document).ready(function () {
     $('#notifications_count').add('.close_notifications_box').click(function (event) {
         $('#notifications_box').slideToggle('fast');
     });
+
+    $('.notification').live('click', function (event) {
```
mitar: `live`? Why? I would use `.on` or simply add event handler in createDOM.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -259,6 +259,23 @@ function Notification(data) {
...
+    self.markAsRead = function (dom_element) {
+        if (!self.read) {
+            $.ajax({
+                type: 'PATCH',
+                url: self.resource_uri,
+                data: JSON.stringify({'read': true}),
+                contentType: 'application/json',
+                dataType: 'json',
+                success: function (data, textStatus, jqXHR) {
+                    $('#notifications_count').text(parseInt($('#notifications_count').text()) -1 );
```
mitar: Fix spaces at the end of the line.

mitar: There could be some race condition here. For example, user double-clicking the notification. You could happen to decrement twice. It would be better if you would count all notifications. Slower, but more precise.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -259,6 +259,23 @@ function Notification(data) {
...
+                    $(dom_element).removeClass('unread_notification');
+                    alert("Notification marked as read.");
```
mitar: I suggest you use `window.console.log` when developing. Also, always add TODO comment to remove it. So that we do not forget.

mitar:
So the test you made is failing?

mitar:
All tests pass for me? (8 tests in 97.395s) You have more tests? Something fails?

piplmesh/frontend/static/piplmesh/js/home.js
zupan: When developing I am using `console.log`. I was adding `alert` into commits because i saw them in other parts of the project. Will remove them in next commit.

mitar: We use alery only to inform the user about Ajax error. And you have it for "comment posted" feedback (which could be also `window.console.log`).

zupan: Fixed it in last commit.

piplmesh/frontend/static/piplmesh/js/home.js
zupan: Fixed

piplmesh/frontend/static/piplmesh/js/home.js
zupan: Fixed

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -233,7 +233,38 @@ function Notification(data) {
         var format = gettext("%(author)s commented on post.");
         var author = interpolate(format, {'author': self.comment.author.username}, true);
 
-        var notification = $('<li/>').addClass('notification').data('notification', self).append(
+        var notification = $('<li/>').addClass('notification').bind('click', function () {
```
mitar: Always write all arguments callback can get.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -233,7 +233,38 @@ function Notification(data) {
...
+            if (!self.read) {
+                $.ajax({
+                    type: 'PATCH',
+                    url: self.resource_uri,
+                    data: JSON.stringify({'read': true}),
+                    contentType: 'application/json',
+                    dataType: 'json',
+                    success: function (data, textStatus, jqXHR) {
+                        var unread_notifications_counter = 0;
+                        $.getJSON(URLS.notifications, function (data, textStatus, jqXHR) {
```
mitar: What? Noo noo. :-) Don't do another request. You should traverse DOM and `.data()`-attached notification objects and count there. So the idea is: notification comes, we convert it to object and attach it to DOM. Count is synced with how many notifications we have in the list which are not yet read. So everytime we mark a notification as read, we store that and update it in attached object. And update count. We have to read from DOM because store Ajax request is asynchronous. So we should keep state in attached objects.

zupan: Fixed

piplmesh/frontend/static/piplmesh/js/home.js
zupan: Fixed

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -233,7 +233,36 @@ function Notification(data) {
...
+                $.ajax({
+                    type: 'PATCH',
+                    url: self.resource_uri,
+                    data: JSON.stringify({'read': true}),
+                    contentType: 'application/json',
+                    dataType: 'json',
+                    success: function (data, textStatus, jqXHR) {
+                        var unread_notifications_counter = 0;
+                        $('.notification').each(function (i, notification) {
+                            if (!$(notification).data('notification').read) {
```
mitar, line 248: This is not correct. Where you mark the notification user just read as read? You have to set this before you count. :-) So in callback from request, you set read to true in data/notification object, and then you count what is the new number.

piplmesh/api/tests/test_basic.py
```diff
@@ -143,6 +143,83 @@ def test_basic(self):
...
+        post_uri = response['location']
+
+        response = self.client.get(post_uri)
+        self.assertEqual(response.status_code, 200)
+        response = json.loads(response.content)
+
+        self.assertEqual(response['message'], 'Test post for notifications.')
+        self.assertEqual(response['author']['username'], self.user_username)
+        self.assertNotEqual(response['created_time'], None)
+        self.assertNotEqual(response['updated_time'], None)
```
mitar: No need to test all this again, if they are tested already in other tests. Concentrate here on assertions needed for your test case, for notifications. So probably you have to check is published flag, and that there are no comments. You could also be dirty and go and directly access the database and check if subscriptions are as they should be.

piplmesh/api/tests/test_basic.py
```diff
@@ -143,6 +143,83 @@ def test_basic(self):
...
+        self.assertEqual(response['comments'], [])
+        self.assertEqual(response['attachments'], [])
+        self.assertEqual(response['is_published'], True)
+
+        post_created_time = response['created_time']
+        post_updated_time = response['updated_time']
+
+        # Delay so next update will be for sure different
+        time.sleep(1)
```
mitar: So no need for this, you are not checking if updated_time works correctly. This we will do in other tests.

piplmesh/api/tests/test_basic.py
```diff
@@ -143,6 +143,83 @@ def test_basic(self):
...
+        self.assertEqual(response.status_code, 200)
+        response = json.loads(response.content)
+
+        self.assertEqual(response['comments'][0], self.fullURItoAbsoluteURI(comment_uri))
+        self.assertEqual(response['created_time'], post_created_time)
+        self.assertNotEqual(response['updated_time'], post_updated_time)
+
+        # Checking notification
+
+        response = self.client.get(self.resourceListURI('notification'))
```
mitar, line 180: Now, here you have a problem. :-) Notifications is processed in the background, in Celery. So when you get here, you are too fast, notifications has not yet been processed. The idea is that client gets response as quickly as possible, so that we do not necessary block web Django process, and keep connection open. And then when notification is ready, we push.

But there is even one larger problem here, testing with background tasks is not really easy, you should not be required to run some background process to test things, and then wait.

So the solution would be, that just for the time of testing, we force all background tasks to be executed in response generation itself.

mitar, line 180: Ha ha, there [is one](http://celery.readthedocs.org/en/latest/configuration.html#celery-always-eager). :-)))

mitar, line 180: Combine it with [this](https://docs.djangoproject.com/en/dev/topics/testing/#overriding-settings) and you will know what to do. :-)

zupan:
Tests are finally breaking at the same error that causing problem in application when trying to mark notification as read.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -233,7 +233,37 @@ function Notification(data) {
...
+            if (!self.read) {
+                $.ajax({
+                    type: 'PATCH',
+                    url: self.resource_uri,
+                    data: JSON.stringify({'read': true}),
+                    contentType: 'application/json',
+                    dataType: 'json',
+                    success: function (data, textStatus, jqXHR) {
+                        var unread_notifications_counter = 0;
+                        self.read = true;
```
mitar, line 245: Swap two lines above around. `var unread_notifications_counter = 0;` is part of the for each loop.

zupan, line 245: Fixed

piplmesh/api/tests/test_basic.py
```diff
@@ -143,6 +143,66 @@ def test_basic(self):
...
+        post_uri = response['location']
+
+        response = self.client.get(post_uri)
+        self.assertEqual(response.status_code, 200)
+        response = json.loads(response.content)
+
+        self.assertEqual(response['comments'], [])
+        self.assertEqual(response['is_published'], True)
+
+        # Test authorization
```
mitar: Why this?

piplmesh/api/tests/test_basic.py
```diff
@@ -143,6 +143,66 @@ def test_basic(self):
...
+        self.assertEqual(response.status_code, 200)
+
+        # Adding a comment
+
+        comments_resource_uri = self.fullURItoAbsoluteURI(post_uri) + 'comments/'
+
+        response = self.client2.post(comments_resource_uri, '{"message": "Test comment 1."}', content_type='application/json')
+        self.assertEqual(response.status_code, 201)
+
+        comment_uri = response['location']
```
mitar, line 170: Is this URI directly to the comment or URI to all comments for the post?

piplmesh/api/tests/test_basic.py
```diff
@@ -143,6 +143,66 @@ def test_basic(self):
...
+
+        response = self.client.get(self.resourceListURI('notification'))
+        self.assertEqual(response.status_code, 200)
+        response = json.loads(response.content)
+
+        self.assertEqual(len(response['objects']), 1)
+
+        notification_uri = response['objects'][0]['resource_uri']
+
+        self.assertEqual(response['objects'][0]['read'], False)
```
mitar, line 188: Check here that notification points to comment/post, or whatever the content it has.

piplmesh/api/resources.py
```diff
@@ -54,7 +54,7 @@ class NotificationResource(resources.MongoEngineResource):
 
     class Meta:
         queryset = api_models.Notification.objects.all()
-        allowed_methods = ('get',)
+        allowed_methods = ('get', 'patch',)
```
mitar, line 68: By allowing patch, you are allowing user to change ANY field of notification object. This is probably not what you want.

You should define other than read mark fields as read-only.

piplmesh/api/tests/test_basic.py
zupan, line 170: This URI is directly to the comment. Will change to all post comments.

mitar, line 170: No, is OK. But just then code bellow is wrong, `len(response['objects'])` is probably wrong as it is just one object in response?

zupan, line 170: You are right. I will check comment message instead of number of comments.

mitar:
> Tests are finally breaking at the same error that causing problem in application when trying to mark notification as read.

Fixed. Please update from main repository.

piplmesh/api/tests/test_basic.py
zupan, line 188: Fixed

piplmesh/api/resources.py
zupan, line 68: Fixed

zupan:
Everything is working and up to date.

piplmesh/api/tests/test_basic.py
```diff
@@ -143,6 +143,61 @@ def test_basic(self):
...
+        self.assertEqual(response['objects'][0]['post'], self.fullURItoAbsoluteURI(post_uri))
+
+        # Mark notification as read
+        response = self.client.patch(notification_uri, '{"read": true}', content_type='application/json')
+        # TODO: after solving the problem remove this print
```
mitar: Remove this TODO. :-)

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -233,7 +233,37 @@ function Notification(data) {
...
+                        $('.notification').each(function (i, notification) {
+                            if (!$(notification).data('notification').read) {
+                                unread_notifications_counter++;
+                            }
+                        });
+                        $('#notifications_count').text(unread_notifications_counter);
+                        notification.addClass('read_notification');
+
+                        // TODO: This has to be removed
+                        window.console.log("Notification marked as read.");
```
mitar: Remove this. :-)

mitar: (We can see that it works by CSS class change, and user feedback there.)

piplmesh/api/resources.py
```diff
@@ -52,9 +52,16 @@ class NotificationResource(resources.MongoEngineResource):
     post = tastypie_mongoengine_fields.ReferenceField(to='piplmesh.api.resources.PostResource', attribute='post', null=False, full=False)
     comment = fields.CustomReferenceField(to='piplmesh.api.resources.CommentResource', getter=lambda obj: obj.post.comments[obj.comment], setter=lambda obj: obj.pk, null=False, full=True)
 
+    def __init__(self, *args, **kwargs):
+        super(resources.MongoEngineResource, self).__init__()
+
+        for name, field in self.fields.iteritems():
```
mitar: This is not exactly as I imagined but is a nice idea. It probably seems to work, but there are some slight issues with such approach:
- you are in fact accessing internal attributes of field objects, you are not sure if you are not breaking something with this, it might be that there is some other things which are configured in field based on initial value before you change it
- you are changing the readonly flag only for objects, not for class, so instances of the class will have their fields readonly, but if I try to passively check fields without creating a class instance, I will get another picture

So yes, your approach seems to work but it is error-prone. And this is one thing I would like to teach you. To think about such things. Not all solutions are always the best. For code which should run for long time, you should use only documented APIs (even those change). And use undefined ways only if there is no other way.

Here, there is official way. It is ugly and tedious, but it is: define all resource fields manually. Currently, most of them are inferred automatically (see `get_fields` method in `MongoEngineResource`) from the fields defined the document, only fields relating to some other resource has to be defined manually. But in this case you have to define it manually.

The second approach is to override `get_fields` to post-process fields automatically. This would solve the second issue above, fields would have flag also for classes, but would not solve the first one.

The third approach is to change `get_fields` implementation in `MongoEngineResource` to allow hooking in and alter field creation parameters dynamically. I will prepare this now.

mitar: Pull from main repository and use updated library.

zupan: I chose this approach because I didn't wanted to define all resource fields manually. Will change it.

mitar: Change by overriding `api_field_options` I added for you to `MongoEngineResource`.

piplmesh/api/tests/test_basic.py
```diff
@@ -143,6 +144,64 @@ def test_basic(self):
...
+        self.assertEqual(response.status_code, 202)
+
+        response = self.client.get(notification_uri)
+        self.assertEqual(response.status_code, 200)
+        response = json.loads(response.content)
+
+        self.assertEqual(response['read'], True)
+
+        # Testing readonly parameter
+        response = self.client.patch(notification_uri, '{"created_time": "%s"}' % datetime.now().isoformat(), content_type='application/json')
```
mitar, line 202: isoformat is probably not the best, but probably good enough for tests.

mitar:
I merged it.

PR #310, opened 2012-09-27
Weather,clouds and panorama layer for google maps

Gitek1:
Added checkbox for weather, clouds and panorama layer.

### DIFF
--- a/piplmesh/panels/map/static/piplmesh/panel/map/panel.js
+++ b/piplmesh/panels/map/static/piplmesh/panel/map/panel.js
@@ -29,13 +29,13 @@ $(document).ready(function () {
 
     var map = new google.maps.Map($('#map').get(0), options);
 
-    // TODO: This should be user-configurable
     // Necessary libraries are included with googleapis in piplmesh/panels/map/templates/panel/map/panel.html
     var map_layers = [
         {
             'id': 'weather',
             'label': gettext("Weather"),
             'layer': new google.maps.weather.WeatherLayer({
+                // TODO: This should be user-configurable
                 'temperatureUnits': google.maps.weather.TemperatureUnit.CELSIUS
             })
         },
@@ -84,18 +84,18 @@ $(document).ready(function () {
 
     function addMapLayerOption(map_layer) {
         var checkbox_container = $('<div/>').appendTo('#map-layers');
-        var checkbox = $('<input/>').prop({
+        var checkbox = $('<input/>').attr({
             'id': 'map-layer-' + map_layer.id,
             'type': 'checkbox',
             'name': 'map-layer-' + map_layer.id
         }).change(function (event) {
             setLayerVisibility(checkbox, map_layer);
         }).appendTo(checkbox_container);
-        $('<label/>').prop('for', 'map-layer-' + map_layer.id).text(map_layer.label).appendTo(checkbox_container);
+        $('<label/>').attr('for', 'map-layer-' + map_layer.id).text(map_layer.label).appendTo(checkbox_container);
     }
 
     function setLayerVisibility(checkbox, map_layer) {
-        if ((checkbox).prop('checked')) {
+        if (checkbox.prop('checked')) {
             map_layer.layer.setMap(map);
         }
         else {
@@ -138,7 +138,7 @@ $(document).ready(function () {
         $('#advanced-map-container').show();
         $('<div/>').prop('id', 'map-layers').appendTo('#advanced-map');
         // Map layer options should be added after #map-layers is created
-        $.each(map_layers, function(key, map_layer) {
+        $.each(map_layers, function(index, map_layer) {
             addMapLayerOption(map_layer);
         });
 
--- a/piplmesh/panels/map/static/piplmesh/panel/map/panel.js
+++ b/piplmesh/panels/map/static/piplmesh/panel/map/panel.js
@@ -41,7 +41,6 @@ $(document).ready(function () {
     map_layers_labels.panoramio = gettext('Panoramio');
     map_layers.panoramio = new google.maps.panoramio.PanoramioLayer();
 
-
     map.mapTypes.set('OpenStreetMap', new google.maps.ImageMapType({
         'getTileUrl': function(coordinates, zoom) {
             // TODO: Is there HTTPS version - it would be better to use that so we do not get mixed content errors if we run our portal over HTTPS
@@ -78,7 +77,7 @@ $(document).ready(function () {
             $('<input/>').prop({
                 'id': 'map-layer-' + layer,
                 'type': 'checkbox',
-                'name': 'map-layer--' + layer
+                'name': 'map-layer-' + layer
             }).appendTo('#map-layers');
             $('<label/>').prop('for', 'map-layer-' + layer).text(label).appendTo('#map-layers');
             $('<br/>').appendTo('#map-layers');
--- a/piplmesh/panels/map/static/piplmesh/panel/map/panel.js
+++ b/piplmesh/panels/map/static/piplmesh/panel/map/panel.js
@@ -29,6 +29,19 @@ $(document).ready(function () {
 
     var map = new google.maps.Map($('#map').get(0), options);
 
+    // Necessary libraries are included with googleapis in piplmesh/panels/map/templates/panel/map/panel.html
+    var map_layers = new Object;
+    var map_layers_labels = new Object;
+    map_layers_labels.weather = gettext('Weather');
+    map_layers.weather = new google.maps.weather.WeatherLayer({
+        temperatureUnits: google.maps.weather.TemperatureUnit.CELSIUS
+    });
+    map_layers_labels.clouds = gettext('Clouds');
+    map_layers.clouds = new google.maps.weather.CloudLayer();
+    map_layers_labels.panoramio = gettext('Panoramio');
+    map_layers.panoramio = new google.maps.panoramio.PanoramioLayer();
+
+
     map.mapTypes.set('OpenStreetMap', new google.maps.ImageMapType({
         'getTileUrl': function(coordinates, zoom) {
             // TODO: Is there HTTPS version - it would be better to use that so we do not get mixed content errors if we run our portal over HTTPS
@@ -58,6 +71,30 @@ $(document).ready(function () {
         map.draggable = true;
         map.streetViewControl = true;
         map.mapTypeControlOptions.position = google.maps.ControlPosition.TOP_RIGHT;
+    }
+
+    function addMapLayerOption(layer, label) {
+        if($('#map-layers').lenght != 0) {
+            $('<input/>').prop({
+                'id': 'map-layer-' + layer,
+                'type': 'checkbox',
+                'name': 'map-layer--' + layer
+            }).appendTo('#map-layers');
+            $('<label/>').prop('for', 'map-layer-' + layer).text(label).appendTo('#map-layers');
+            $('<br/>').appendTo('#map-layers');
+            $('#map-layer-' + layer).click(function (event) {
+                setLayerVisibility(layer);
+            });
+        }
+    }
+
+    function setLayerVisibility(layer) {
+        if ($('#map-layer-' + layer).attr('checked')) {
+            map_layers[layer].setMap(map);
+        }
+        else {
+            map_layers[layer].setMap(null);
+        }
     }
 
     function closeAdvancedMap() {
@@ -93,6 +130,12 @@ $(document).ready(function () {
         $('#advanced-map-container').hide();
         $('#map').detach().prependTo('#advanced-map-container');
         $('#advanced-map-container').show();
+        $('<div/>').prop('id', 'map-layers').appendTo('#advanced-map');
+        // Map layer options should be added after #map-layers is created
+        for (var layer in map_layers) {
+            addMapLayerOption(layer, gettext(map_layers_labels[layer]));
+        }
+
         refreshMapCenter();
         $(document).keyup(function (event) {
             if (event.keyCode == 27) {
--- a/piplmesh/panels/map/static/piplmesh/panel/map/panel.js
+++ b/piplmesh/panels/map/static/piplmesh/panel/map/panel.js
@@ -29,6 +29,7 @@ $(document).ready(function () {
 
     var map = new google.maps.Map($('#map').get(0), options);
 
+    // TODO: This should be user-configurable
     // Necessary libraries are included with googleapis in piplmesh/panels/map/templates/panel/map/panel.html
     var map_layers = [
         {
@@ -37,6 +38,11 @@ $(document).ready(function () {
             'layer': new google.maps.weather.WeatherLayer({
                 'temperatureUnits': google.maps.weather.TemperatureUnit.CELSIUS
             })
+        },
+        {
+            'id': 'clouds',
+            'label': gettext("Clouds"),
+            'layer': new google.maps.weather.CloudLayer()
         },
         {
             'id': 'panoramio',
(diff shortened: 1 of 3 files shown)


Gitek1:
Check it.

mitar:
I don't see clouds working.

Why all labels on the map are hidden when you enable weather? I would leave labels on, even if things become more crowded, because layers in general it would be useful to have labels on.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -29,6 +29,18 @@ $(document).ready(function () {
 
     var map = new google.maps.Map($('#map').get(0), options);
 
+    // Necessary libraries are included with googleapis in piplmesh/panels/map/templates/panel/map/panel.html
+    var map_layers = new Object;
```
mitar: All this is not good way to define things. Order should be predictable. So use list. Then define each layer as one entry there. Something like:

```
var map_layers = [
    {
        'label': gettext("Weather"),
        'layer': ew google.maps.weather.WeatherLayer(...)
    },
    {
        ...
    }
];
```

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -29,6 +29,18 @@ $(document).ready(function () {
...
+    var map_layers_labels = new Object;
```
mitar: In general, new empty object can be done with `{}`. Much nicer code.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -29,6 +29,18 @@ $(document).ready(function () {
...
+    map_layers_labels.weather = gettext('Weather');
+    map_layers.weather = new google.maps.weather.WeatherLayer({
+        temperatureUnits: google.maps.weather.TemperatureUnit.CELSIUS
```
mitar: Use `'` around keys.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -29,6 +29,18 @@ $(document).ready(function () {
...
+    });
+    map_layers_labels.clouds = gettext('Clouds');
```
mitar: User-exposed strings should be in `"`.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -93,6 +129,12 @@ $(document).ready(function () {
         $('#advanced-map-container').hide();
         $('#map').detach().prependTo('#advanced-map-container');
         $('#advanced-map-container').show();
+        $('<div/>').prop('id', 'map-layers').appendTo('#advanced-map');
+        // Map layer options should be added after #map-layers is created
+        for (var layer in map_layers) {
```
mitar: Use jQuery `.each`.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +72,30 @@ $(document).ready(function () {
...
 
+    function addMapLayerOption(layer, label) {
+        if($('#map-layers').lenght != 0) {
+            $('<input/>').prop({
+                'id': 'map-layer-' + layer,
+                'type': 'checkbox',
+                'name': 'map-layer-' + layer
+            }).appendTo('#map-layers');
+            $('<label/>').prop('for', 'map-layer-' + layer).text(label).appendTo('#map-layers');
+            $('<br/>').appendTo('#map-layers');
```
mitar: Why is `<br/>` needed? In general this is something which is related to design. So you can do that with CSS. Probably you just want one div around and it will be same effect?

Gitek1:
Fixed.

I see clouds when zoomed out. I removed them in last commit, I guess they are not that usable.

I think labels cant be shown in weather mode, there is no option in documentation.
Documentation says ( https://developers.google.com/maps/documentation/javascript/layers#WeatherLayer ):
"When the weather layer is enabled and the zoom level is set to 12 or lower, administrative labels, such as the names of streets or regions, will be disabled on your map"

mitar:
I would have cloud layer. If not, it looks interesting. :-) It is already optional. But also later on we can make people be able to configure which layers to see.

mitar:
OK, then if this is how weather layer behaves, let it be.

Can you maybe add a tooltip over the layer entry in the list on the right? Then we could have a tooltip like "Hides labels on zoom level 12 or lower."

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -29,6 +29,22 @@ $(document).ready(function () {
 
     var map = new google.maps.Map($('#map').get(0), options);
 
+    // Necessary libraries are included with googleapis in piplmesh/panels/map/templates/panel/map/panel.html
+    var map_layers = [
+        {
+            'id': 'weather',
+            'label': gettext("Weather"),
+            'layer': new google.maps.weather.WeatherLayer({
+                'temperatureUnits': google.maps.weather.TemperatureUnit.CELSIUS
```
mitar, line 40: Is there a way to get this from user's locale? If not, add TODO that this should be user-configurable.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +76,30 @@ $(document).ready(function () {
...
+            }).appendTo('#map-option-' + id);
+            $('<label/>').prop('for', 'map-layer-' + id).text(label).appendTo('#map-option-' + id);
+            $('#map-layer-' + id).click(function (event) {
+                setLayerVisibility(id, layer);
+            });
+        }
+    }
+
+    function setLayerVisibility(id, layer) {
+        if ($('#map-layer-' + id).attr('checked')) {
```
mitar: Use `prop`, not `attr`. [Read here](http://api.jquery.com/prop/).

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -93,6 +133,12 @@ $(document).ready(function () {
         $('#advanced-map-container').hide();
         $('#map').detach().prependTo('#advanced-map-container');
         $('#advanced-map-container').show();
+        $('<div/>').prop('id', 'map-layers').appendTo('#advanced-map');
+        // Map layer options should be added after #map-layers is created
+        $.each(map_layers, function(key, option) {
```
mitar: Rename `option` to `layer`. You iterate over layers, so each element is a layer, no? Or rename `map_layers` to `map_layers_options`.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
mitar, line 40: (You always have to think about such assumptions. World is global today! Products/services should also be!)

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +76,30 @@ $(document).ready(function () {
         map.mapTypeControlOptions.position = google.maps.ControlPosition.TOP_RIGHT;
     }
 
+    function addMapLayerOption(id, label, layer) {
+        if($('#map-layers').lenght != 0) {
```
mitar: Space before `(`.

Have you run this code? Because it is not `lenght` but `length`.

In general it is better:

```
if ($('#map-layers').length) {
```

Or even:

```
if ($('#map-layers')) {
```

mitar: The question is, is it really necessary? In general it is better sometimes to get an error if something is not as expected (like, there is no #map-layers).

But in this case you can also leave to jQuery to simply not do anything, when you want to append to it. And this is it. No need for special check.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +76,30 @@ $(document).ready(function () {
...
+            $('<div/>').prop('id','map-option-' + id).appendTo('#map-layers');
```
mitar: Comma after `,`.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -93,6 +133,12 @@ $(document).ready(function () {
         $('#advanced-map-container').hide();
         $('#map').detach().prependTo('#advanced-map-container');
         $('#advanced-map-container').show();
+        $('<div/>').prop('id', 'map-layers').appendTo('#advanced-map');
+        // Map layer options should be added after #map-layers is created
+        $.each(map_layers, function(key, option) {
+            addMapLayerOption(option.id, option.label, option.layer);
```
mitar: Why not simply pass `option` object?

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
mitar: Do you really need ID on this div? You could just store jQuery div object into a variable and use that as a target of `appendTo`. No need to do the query against ID there.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +76,30 @@ $(document).ready(function () {
...
     }
 
+    function addMapLayerOption(id, label, layer) {
+        if($('#map-layers').lenght != 0) {
+            $('<div/>').prop('id','map-option-' + id).appendTo('#map-layers');
+            $('<input/>').prop({
+                'id': 'map-layer-' + id,
+                'type': 'checkbox',
+                'name': 'map-layer-' + id
+            }).appendTo('#map-option-' + id);
```
mitar: So you could just append to object stored in a variable here.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +76,30 @@ $(document).ready(function () {
...
+            $('<label/>').prop('for', 'map-layer-' + id).text(label).appendTo('#map-option-' + id);
```
mitar: So you could just append to object stored in a variable here, at the end.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +76,30 @@ $(document).ready(function () {
...
+            $('#map-layer-' + id).click(function (event) {
```
mitar: Hm, is click the best one? Or should be maybe instead change? Or both? Maybe both?

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -93,6 +133,12 @@ $(document).ready(function () {
         $('#advanced-map-container').hide();
         $('#map').detach().prependTo('#advanced-map-container');
         $('#advanced-map-container').show();
+        $('<div/>').prop('id', 'map-layers').appendTo('#advanced-map');
```
mitar, line 142: Same here, do you really need to set ID, or you could store div object into variable and pass it to `addMapLayerOption`?

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +76,30 @@ $(document).ready(function () {
...
+                'name': 'map-layer-' + id
+            }).appendTo('#map-option-' + id);
+            $('<label/>').prop('for', 'map-layer-' + id).text(label).appendTo('#map-option-' + id);
+            $('#map-layer-' + id).click(function (event) {
+                setLayerVisibility(id, layer);
+            });
+        }
+    }
+
+    function setLayerVisibility(id, layer) {
```
mitar: Similar here, you could pass the object?

So, the problem with all this relaying on IDs is that sometimes you could set invalid IDs, or that there are duplicates. Much easier is to do the query once, or construct the object, and then pass it around. Also from performance stand point (but here we do not really have that issue).

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
mitar: And same here. You should chain jQuery calls together. You create input above, just attach also event handler there.

Gitek1:
Fixed.

I don't think weather option needs a tooltip, clouds may be more problematic.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
Gitek1: Yes, I ran the code and it worked, not even a error in chrome console (?), that's why I didn't notice the typo, so now I just removed it

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
Gitek1: I think both would cancel each other out. I used change in the last commit.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
Gitek1, line 142: I kept this one (#map-layers) because I used it in css to style the div.

mitar:
> I don't think weather option needs a tooltip, clouds may be more problematic.

A tooltip explaining why labels disappeared? I think this is useful because if people enable more layers at the same time, to understand which one hide labels.

What is the issue with clouds?

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
mitar: There was no error because the value was always undefined, which is always false. So my question is, does the logic of your program worked. If it works with this being always false then the question which appears is: do we really need this whole block of code inside the if. :-)

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
mitar: So I remember from past that there are some issues with checkboxes. So because you can change them also through keyboard (navigate to it with tab and press space). So you should make it work also with that.

Anyway, make it work with mouse and keyboard.

Hm, or it was maybe input box and keypress and change. :-)

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
mitar, line 142: You can keep it defined, but don't use it in the code, if you can store it into a variable.

mitar:
One more thing I found out. It is better to use `.attr` when defining new DOM in the code, and `.prop` when you want to set or access current values after DOM was in use.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -29,6 +29,28 @@ $(document).ready(function () {
 
     var map = new google.maps.Map($('#map').get(0), options);
 
+    // TODO: This should be user-configurable
```
mitar: ??? Why TODO here? It should be down there with celsius line.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +82,27 @@ $(document).ready(function () {
         map.mapTypeControlOptions.position = google.maps.ControlPosition.TOP_RIGHT;
     }
 
+    function addMapLayerOption(map_layer) {
+        var checkbox_container = $('<div/>').appendTo('#map-layers');
+        var checkbox = $('<input/>').prop({
```
mitar: So `attr` here.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +82,27 @@ $(document).ready(function () {
...
+            'id': 'map-layer-' + map_layer.id,
+            'type': 'checkbox',
+            'name': 'map-layer-' + map_layer.id
+        }).change(function (event) {
+            setLayerVisibility(checkbox, map_layer);
+        }).appendTo(checkbox_container);
+        $('<label/>').prop('for', 'map-layer-' + map_layer.id).text(map_layer.label).appendTo(checkbox_container);
```
mitar: And here.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -60,6 +82,27 @@ $(document).ready(function () {
...
+    }
+
+    function setLayerVisibility(checkbox, map_layer) {
+        if ((checkbox).prop('checked')) {
```
mitar: But here `prop` is good.

Why `()` around `checkbox`?

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
mitar, line 142: (Of course is variable is easily passed around. You do not have to complicate too much here. I just wanted to point out that this can be done instead of using IDs everywhere automatically.)

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
```diff
@@ -93,6 +136,12 @@ $(document).ready(function () {
         $('#advanced-map-container').hide();
         $('#map').detach().prependTo('#advanced-map-container');
         $('#advanced-map-container').show();
+        $('<div/>').prop('id', 'map-layers').appendTo('#advanced-map');
+        // Map layer options should be added after #map-layers is created
+        $.each(map_layers, function(key, map_layer) {
```
mitar: `key`? What is `key`? Rename it to mean something.

Gitek1: This function takes index key and value from map_layers. I renamed key to index in last commit.

piplmesh/panels/map/static/piplmesh/panel/map/panel.js
Gitek1: I understood that "map_layers" should be configurable, I fixed this in last commit.

Gitek1:
Fixed.

The thing with clouds is that they don't show, even if enabled, when you are zoomed in. When user clicks the weather something happens no matter what zoom level the user is on.

mitar:
So it is not possible to configure tooltip ("title" DOM attribute) for clouds and weather. Something like:
- _Visible on zoom level X and above._
- _Hides map labels on zoom level X and above._

mitar:
Otherwise good.

Gitek1:
Added tooltips

Issue #308, opened 2012-09-27
When user post something on channel show it


mitjap:
Current implementation notifies user about incoming posts but does not display it. Make necessary changes so posts from user will always show immediately after post.

mitar:
Related: #163

mitjap:
In case user has slow connection i guess post should be displayed before we get response from server. This could be tricky to implement thought.

mitar:
No. There is nothing wrong if there is a delay. Also Facebook has it. So we just have to provide some feedback (processing/loading). Because sometimes we will want also to do some spam checking and stuff.

The other way is also that we configure Tastypie to return full object not successful post, not just URI.

mitar:
Anyway, this is probably just an if-sentence, if author == current user, display, otherwise go into current logic.

PR #311, opened 2012-09-27
issue #308 - show posts by user automatically

mitjap:
Automatically shows post to the author. #308

### DIFF
--- a/piplmesh/frontend/static/piplmesh/js/common.js
+++ b/piplmesh/frontend/static/piplmesh/js/common.js
@@ -1,3 +1,7 @@
+function getLocation(href) {
+    return $('<a/>').prop('href', href).get(0)
+}
+
 $(document).ready(function () {
     $.ajaxSetup({
         'timeout': 5000,
--- a/piplmesh/frontend/static/piplmesh/js/common.js
+++ b/piplmesh/frontend/static/piplmesh/js/common.js
@@ -1,5 +1,5 @@
 function getLocation(href) {
-    return $('<a/>').prop('href', href).get(0)
+    return $('<a/>').prop('href', href).get(0);
 }
 
 $(document).ready(function () {
--- a/piplmesh/frontend/static/piplmesh/js/home.js
+++ b/piplmesh/frontend/static/piplmesh/js/home.js
@@ -306,17 +306,11 @@ function showHiddenPosts() {
     $('ul > li.notShown').show('fast').removeClass('notShown');
 }
 
-function getLocation(href) {
-    var l = document.createElement("a");
-    l.href = href;
-    return l;
-}
-
 $(document).ready(function () {
     initializePanels();
 
     // List of URIs of posts by user
-    $($('.posts')).data('user_posts_URIs', []);
+    $('.posts').data('user_posts_URIs', []);
 
     $.updates.registerProcessor('home_channel', 'post_new', function (data) {
         new Post(data.post).addToTop();
@@ -359,8 +353,8 @@ $(document).ready(function () {
             'contentType': 'application/json',
             'dataType': 'json',
             'success': function (data, textStatus, jqXHR) {
-                var full_location_uri = getLocation(jqXHR.getResponseHeader('location')).href;
-                $('.posts').data('user_posts_URIs').push(full_location_uri);
+                var full_post_uri = getLocation(jqXHR.getResponseHeader('location')).href;
+                $('.posts').data('user_posts_URIs').push(full_post_uri);
                 $('#post_text').val(input_box_text).css('min-height', 25);
             },
             'error': function (jqXHR, textStatus, errorThrown) {
--- a/piplmesh/frontend/static/piplmesh/js/home.js
+++ b/piplmesh/frontend/static/piplmesh/js/home.js
@@ -168,13 +168,13 @@ function Post(data) {
         });
     }
 
-    function postByUser () {
-        return $.user_posts_URIs.some(function(user_post_URI) {
-            return user_post_URI.lastIndexOf(self.resource_uri) > -1;
-        });
+    function postByUser() {
+        var user_posts_URIs = $('.posts').data('user_posts_URIs');
+        var full_resource_uri = getLocation(self.resource_uri).href;
+        return $.inArray(full_resource_uri, user_posts_URIs) != -1;
     }
 
-    function showPost (post) {
+    function showPost(post) {
         // TODO: Animation has to be considered and maybe improved
         post.show('fast');
     };
@@ -306,11 +306,17 @@ function showHiddenPosts() {
     $('ul > li.notShown').show('fast').removeClass('notShown');
 }
 
+function getLocation(href) {
+    var l = document.createElement("a");
+    l.href = href;
+    return l;
+}
+
 $(document).ready(function () {
     initializePanels();
 
-    // list of URIs of posts by user
-    $.user_posts_URIs = new Array();
+    // List of URIs of posts by user
+    $($('.posts')).data('user_posts_URIs', []);
 
     $.updates.registerProcessor('home_channel', 'post_new', function (data) {
         new Post(data.post).addToTop();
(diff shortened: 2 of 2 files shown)


mitar:
You should configure your git to have same configuration s you are using on GitHub. Then it can link (and I can link) commits to you easier.

mitar:
(So same author data info, like e-mail.)

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -292,6 +309,9 @@ function showHiddenPosts() {
 $(document).ready(function () {
     initializePanels();
 
+    // list of URIs of posts by user
```
mitar: Always start a comment with upper case letter.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -292,6 +309,9 @@ function showHiddenPosts() {
...
+    $.user_posts_URIs = new Array();
```
mitar: I would use jQuery `.data()` to attach this to `.posts` DOM element?

You can create array easier with simply `[]`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -168,6 +168,17 @@ function Post(data) {
         });
     }
 
+    function postByUser () {
+        return $.user_posts_URIs.some(function(user_post_URI) {
+            return user_post_URI.lastIndexOf(self.resource_uri) > -1;
+        });
+    }
+
+    function showPost (post) {
```
mitar: For non-anonymous functions, style is that there is no space before `(`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -168,6 +168,17 @@ function Post(data) {
         });
     }
 
+    function postByUser () {
+        return $.user_posts_URIs.some(function(user_post_URI) {
+            return user_post_URI.lastIndexOf(self.resource_uri) > -1;
```
mitar: I don't like this comparison? It cannot be equality?

I don't like that you can match any substring? Can you limit this somehow, so that it matches really just the portion that should be equal? Like trailing part or something?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -168,6 +168,17 @@ function Post(data) {
         });
     }
 
+    function postByUser () {
+        return $.user_posts_URIs.some(function(user_post_URI) {
```
mitar: Use [inArray](http://api.jquery.com/jQuery.inArray/). We are using jQuery to make code more portable, so use it.

piplmesh/frontend/static/piplmesh/js/home.js
mitjap: When posting we get location from header with full URI: http://127.0.0.1:8000/api/v1/......, but when getting this value over ajax field resource_uri has only the last part from /api/v1/.... i'll make a regular expression for that

piplmesh/frontend/static/piplmesh/js/home.js
mitjap: .some() is available since javascript 1.6 https://developer.mozilla.org/en-US/docs/JavaScript/New_in_JavaScript/1.6. Can't find when it was released and in what implementations it is present.

.some is good becouse when it finds first that matches that predicate it stops, where inArray can only search by value and not function/predicate. jQuery.grep(array, function(element, index)) could replace .some() but with lesser performance. Not that there will be more than a few elements in array :) So do i use grep() instead and count how many results were returned?

``` javascript
return grep(array, function) > 0;
```

mitjap:
will fix that. sorry for extra work

piplmesh/frontend/static/piplmesh/js/home.js
mitar: You could maybe use something like [this](http://stackoverflow.com/questions/736513/how-do-i-parse-a-url-into-hostname-and-path-in-javascript) to get full URI from `resource_uri` and then compare that?

mitjap: i think i found better library under MIT license 
http://code.google.com/p/jsuri/ 
it looks more capable and stable

mitar: I was thinking of that but I think we do not need to convert full URI to pathname and check that, but to convert pathname to full URI in the same way browser would do it and then check that. So you can create `<a>` and then query what is the full URI as browser sees it? I think this would be the best way to approach the problem.

I could imagine that someday we will run on multiple servers, so it would be important what is full URI to the post. But this would be probably ugly. Hmm.

Yes, the other way is to change both URLs to pathnames and compare that only.

But I am not sure if we need whole library for this? Seems DOM already supports that.

piplmesh/frontend/static/piplmesh/js/home.js
mitar: It seems most browsers support only [JavaScript 1.5](https://en.wikipedia.org/wiki/JavaScript#Versions).

mitar: You are right that `inArray` is not suitable for here. But grep is also ugly. I think that I have been using it in my code, but I don't like it.

I would extend jQuery object with `.is` function. Similar to how they have [is](http://api.jquery.com/is/) for elements, I would define utility function (compare [this](http://api.jquery.com/each/) and [this](http://api.jquery.com/jQuery.each/)).

You can define this in `common.js` and then use it.

mitjap: I don't think this is still necessary. We will parse URI before searching, then we just execute 'inArray'

mitar: Good idea. :-)

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -289,9 +306,18 @@ function showHiddenPosts() {
     $('ul > li.notShown').show('fast').removeClass('notShown');
 }
 
+function getLocation(href) {
```
mitar: Move this to `common.js`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -289,9 +306,18 @@ function showHiddenPosts() {
...
+    var l = document.createElement("a");
```
mitar: I would use jQuery, so, `return $('<a/>').prop('href', href).get(0)`.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -289,9 +306,18 @@ function showHiddenPosts() {
...
+    l.href = href;
+    return l;
+}
+
 $(document).ready(function () {
     initializePanels();
 
+    // List of URIs of posts by user
+    $($('.posts')).data('user_posts_URIs', []);
```
mitar: `$($(`???

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -333,6 +359,8 @@ $(document).ready(function () {
             'contentType': 'application/json',
             'dataType': 'json',
             'success': function (data, textStatus, jqXHR) {
+                var full_location_uri = getLocation(jqXHR.getResponseHeader('location')).href;
```
mitar: `full_location_uri` -> `full_post_uri`.

mitar:
Otherwise good.

mitjap:
Fixed.

piplmesh/frontend/static/piplmesh/js/common.js
```diff
@@ -1,3 +1,7 @@
+function getLocation(href) {
+    return $('<a/>').prop('href', href).get(0)
```
mitar: `;` at the end.

PR #312, opened 2012-09-27
issue #293 - provide user with information about ongoing location mocking

mitjap:
#293


piplmesh/frontend/views.py
```diff
@@ -72,9 +72,10 @@ class LocationsView(generic_views.FormView):
     success_url = urlresolvers.reverse_lazy('home')
 
     def form_valid(self, form):
-        location = form.cleaned_data['locations']
-        if location == '':
+        location = form.cleaned_data['location']
+        if location == 'None':
```
mitar: Why none? I liked empty string more!

piplmesh/frontend/views.py
```diff
@@ -72,9 +72,10 @@ class LocationsView(generic_views.FormView):
...
             nodes.flush_session(self.request)
+            self.request.session[nodes.MOCKING_SESSION_KEY] = False
```
mitar: This should not be needed? `flush_session` should simply remove this key? And this should be enough everywhere that it is false. We should not require that there is explicit disable for mocking in session. If there is nothing in session, it should be disabled.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -324,6 +324,11 @@ input.gsc-search-button {
     top: 2px;
 }
 
+label[for=id_location]
+{
```
mitar: We put `{` on line up.

piplmesh/frontend/static/piplmesh/css/style.css
```diff
@@ -324,6 +324,11 @@ input.gsc-search-button {
     top: 2px;
 }
 
+label[for=id_location]
```
mitar: Uhh, I am afraid of such CSS selectors. I would much more prefer that you wrap the whole form into some id-div (or just add id to li). And then simply hide all labels inside that id.

piplmesh/frontend/forms.py
```diff
@@ -7,16 +7,21 @@
...
-    
+    """
+
     subject = forms.CharField(label=_("Subject"))
     email = forms.EmailField(label=_("Your e-mail address"))
     message = forms.CharField(widget=widgets.Textarea())
 
 def location_choices():
-    yield ('', _("Don't mock location"))
+    yield ('None', _("Don't mock location"))
```
mitar: Empty string!

piplmesh/frontend/forms.py
```diff
@@ -7,16 +7,21 @@
...
     for backend, node in nodes.get_all_nodes_with_backends():
-        yield ('%s-%s' % (backend.get_full_name(), node.id), node.name)
+        yield (nodes.get_full_node_id(backend.get_full_name(), node.id), node.name)
```
mitar: Wouldn't it be better if you could just pass in backend and node, and `get_full_node_id` would extract full name and id?

piplmesh/frontend/context_processors.py
```diff
@@ -23,5 +23,5 @@ def global_vars(request):
         'request_get_next': request.REQUEST.get(auth.REDIRECT_FIELD_NAME),
 
         # Forms
-        'locations_form' : forms.LocationsForm() if getattr(request, 'user', None) and request.user.is_authenticated() and request.user.is_staff else None,
+        'locations_form' : forms.LocationsForm(initial={'location': forms.initial_location(request)}) if getattr(request, 'user', None) and request.user.is_authenticated() and request.user.is_staff else None,
```
mitar: OK, this is getting ugly. :-)

I would suggest, that you change code of this whole processor so that there is something like:

```
context = {
   ... static values always defined
}

if request can be mocked:
   location_form = ...

   context.update({
       'location_form': location_form,
   })

return context
```

piplmesh/frontend/forms.py
```diff
@@ -7,16 +7,21 @@
...
+
+def initial_location(request):
+    node_id = request.session.get(nodes.SESSION_KEY, None)
+    backend_name = request.session.get(nodes.BACKEND_SESSION_KEY, None)
```
mitar: No, you should access `request.node`.

piplmesh/nodes/__init__.py
```diff
@@ -12,6 +12,12 @@
 CLOSEST_LONGITUDE_SESSION_KEY = '_nodes_longitude'
 MOCKING_SESSION_KEY = '_nodes_mocking'
 
+def get_full_node_id(backend_name, node_id):
```
mitar: Hm, we might move this to the method of node class?

I believe we store backend information in the node class (we inject from outside, when we crate it). If not, we should. :-)

So node class should have information from which backend it is. And then this could be method on the class.

piplmesh/frontend/views.py
mitjap: When user selected value with empty id that somehow didn't validate correctly and TemplateResponseMixin raised error for missing template.

piplmesh/frontend/forms.py
mitar: And as just described, node should know its backend. So you should not retry getting this information from session.

We might even allow node to know if it being mocked?

piplmesh/frontend/forms.py
```diff
@@ -7,16 +7,21 @@
...
+    return nodes.get_full_node_id(backend_name, node_id) if nodes.is_mocking(request) else 'None'
```
mitar: You can make this more readable with using normal if-sentence. Return `''`.

piplmesh/frontend/views.py
mitar: Aha, this is the test of what it happens if it is invalid form. :-) So it tries to find template to display error with. :-)

Hm, maybe we should not require the field. Then it will work? Because in some way, we could say that not providing the field is disabling it. And empty string is probably the same as not providing it.

mitar: Hm, I even think that Django adds empty element automatically if you do not require choice-configured field?

piplmesh/frontend/static/piplmesh/css/style.css
mitjap: Why. I think it is very pretty solution. Are you familiar with any known issues?
It looks like it is implemented in every major version - http://kimblim.dk/css-tests/selectors/

piplmesh/frontend/views.py
mitjap: You're right. It was done correctly and I "broke" it :)

piplmesh/frontend/views.py
mitjap: I'm sure it would work perfectly if you provided a template. Not sure what's the difference between '' and 'some_text'.  I think there would be much debugging needed to find out what happens in background.

mitar: So the correct way would be to change the view so that if form is invalid, user is redirected back to initiator page (currently hard-coded to home) and through Django messages (error message) she is notified of the form error.

mitar: Because in our case there is no reason to provide a template. Form is not linked to any particular page.

mitar: I cannot agree here with you that `''` is worse for debugging than some text? I think setting it to not be required and using `''` makes things simpler.

piplmesh/frontend/static/piplmesh/css/style.css
mitar: I know I am old-fashioned here. I even don't like `>` selector.

But this is from my experience. That once you name the div with `id`, because there was one need. Then designer finds also other good uses of it. And then can create selectors which are possible because there is that `id`. So it is more the idea, that if there was once such a need, it could happen also in the future, and it will open also new ideas how to design things. You must also know that most designers have problems creating complex selectors. So providing them with such tools makes things easier for them.

piplmesh/frontend/forms.py
mitjap: That was my first implementation but when you were building id from session all you had were strings. If this was java I would use same name but different signature.

piplmesh/frontend/forms.py
mitjap: With current implementation, nodes does not know what is its backend. I think this should be different ticket. Agree?

piplmesh/frontend/views.py
mitjap: I'm not saying '' is worst than some text. I was saying that its hard to debug things with no debugger except print :D

piplmesh/frontend/forms.py
mitar: You can test for type of input arguments. :-)

If it is already sting, you use that, otherwise you call `get_full_name()` on it. :-)

Anyway, you will also not read from session, but will use `request.node`. :-) And then do `request.node.get_full_node_id()` even. :-)

piplmesh/frontend/views.py
mitar: Use FireBug and use "Net" panel there. You will see output Django throws at you. Safari/WebKit even renders that output so you get nice Django debugging page on error. :-)

piplmesh/frontend/forms.py
mitar: No. This is not independent feature or feature which would build upon this one, but is requirement for good implementation of this feature. You can create different ticket, but then you will have to do it first, and then come back here. :-)

piplmesh/frontend/static/piplmesh/css/style.css
mitjap: you mean `>` selector in jquery? :)

mitar: No, `>` in CSS. jQuery abstracts differences so there should not be any issues. But it works in the same way.

piplmesh/frontend/views.py
mitjap:
```
Aha, this is the test of what it happens if it is invalid form. :-) So it tries to find template to display error with. :-)

Hm, maybe we should not require the field. Then it will work? Because in some way, we could say that not providing the field is disabling it. And empty string is probably the same as not providing it.
```

Sorry didn't read you post careful enough.  You already provided a solution here. Yes, not requiring this field fixes this issue.  `id:''` means an absence of this field.

Edit: in fact I overlooked that comment.

mitjap:
Can you please check now.

piplmesh/frontend/context_processors.py
```diff
@@ -23,5 +26,7 @@ def global_vars(request):
         'request_get_next': request.REQUEST.get(auth.REDIRECT_FIELD_NAME),
 
         # Forms
-        'locations_form' : forms.LocationsForm() if getattr(request, 'user', None) and request.user.is_authenticated() and request.user.is_staff else None,
+        'locations_form' : locations_form,
```
mitar: I rather see my approach of adding to the context if the value should be there. So not setting it to None, but not even being in the dictionary if the value should not be there.

mitjap: Ok, no problem

piplmesh/frontend/context_processors.py
```diff
@@ -9,6 +9,9 @@ def global_vars(request):
     """
     Adds global context variables to the context.
     """
+    locations_form = None
+    if getattr(request, 'user', None) and request.user.is_authenticated() and request.user.is_staff:
+        locations_form = forms.LocationsForm(initial={'location': forms.initial_location(request)})
```
mitar: I would everywhere remove S from locations in LocationsForm and others.

piplmesh/frontend/context_processors.py
```diff
@@ -23,5 +26,7 @@ def global_vars(request):
         'request_get_next': request.REQUEST.get(auth.REDIRECT_FIELD_NAME),
 
         # Forms
-        'locations_form' : forms.LocationsForm() if getattr(request, 'user', None) and request.user.is_authenticated() and request.user.is_staff else None,
+        'locations_form' : locations_form,
     }
+
+
```
mitar, line 30: Unnecessary lines.

piplmesh/frontend/forms.py
```diff
@@ -4,19 +4,24 @@
 
 from piplmesh import nodes
 
+NO_MOCKING_ID = ''
```
mitar, line 7: Good!

piplmesh/frontend/forms.py
```diff
@@ -4,19 +4,24 @@
...
 
 def location_choices():
-    yield ('', _("Don't mock location"))
+    yield (NO_MOCKING_ID, _("Don't mock location"))
     for backend, node in nodes.get_all_nodes_with_backends():
-        yield ('%s-%s' % (backend.get_full_name(), node.id), node.name)
+        yield (nodes.get_full_node_id(backend.get_full_name(), node.id), node.name)
+
+def initial_location(request):
+    return request.node.get_full_node_id() if nodes.is_mocking(request) else NO_MOCKING_ID
```
mitar: Please make normal if sentence here.

piplmesh/frontend/views.py
```diff
@@ -72,11 +73,12 @@ class LocationsView(generic_views.FormView):
     success_url = urlresolvers.reverse_lazy('home')
 
     def form_valid(self, form):
-        location = form.cleaned_data['locations']
-        if location == '':
+        location = form.cleaned_data['location']
+        print 'Debug: %s' % location
```
mitar: Remove.

piplmesh/frontend/views.py
```diff
@@ -19,6 +19,7 @@
 from pushserver.utils import updates
 
 from piplmesh import nodes
+from piplmesh.nodes import models as node_models
```
mitar: `nodes_models`.

piplmesh/frontend/views.py
mitjap: Just saw it :)

piplmesh/frontend/views.py
```diff
@@ -72,11 +73,12 @@ class LocationsView(generic_views.FormView):
...
     def form_valid(self, form):
-        location = form.cleaned_data['locations']
-        if location == '':
+        location = form.cleaned_data['location']
+        print 'Debug: %s' % location
+        if location == forms.NO_MOCKING_ID:
             nodes.flush_session(self.request)
         else:
-            node_backend, node_id = location.split('-', 1)
+            node_backend, node_id = node_models.parse_full_node_id(location)
```
mitar: It is strange that `get_full_node_id` is in nodes, but `parse_full_node_id` is models.

piplmesh/nodes/__init__.py
```diff
@@ -3,6 +3,8 @@
 from django.core import exceptions
 from django.utils import importlib
 
+NODE_ID_SEPARATOR = '-'
```
mitar: You are not really using it bellow in `get_full_node_id`.

piplmesh/nodes/backends.py
```diff
@@ -4,7 +4,11 @@
 
 from . import data
 
-class RandomNodesBackend(object):
+class NodeBackend(object):
+    def get_full_name(self):
+        return "%s.%s" % (self.__module__, self.__class__.__name__)
```
mitar: Static string, `'`.

piplmesh/nodes/models.py
```diff
@@ -1,4 +1,7 @@
+from piplmesh import nodes
+
 class Node(object):
+
```
mitar, line 4: Unnecessary empty line.

piplmesh/nodes/__init__.py
mitar: Or maybe you should remove this?

piplmesh/nodes/models.py
```diff
@@ -13,4 +16,10 @@ def is_outside_request(self):
...
 
     def is_inside_request(self):
-        return not self._outside_request
\ No newline at end of file
+        return not self._outside_request
+
+    def get_full_node_id(self):
+        return '%s%s%s' % (self.backend.get_full_name(), nodes.NODE_ID_SEPARATOR, self.id)
+
+def parse_full_node_id(full_node_id):
```
mitar: You could make this class-method of Node object.

piplmesh/nodes/__init__.py
mitjap: What do you mean by that?

Edit: Nevermind, I know what you meant.

mitar: I would move this to models?

piplmesh/nodes/__init__.py
mitjap: This should already be gone :)

piplmesh/frontend/forms.py
```diff
@@ -4,19 +4,24 @@
...
     subject = forms.CharField(label=_("Subject"))
     email = forms.EmailField(label=_("Your e-mail address"))
     message = forms.CharField(widget=widgets.Textarea())
 
 def location_choices():
-    yield ('', _("Don't mock location"))
+    yield (NO_MOCKING_ID, _("Don't mock location"))
     for backend, node in nodes.get_all_nodes_with_backends():
-        yield ('%s-%s' % (backend.get_full_name(), node.id), node.name)
+        yield (nodes.get_full_node_id(backend.get_full_name(), node.id), node.name)
```
mitar: Why not using `node.get_full_node_id()` here?

piplmesh/frontend/views.py
mitjap: Its a 'static' method. Parsing node id must be possible even without an instance of node.

piplmesh/frontend/context_processors.py
```diff
@@ -21,7 +20,10 @@ def global_vars(request):
...
         'logo_url': "piplmesh/images/logo-%s.png" % translation.get_language(),
         'request_get_next': request.REQUEST.get(auth.REDIRECT_FIELD_NAME),
-
-        # Forms
-        'locations_form' : forms.LocationsForm() if getattr(request, 'user', None) and request.user.is_authenticated() and request.user.is_staff else None,
     }
+
+    # Location
+    if getattr(request, 'user', None) and request.user.is_authenticated() and request.user.is_staff:
+        context.update({'location_form': forms.LocationForm(initial={'location': forms.initial_location(request)})})
```
mitar: Go into new line after first `{`. See how we have elsewhere in the code structured `context.update`.

piplmesh/nodes/__init__.py
```diff
@@ -3,6 +3,7 @@
 from django.core import exceptions
 from django.utils import importlib
 
+
```
mitar: Unnecessary empty line.

mitar:
Looks good. The only thing missing is what happens if there is form error. :-) So that template strange exception stuff ...

PR #314, opened 2012-09-28
Fixed gender when registering with services.

mfrlin:
Where that was possible.


mitar:
No, this is not a good solution.

piplmesh/account/backends.py
```diff
@@ -75,9 +75,8 @@ def authenticate(self, facebook_access_token, request):
             # TODO: Do we know if all e-mail addresses given by Facebook are verified?
             # TODO: Does not Facebook support multiple e-mail addresses? Which one is given here?
             user.email = facebook_profile_data.get('email') or None
-        if user.gender is None:
-            # TODO: Does it really map so cleanly?
-            user.gender = facebook_profile_data.get('gender') or None
+        
+        user.gender = facebook_profile_data.get('gender') # Facebook always provide gender so its OK to do that
```
mitar, line 79: You cannot override data from Facebook, if user has already configured data. That is the idea why we are checking "is None" in original code.

So if gender is already set, then we assume that user wants that configuration. Even if on Facebook has maybe different value. Yes, concretely for gender this is a bit less common (for name and other things could be more). But still.

mitar, line 79: Furthermore, you removed TODO about checking how does Facebook values for gender map to our values. Somebody hast to check that.

mitar:
I think that the only problem with gender field is that it is required. It should not be required and default value should be None. Then everything would work.

mitar:
(Hopefully, but let's try it.)

piplmesh/account/backends.py
mfrlin, line 79: I did check and they use male and female.

But when user registers with facebook, gender is already set so all users get the default value of female and they have no way of reverting it.

For connecting exsiting users with facebook acount we shouldn't use registration I think but just get the facebook link and token and things like that, that enable user to login with facebook.

mfrlin:
ok, so make gender field not required?

mitar:
Yes. You can do that in the definition of gender field.

mitar:
Hm, but it seems field is already non required. So maybe just default should be set to None.

piplmesh/account/backends.py
mitar, line 79: > I did check and they use male and female.

So same as we do?

> But when user registers with facebook, gender is already set so all users get the default value of female and they have no way of reverting it.

Yes, and this is a mistake which should we fix. So in general it should not be set if user has not set it explicitly.

> For connecting exsiting users with facebook acount we shouldn't use registration I think but just get the facebook link and token and things like that, that enable user to login with facebook.

We have just a concept of linking, if user has not registered with us but just used Facebook, his lazy user gets linked with Facebook. If he registered with us, his lazy user is converted to real user. If he links then with Facebook, his real user is linked with Facebook.

mfrlin:
Ok, let's close this down since it went in wrong direction. It's easier to start again.

mitar:
OK.

Issue #315, opened 2012-09-28
Redirect user on validation error in LocationForm


mitjap:
LocationForm is used to mock location. But it has no template backing it and cannot really have a template because it is displayed on all pages, not on one particular view. So on validation error redirect user where he came from and show him error message through Django messages instead.

How to make validation fail:
- set location field to required,
- mock location
- disable mocking by selecting first element in list

PR #317, opened 2012-09-28
issue #315 - redirect user on error in LocationForm

mitjap:
#315

### DIFF
--- a/piplmesh/frontend/views.py
+++ b/piplmesh/frontend/views.py
@@ -5,6 +5,7 @@ from django.conf import settings
 from django.contrib import messages
 from django.core import exceptions, mail, urlresolvers
 from django.core.files import storage
+from django.http import HttpResponseRedirect
 from django.test import client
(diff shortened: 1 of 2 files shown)


piplmesh/frontend/views.py
```diff
@@ -5,6 +5,7 @@
 from django.contrib import messages
 from django.core import exceptions, mail, urlresolvers
 from django.core.files import storage
+from django.http import HttpResponseRedirect
```

piplmesh/frontend/views.py
```diff
@@ -85,6 +86,9 @@ def form_valid(self, form):
 
         return super(LocationView, self).form_valid(form)
 
+    def form_invalid(self, form):
+        return HttpResponseRedirect(self.get_success_url())
```
mitar: You should also display error message! Read [here](https://docs.djangoproject.com/en/dev/ref/contrib/messages/).

PR #319, opened 2012-09-28
Fixed gender issue

mfrlin:


piplmesh/account/backends.py
```diff
@@ -203,7 +205,7 @@ def authenticate(self, google_access_token, request):
             user.email = google_profile_data.get('email') or None
             if google_profile_data.get('verified_email'):
                 user.email_confirmed = True
-        if user.gender is None:
+        if user.gender is None and google_profile_data.get('gender') is not "other":
```
mitar: Constant string, so use `'` around "other".

Use `!=` instead of `is not 'other'`.

piplmesh/account/backends.py
```diff
@@ -42,6 +42,8 @@ def user_class(self):
 class FacebookBackend(MongoEngineBackend):
     """
     Facebook authentication.
+    
+    Facebook uses strings 'male' and 'female' for respresenting user gender.
```
mitar, line 46: Then you can remove TODO bellow, no? (Only for Facebook.)

piplmesh/account/backends.py
```diff
@@ -203,7 +204,7 @@ def authenticate(self, google_access_token, request):
             user.email = google_profile_data.get('email') or None
             if google_profile_data.get('verified_email'):
                 user.email_confirmed = True
-        if user.gender is None:
+        if user.gender is None and google_profile_data.get('gender') != "other":
```
mitar: `'` instead of `"`.

PR #327, opened 2012-10-07
Panels - fix for bug caused by jQuery update

NejcZupec:
I've updated jQuery UI and added one small code fix.  Panels can be again dragged and dropped.


mitar:
Great!

PR #330, opened 2013-07-07
push hugs and runs to all clients

igypigi:


piplmesh/frontend/views.py
```diff
@@ -164,6 +164,43 @@ def send_update_on_comment(sender, comment, post, request, bundle, **kwargs):
...
+def send_update_on_hug_run(sender, post, request, **kwargs):
+    hugs = []
+    for hug in api_models.Post.objects.get(pk=post.id).hugs:
+        s = {
+            'id': hug.id,
+            'author': {
+                'username': hug.author,
+                },
+            'created_time': hug.created_time,
+            'resource_uri': '/api/v1/post/' + str(post.id) + '/hugs/' + str(hug.id) + '/',
```
mitar, line 177: No way you are constructing URLs on the server side manually! While I might look over that in client code, on the server side you should use [reverse](https://docs.djangoproject.com/en/1.4/topics/http/urls/#reverse).

piplmesh/frontend/views.py
```diff
@@ -164,6 +164,43 @@ def send_update_on_comment(sender, comment, post, request, bundle, **kwargs):
...
 
+@dispatch.receiver(signals.hug_run_updated)
+def send_update_on_hug_run(sender, post, request, **kwargs):
+    hugs = []
+    for hug in api_models.Post.objects.get(pk=post.id).hugs:
+        s = {
+            'id': hug.id,
+            'author': {
+                'username': hug.author,
+                },
```
mitar, line 175: Style. Unindent.

piplmesh/frontend/views.py
```diff
@@ -164,6 +164,43 @@ def send_update_on_comment(sender, comment, post, request, bundle, **kwargs):
...
+        }
+        hugs.append(s)
+
+    runs = []
+    for run in api_models.Post.objects.get(pk=post.id).runs:
+        s = {
+            'id': run.id,
+            'author': {
+                'username': run.author,
+                },
```
mitar, line 187: Style.

piplmesh/api/signals.py
```diff
@@ -6,3 +6,5 @@
 
 # Signals dispatched when resources are updated
 post_updated = dispatch.Signal(providing_args=('post', 'request', 'bundle'))
+
+hug_run_updated = dispatch.Signal(providing_args=('post', 'request'))
```
mitar, line 10: Provide and pass all useful things. Like hug/run object itself, post, request, and bundle. See how comments signal is done.

mitar, line 10: (Always when you are defining API, even an internal one, think about allowing others to do things you might not anticipated.)

piplmesh/frontend/views.py
```diff
@@ -164,6 +164,43 @@ def send_update_on_comment(sender, comment, post, request, bundle, **kwargs):
     # want REST request to finish quick
     tasks.send_update.delay(serialized_update)
 
+@dispatch.receiver(signals.hug_run_updated)
+def send_update_on_hug_run(sender, post, request, **kwargs):
```
mitar, line 168: I don't understand. Why you again fetch Post object here in this function and then send whole list of all runs and hugs?

Why not just send an update for the particular change? If run is removed, send only that one removed run. If run is added, send only that. Why are you sending whole lists of everything every time. This is really not nice approach. Optimize please.

piplmesh/frontend/views.py
```diff
@@ -164,6 +164,43 @@ def send_update_on_comment(sender, comment, post, request, bundle, **kwargs):
...
+                'username': hug.author,
+                },
+            'created_time': hug.created_time,
+            'resource_uri': '/api/v1/post/' + str(post.id) + '/hugs/' + str(hug.id) + '/',
+        }
+        hugs.append(s)
+
+    runs = []
+    for run in api_models.Post.objects.get(pk=post.id).runs:
+        s = {
```
mitar, line 183: Don't say to me that you are creating manually something which looks like run and hug REST objects? This is really not good approach. What if we add later on some additional fields to run and hug resources? Then we will have to search around the code and find all the places where we have to update the code as well?

See how `send_update_on_comment` is done.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -132,77 +243,13 @@ function Post(data) {
...
-
-        hug_run_link_click(hug_link, run_link, 'hug', gettext("Hug"), gettext("Unhug"), gettext("Run"));
-        hug_run_link_click(run_link, hug_link, 'run', gettext("Run"), gettext("Unrun"), gettext("Hug"));
+        var hug_link = $('<a/>').addClass('hand');
+        var run_link = $('<a/>').addClass('hand');
 
         var hug = $('<li/>').addClass('hug').append(hug_link);
         var run = $('<li/>').addClass('run').append(run_link);
 
+
```
mitar, line 252: Unnecessary line?

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -113,6 +113,117 @@ function makePanelsOrderUpdatable() {
...
+                    'type': 'DELETE',
+                    'url': url
+                });
+            }
+        });
+    }
+
+    hug_run_link_click(hug_link, run_link, 'hug', gettext("Hug"), gettext("Unhug"), gettext("Run"));
+    hug_run_link_click(run_link, hug_link, 'run', gettext("Run"), gettext("Unrun"), gettext("Hug"));
+
```
mitar, line 181: Unnecessary empty line.

piplmesh/frontend/static/piplmesh/js/home.js
```diff
@@ -113,6 +113,117 @@ function makePanelsOrderUpdatable() {
     });
 }
 
+var createHugRun = function (data, hug_link, run_link, hugs_runs) {
+    var self = data;
+
+    hug_link.data('selected', false);
+    hug_link.text(gettext("Hug")).css('font-weight', 'normal');
+    hug_link.unbind();
```
mitar, line 121: Huh, this is dangerous. `.unbind` removes all event handlers, maybe even some from some other parts of the code. This is not the best way to do it.

BTW, `.off` is a preferred way to do this in newer jQuery versions.

mitar:
Otherwise it looks like it works. Nice.

